PR #577, opened 2014-04-18
Html legend

davejlong:
Adds a basic HTML legend mixin. For now it only supports vertical legends, but I'll be adding options similar to the `dc.legend` mixin.


manju-reddys:
I strongly feel that HTML legend make more sense and it in-fact make life much easy atleast in the following cases:

Case-1: huge list of items (see image 1 in our application - summarize data of 300K records since we couldn't show the complete list we added ```"Others"``` to show there more items below )

- Showing them as a vertical list would be impossible because the SVG size may not be enough to fit all of them within the view-able area.
- Even if we manage to show with adjusting gaps etc... it will be nightmare for user to walk through the list to find the particular item in the list.
- Otherwise we must provide the arrow (up/down) to indicate more items are available and this logic would be too complicate and also its too much burden to user to click through the list to find the item of interest.

Case-2: Add advance features:

- Its not easy to implement the search or filter feature in SVG when the list is huge to find the item of interest (see google maps - search by typing address - for example how it will be helpful).
- In many cases it will be convenient to take controls of legends items to enable/disable filter on click of item based on certain condition (ex: if user is Guest / Author) - this just an example.
- Would allow to taking advantage of JS frameworks data binding (Angular 2) and custom template to add meaningful presentation etc...
- Add collapse and expand the legend or add more features like drag to position on the chart.

Case-3: Customize (see the google map image)

- Allow to skin the legend to the application theme (material design!).
- Not everyone JS developer are styling experts, many of the JS developer use the CSS frameworks like bootstrap to develop the website/web-applications and those frameworks doesn't provide the styling feature for SVG.

legend
google-maps

gordonwoodhull:
Merged via #1392 for 3.0.0

PR #599, opened 2014-05-22
Add .barWidthAccessor to manually specify (per-bar) widths for barCharts

bloer:
This is an attempt to address my question on StackOverflow at http://stackoverflow.com/questions/23764932/dc-js-bar-chart-different-widths-and-title-problems .

The width is specified in pixels rather than in the default xaxis units because I couldn't think of a way to have a sensible function that would work with time axes.

### DIFF
--- a/spec/bar-chart-spec.js
+++ b/spec/bar-chart-spec.js
@@ -56,7 +56,19 @@ describe('dc.barChart', function() {
                 });
             });
         });
-
+        
+        describe("with custom barWidthAccessor",function() {
+            beforeEach(function() {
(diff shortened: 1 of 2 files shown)


pelegm:
I think this is generally a good idea. When your data is highly skewed (see for example [this SO question](http://stackoverflow.com/questions/19288865/generating-histogram-for-highly-skewed-data)), you may wish to split the bins not evenly, to see better with better resolution what happens in highly dense areas, and to "merge" very low bars into a one wider (and higher) bar; this, without changing the scale itself of the x-axis, which may remain, say, linear.

PR #601, opened 2014-05-23
Add 'clickFilterBehavior' to base-mixin to allow ctrl-click joins

bloer:
The clickFilterBehavior option is intended for ordinal charts, bubble charts, pie charts, etc, i.e., those where clicking on an svg node filters to that node. I have not tested what would happen if a user tried to enable on a brush filter for example. 

.clickFilterBehavior has 4 modes:
- "replace": The clicked node replaces the current filter list. Note that it is not possible to remove the filter in this case without a reset link.
- "toggle": Filtering for the clicked node is toggled from its current state. Therefore, if no filters are applied, clicking a node will filter all but that node (by applying an inverse filter under the hood).
- "replaceThenToggle": This is the current behavior. If there are no filters applied, act as 'replace'. If there is one ore more filter applied, act as 'toggle'. 
- "replaceCtrlToggle": This acts like selecting graphics objects in e.g. PowerPoint.  Left-click acts as 'replace', and ctrl-click acts as 'toggle'.

### DIFF
--- a/src/base-mixin.js
+++ b/src/base-mixin.js
@@ -65,6 +65,7 @@ dc.baseMixin = function (_chart) {
     var _legend;
 
     var _filters = [];
+    var _inverseFiltering = false;
     var _filterHandler = function (dimension, filters) {
         dimension.filter(null);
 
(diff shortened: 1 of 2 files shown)


Mavlarn:
So what's the status of tis pr?

PR #683, opened 2014-08-26
Row Chart: optionally ignore values that are zero

dimirc:
I see that there are some [hacky proposals](https://groups.google.com/forum/#!topic/dc-js-user-group/UERVo4QGX7w) if users need to prevent having rows (elements) that appear with data.value = 0

This PR introduces `chart.ignoreZeroValue(true)` to enable this functionality

### DIFF
--- a/src/row-chart.js
+++ b/src/row-chart.js
@@ -53,6 +53,8 @@ dc.rowChart = function (parent, chartGroup) {
 
     var _rowData;
 
+    var _ignoreZeroValue = false;
+
     _chart.rowsCap = _chart.cap;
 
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
I am closing this in favor of the fake groups, for now, and a more transparent data model in the future. All of @mr23's examples of generalizations convince me that this is just a filtering problem and dc.js takes objects which provide some data, rather than crossfilter objects per se.

I opened #921 to include [the FAQ adapters](https://github.com/dc-js/dc.js/wiki/FAQ#filter-the-data-before-its-charted) in dc proper.

PR #691, opened 2014-08-29
Guard for -Infinite/Infinite in key/valueAccessors

kubelsmieci:
If scale is log(), then value can be calcuted to be -Infinite. d3.js throws exception when assigning attribute to element: 'Error: Invalid value for <g> attribute transform="translate(-Infinity,191)".'
Probably similar changes should be made in other places/charts.


gordonwoodhull:
@kubelsmieci, I'm going to close this for the reason above.  Please comment if you think the changes to bubbleChart are still needed.

PR #695, opened 2014-09-08
Show filtered elements on render and not only on redraw

npmiller:
Filtered elements were not shown on render only on redraw. Useful to resize charts for example.


gordonwoodhull:
Now I wonder if this had something to do with linked brushes (so a chart could be filtered even though it has no brush). I'm going to close this and we can address it when we have linked brushes in 2.1.

Issue #681, opened 2014-08-24
Highlighted bar in sync between bar charts of same dimension


dimirc:
I have 2 bar charts that use the same dimension (by month). One showing quantity of orders and another the sum of the total of the orders.

Each bar represents a month, so I wanted that if I press one month in any of the charts, that the other chart also presents the same highlighted month too. (in sync)

Finally checking the code and reading some suggestions from the [user group question](https://groups.google.com/forum/#!topic/dc-js-user-group/1nIScnbY0vs) I was able to do it.

I'd like to help on trying to put this feature/functionality in the core instead. Maybe we could have a new method on base-mixin so that charts that are part of same syncGroup to pass the filter to the others charts from same group and reflect correctly the highlighted element?

PR #697, opened 2014-09-09
Add filtering information to the data table

npmiller:
Add the information if a line of the table is filtered or not. This info is stored in a CSS class .selected or .deselected. This allows us to display filtered data on the table as well.

### DIFF
--- a/src/data-table.js
+++ b/src/data-table.js
@@ -162,7 +162,12 @@ dc.dataTable = function (parent, chartGroup) {
 
         var rowEnter = rows.enter()
             .append('tr')
-            .attr('class', ROW_CSS_CLASS);
+            .attr("class", function(d) {
+                var baseClasses = ROW_CSS_CLASS;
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
I think this will be relevant with the fix for #681 and a constellation of other problems which is in [the sync-filter branch](https://github.com/dc-js/dc.js/tree/sync-filter). Linking this to those to merge at the same time, and add a test for this group-as-dimension trick that makes this one work.

vmantese:
http://loganalysis.github.io/analytics-js/ is down. 
npmiller do you mind reposting that?

npmiller:
@vmantese ah, sorry, the url changed, use this one:
https://loganalysis.github.io/analytics.js/

vmantese:
Beautiful.

vmantese:
Can we close this?

gordonwoodhull:
Sure

PR #698, opened 2014-09-09
Add layer rendering to doRedraw for the map

npmiller:
The doRedraw function now add the missing layers and delete the useless ones.
It is way quicker than calling doRender.


mtraynham:
I've updated #719 to handle add/removal of layers with d3 bindings during redraw.

gordonwoodhull:
Subsumed by #719

PR #719, opened 2014-10-02
GeoChoropleth refactor

mtraynham:
Kind of a large refactor, which adds:
- Auto zoom to geo features with the current projection
- Graticule
- Sphere
- Title accessor for features
- Simplifies render and redraw functions

TODO:
- ~~Fix tests (changes to SVG class names don't match)~~
- Add moar tests!
- ~~Document~~

Future:
- [Projection change transitions](http://bl.ocks.org/mbostock/3711652):

```
var _projectionTween = function (projectionA, projectionB, width, height) {
    return function (d) {
        var t = 0,
            projection = d3.geo.projection(function (λ, φ) {
                λ *= 180 / Math.PI;
                φ *= 180 / Math.PI;
                var p0 = projectionA([λ, φ]),
                    p1 = projectionB([λ, φ]);
                return [(1 - t) * p0[0] + t * p1[0], (1 - t) * -p0[1] + t * -p1[1]];
            }).scale(1).translate([width / 2, height / 2]),
            path = d3.geo.path().projection(projection);

        return function (_) {
            t = _;
            return path(d);
        };
    };
},
```
- ~~[d3.geo.zoom (Jason Davies license?)](http://www.jasondavies.com/maps/rotate/)~~

Strikethroughs have been completed.

### DIFF
--- a/src/geo-choropleth-chart.js
+++ b/src/geo-choropleth-chart.js
@@ -134,7 +134,7 @@ dc.geoChoroplethChart = function (parent, chartGroup) {
      * @memberof dc.geoChoroplethChart
      * @instance
      * @param {String} name
-     * @return {dc.geoChoroplethChart}
+     * @returns {dc.geoChoroplethChart}
      */
     _chart.removeLayer = function (name) {
         delete _layers[name];
@@ -252,7 +252,7 @@ dc.geoChoroplethChart = function (parent, chartGroup) {
      * @memberof dc.geoChoroplethChart
      * @instance
      * @param {Boolean} [showGraticule=false]
-     * @return {Boolean|dc.geoChoroplethChart}
+     * @returns {Boolean|dc.geoChoroplethChart}
      */
     _chart.showGraticule = function (showGraticule) {
         if (!arguments.length) {
@@ -307,6 +307,12 @@ dc.geoChoroplethChart = function (parent, chartGroup) {
             .selectAll('path')
             .data(dc.pluck('features')).enter()
             .append('path')
+            .attr('key', function (d) {
+                var parent = d3.select(this.parentNode).datum();
+                if (parent.keyAccessor) {
+                    return parent.keyAccessor(d);
+                }
+            })
             .attr('fill', 'white')
             .attr('d', _path)
             .on('click', function (d) {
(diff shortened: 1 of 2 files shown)


src/geo-choropleth-chart.js
```diff
@@ -193,82 +160,398 @@ dc.geoChoroplethChart = function (parent, chartGroup) {
...
+        }
+        _zoomable = _;
+        return _chart;
+    };
+
+    // PLOT
+    _chart._doRedraw = function () {
+        var data = _layeredData(_chart.data());
+
+        for (var layerName in _layers) {
```
mtraynham: It does not since we never bind the layer data using d3 and instead just loop them.  Should be a fairly easy change.

mtraynham:
Oof that merge.  I've updated to handle the layer logic better and satisfy #698.  Since it binds the layers in now, it should automatically remove on redraw :)

dberger1989:
This is such a cool feature, but when using a flat-surface projection like mercator, it adds strange effects (as i believe geo.zoom intends for). Is it possible to use the new zoom to get a straight normal zoom onto a flat projection?

Thanks!

mtraynham:
@dberger1989 It's been awhile since I've looked at the code, so it would need a bit of investigation.  I do remember that some of the projections do not support rotation and that may be throwing things off.  There are a few if-statements that may need to be fixed:

- This one handles fitting the projection on render to bring everything correctly into view.
https://github.com/dc-js/dc.js/pull/719/files#diff-ff506ba8d9c0b671b44c576a6aa64674R51
- The d3.geo.zoom function is for interaction, which was derived from [Jason Davies implementation](http://www.jasondavies.com/maps/rotate/).
https://github.com/dc-js/dc.js/pull/719/files#diff-ff506ba8d9c0b671b44c576a6aa64674R503

mtraynham:
Noticed this was getting out of date, rebased against master.

d3netxer:
so these changes were never pulled in?

gordonwoodhull:
Hi @d3netxer, that's correct. 

I'm in favor of pulling this in for 3.0 but the rebase is probably going to be difficult, given that we've already ported `geoChoroplethChart` to d3v4 in the 3.0 branch. 

I also understand that this may not be backward-compatible. May need more tests too.

So definitely not trivial, but welcome if @mtraynham and/or @kum-deepak want to take it on.

Issue #755, opened 2014-11-08
createTitles updateTitles parameter discrepancy


r4j4h:
Hey guys,

I love dc.js, and while extending it I ran into a type issue. This may be something I am overlooking, but I've tracked it down and I think it's real. It's about a `.title callback` during a `pie chart's lifecycle`.

Basically, if you set a .title function, and watch the arguments passed to it, they differ.

From `createTitles`, it gets an object that includes data (containing the real key/value), startAngle, endAngle, and value.

From `updateTitles`, it only gets the data object.

This means that a function written to use `d.key` will break, getting `undefined` initially. The updateTitles function runs shortly after usually and covers this up, but it does seem to be happening.

https://github.com/dc-js/dc.js/blob/ebcd70846233ad8f86e75dfb5ef0043a20d92829/dc.js#L3816

It seems that changing 
`return _chart.title()(d);`
to
`return _chart.title()(d.data);` fixes the issue.

r4j4h:
I updated the branch that I had my title function change to pass the inner datum on. Our discussion here though lends to keeping it passing the outer datum itself (so not including that recent branch change), and changing other things to pass their outer datums instead.

I will ponder and perhaps start on that..

One thing that is beckoning mentioning is if there is a case where there is no outer datum? This would be case where it is illogical and lacking semantic sense, but can be remedied with the meta-understanding of inner vs outer datum and handled with a simple object containing only the data key - thus an empty shell of an outer datum.

gordonwoodhull:
I've decided not to make breaking changes for 2.0... Even though it's such a nice round number and it would be nice to have a clean interface, I think having it be stable is more important than having it make sense. :wink: 

So I'll pull the simple fix for 2.0 that makes it not broken, and consider making the data access consistent for 2.1.

gordonwoodhull:
That's a good question whether there always is an outer datum. I think there always is something which belongs to the chart layout which then contains the datum from crossfilter, but it's worth double checking.

gordonwoodhull:
I found a little bit of the history of the `d` / `d.data` controversy in this comment: https://github.com/dc-js/dc.js/issues/338#issuecomment-26293319

The fix for that issue attempted to change all `d` to `d.data` but I guess it only got through a few charts. 

On the one hand, `d.data` is more what the user would expect, given that they are supplying a crossfilter group or equivalent with key/values pairs. On the other hand, `d` is more powerful because it often contains other, computed data which may be helpful. 

[...]

PR #757, opened 2014-11-11
Forwarding the data object rather than its parent to pieChart's title fu...

r4j4h:
...nction in reference to issue #755


r4j4h:
Hmm, I wonder how GitHub would handle me moving this commit back so that I can rebase instead of the merge... easy enough to do it in another branch, but I'm gonna experiment on this one to see if we can salvage it.

r4j4h:
For reference, I deleted the branch and re-pushed it one commit back (the merge commit). GitHub automatically closed the Issue. Although the branch is now there and a new issue was successfully created, this one is unable to be reopened with a tool tip dialog stating that the branch was `force-pushed or recreated`.

So good behavior, but I think it would be even nicer if it would let us just kind of "clear out" and re-purpose this one.

gordonwoodhull:
Weird, I'm surprised by that. Pretty sure I have done that in the past. Oh well,

PR #766, opened 2014-11-17
Fix ordering option for the barChart.

mhodorogea:
Fix the bug: ordering not working on barChart.

If you write in the barChart definition:

1. **chart.ordering(function (d) { return  d.data.value;   });** 
It will order the barChart xAxis by value increasing.

2. **chart.ordering(function (d) { return  -d.data.value;   });** 
It will order the barChart xAxis by value decreasing.

3. **chart.ordering(function (d) { return  d.data.key;   });** 
It will order the barChart xAxis by key, alphabetical order.

### DIFF
--- a/spec/bar-chart-spec.js
+++ b/spec/bar-chart-spec.js
@@ -951,6 +951,64 @@ describe('dc.barChart', function() {
         });
     });
 
+    describe('check ordering option of the x axis', function() {
+            beforeEach(function() {
+                var rows = [
+                    {x: 'a', y: 1},
+                    {x: 'b', y: 3},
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Sigh, it turns out this was still not quite correct, because ordinarily the ordering function takes a regular group `{key,value}` pair, so the default ordering was not correct and was comparing `undefined`, which can crash crossfilter (see #909).

I am reverting `flattenStacks` and changing the definition of `_ordinalXDomain` to pluck the data from the flattened data and order that, and changing the tests to use ordering functions that work like the other ordering functions (no `.data`).

PR #771, opened 2014-11-20
New select menu widget

madebydna:
I had a need for select menu "chart" type in dc and I have seen it requested in [other places](https://groups.google.com/forum/#!topic/dc-js-user-group/9mHGM-J-ndY), so I created a simple widget that adds this functionality.

The widget creates an HTML select tag with options corresponding to the chart's (crossfilter) groups. By default the option text will display the group key and the group value separated by a colon. The select menu will filter the chart's dimension by the group key represented by the selected option ( also encoded in the option's `value` attribute).

On redraw, the chart filters out options whose value (according to the `valueAccessor`) is < 1. This behavior can be overridden by using the `filterDisplayed` function.

The generated HTML select tag is single select (rather than a multiple select) and on redraw the options that  corresponds to the current filter will be selected. Not sure if a multiple select is more in keeping with the behavior of other `dc.js` charts, but I could change it to a multiple select if needed.

Let me know if there is any interest in pulling this in and, in case there is, whether I should make any changes.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -11,7 +11,6 @@ module.exports = function (grunt) {
 
     grunt.initConfig({
         pkg: grunt.file.readJSON('package.json'),
-
         concat: {
             js: {
                 src: jsFiles,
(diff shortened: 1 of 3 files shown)


anmolkoul:
The single select menu is working nicely. Just saw this thread in a long time. Will take the multi select widget for a spin

gordonwoodhull:
Merged to develop/2.1 branch. Added a `.size()` method to set the height, and an example.

Thanks @madebydna!

Issue #776, opened 2014-11-30
Need a way to remove renderlets


davisford:
I just realized that calling `chart.renderlet()` will build up an array of functions (even the same function), which will continue to spazz out if you end up rebuilding your crossfilter data multiple times, and again calling `renderlet`.

Example of what I mean:

```javascript
var chart = dc.barChart('#barchart');

// rebuild charts with data
function rebuildCharts(data) {
  chart
    .dimension(/* dimension data */)
    .group(/* group data */)
    .renderlet(function (chart) {
        // do something with chart based on data
    })
  dc.renderAll();
}

rebuildCharts([1, 2, 3]);
// renderlet is called and data in closure bound to [1, 2, 3]
rebuildCharts([4, 5, 6]);
// renderlet is called and data in closure bound to [4, 5, 6]
// renderlet is called and data in closure bound to [1, 2, 3]
```

There is no way to remove or clear renderlets, and it doesn't check if the function is already registered.

I think there needs to be a function to clear a specific renderlet, or clear all renderlets.  I only realized this now b/c the scope of the data I was referencing in my renderlet function from the closure was stale (b/c the previous renderlet closure had a reference to it -- so this is causing some bad news memory leaks for me the way I am using it).  

I would propose a method on chart / base chart:

chart.clearRenderlet(fn); // clears that particular function
chart.clearRenderlets();  // just clears them all

Thoughts?  @gordonwoodhull if you agree I'll send a PR.  If you semi-agree, or have a different approach / suggestion, I'd love to hear it.

gordonwoodhull:
fixed by #833

PR #779, opened 2014-11-30
allow clearing of a single renderlet or all renderlets

davisford:
See #776 for rationale

This adds two new functions on chart:

```javascript
var renderletFn = function (chart) { /* do something with chart */ };
// add a renderlet
chart.renderlet(renderletFn);
// clear it
chart.clearRenderlet(renderletFn);
// clear all renderlets
chart.clearRenderlets();
```


r4j4h:
I think gordon's point about the real solution ultimately using d3.dispatch is good to echo here, and I :+1: to having this available in the meantime.

gordonwoodhull:
#776 covers this in a more general way, merged in beta.2. Thanks for bringing this up, @davisford!

Issue #789, opened 2014-12-11
Clip path code broken in latest master


bigianb:
Since updating to master, I see the following error when displaying a chart:
`Failed to execute query: '#1-clip' is not a valid selector.`
I'm pretty sure this is due to pr #733 

Digging a bit more, the issue is that the default anchor name starts with a number ... and DOM ids must start with a letter. Making the change below (i.e. changing '' in the last line to 'id') fixes the issue as the id now starts with a letter.
```
_chart.anchorName = function () {
        var a = _chart.anchor();
        if (a && a.id) {
            return a.id;
        }
        if (a && a.replace) {
            return a.replace('#', '');
        }
        return 'id' + _chart.chartID();
    };
```

r4j4h:
Just updating this issue ticket that my PR is now good to go. :) :+1:

Frozenlock:
Just wanted to chime in: I get the same error when trying any version after early December.

Without being familiar with the code, I tried to pinpoint the origin:
https://github.com/dc-js/dc.js/commit/da2d4eeb1abdd0c12f83d81e5480933e2a9feb45#commitcomment-9174465

gordonwoodhull:
fixed by #800

PR #800, opened 2014-12-16
Applying bigianb's patch as described in issue #789

r4j4h:
I am not sure if links work through titles, this is in reference to issue #789 . It fixed master for me as well as bigianb.

### DIFF
--- a/dc.js
+++ b/dc.js
@@ -959,7 +959,7 @@ dc.baseMixin = function (_chart) {
         if (a && a.replace) {
             return a.replace('#', '');
         }
-        return 'id' + _chart.chartID();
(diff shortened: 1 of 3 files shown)


r4j4h:
@gordonwoodhull So after taking a peek, I think I found a pretty elegant location and test for it in `base-mixin-spec.js` with some other `anchorName` tests.

The existing numerical one still passes unchanged as well! I think that it should rather be testing the "auto-incrementing" nature of it's generated fallback ids instead of just that it contains a number, but either way this is a good forward tsep.

r4j4h:
Sorry to the entire community for taking so long this this :3

PR #824, opened 2015-01-05
Forwarding the data object rather than its parent to pieChart's title fu...

r4j4h:
...nction in reference to issue #755


r4j4h:
Re-created this branch b/c the old one had a merge in it. This has been rebased on top of latest master.

For reference: The old issue #757 had `Bug` Label and `v2.0` milestone

gordonwoodhull:
Thanks @r4j4h!

For future reference it is actually okay to rebase a pull request branch and `git push -f` it. You'll see many maintainers requesting this on Github.

This is because it is highly unlikely that anyone is working on a fork of the same branch, so it is okay if history gets mangled. This seems to be the one exception to the rule.

Issue #817, opened 2014-12-28
Speed up geoChoropleth tests


gordonwoodhull:
These really slow things down. 

Surely more data can be reused.

r4j4h:
I added a PR to help with this, #825

gordonwoodhull:
fixed by #825

PR #825, opened 2015-01-05
Enhancement: Speed up geoChoropleth tests

r4j4h:
My initial analysis showed a LOT of time spent re-parsing JSON fixture, so I altered the tests to re-use the fixtures after they are generated and it seemed to increase the test run speed by a large factor.

This is for Issue #817 .


gordonwoodhull:
Excellent. Great idea.

gordonwoodhull:
Nice, `grunt test` is  about 6-8 seconds or 33% faster on my machine. Merging to beta.2

PR #831, opened 2015-01-07
Improved default ordinal palette

### LINKED ISSUE #827: default color scale (for pie) doesn't work
Hi,

at least for the pieChart, the default category20c is not the best choice IMO:
most of the adjacent colors are with little contract between themselves (eg. the initial 4 colors are different shade of blue).
https://github.com/mbostock/d3/wiki/Ordinal-Scales#category20c

When used for  pie, it means it's making harder to distinguish between slices, and it tend to (wrongly) visually cluster the first 4 slides together, so category 4 seems to be more similar to the first 3 items than the 5th one.

I've put 
       .colors(d3.scale.category10());
 and it looks much better.

What was the reasoning to use 20c() instead of 10() as the default? shouldn't we at least re-shuffle the colors to increase the contrast between adjacent items?

tttp:
better palette IMO - fixes #827. Still not super happy, the palettes I found didn't work with all colors with the text in white (was it a reason to use white as the text?)

### DIFF
--- a/src/color-mixin.js
+++ b/src/color-mixin.js
@@ -6,7 +6,12 @@ as a mix-in for any concrete chart implementation.
 **/
 
 dc.colorMixin = function (_chart) {
-    var _colors = d3.scale.category20c();
+    var _colors = d3.scale.ordinal().range( 
+        // qualitative 12 color schemes set3 from colorbrewer concatenated+removed grey and light colors
+      [ '#1f78b4', '#b2df8a', '#cab2d6','#33a02c', '#fb9a99', '#a6cee3' ,'#e31a1c', '#fdbf6f', '#ff7f00',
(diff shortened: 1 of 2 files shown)


mtraynham:
Take a look at colorbrewer.  
https://github.com/alexmasselot/colorbrewer
http://colorbrewer2.org/

gordonwoodhull:
That's what this patch contains, some selected colors from colorbrewer. 

Another breaking change but a good idea IMO. The default colors we have been using may be optimally distinguishable, but unsightly, and as Xavier points out in #827, too consecutive.

gordonwoodhull:
We've deprecated `schemeCategory20c` in dc.js 3.0, and added `dc.config.defaultColors` as a way to provide a consistent set of ordinal colors for all charts. d3-scale-chromatic provides some good palettes, and we'll move to one of those in 3.1

Issue #775, opened 2014-11-28
setting the chartGroup should put the chart in that group


gordonwoodhull:
As noted here:

http://stackoverflow.com/questions/27167079/dc-js-groupchart-example-with-dc-redraw

setting the chartGroup does not automatically put the chart in that group, and it's necessary to register the chart manually.

https://github.com/dc-js/dc.js/blob/master/src/base-mixin.js#L997

However, if the chartGroup is specified to the constructor, the constructor will call `.anchor()`, which *will* register the chart.  It's doubly weird that another setter will update what the specific setter will not.

PR #834, opened 2015-01-13
Fixes Bug 775 - setting the chartGroup

mtraynham:
Fixes Bug #775 

I think the logic between the two functions (i.e. `anchor` & `chartGroup`) is correct now.  I'm not really a fan of having more than one way of doing things (`anchor`, `chartGroup`, `dc.chartRegistry`), but I can't really pull this out for compatibility.

Also, there was a weird test case from the composite mixin spec that inadvertently failed after my pretty minimal change.  After some investigation, I think it was broken, but passing?  Could use a second pair of eyes on that.

### DIFF
--- a/spec/composite-chart-spec.js
+++ b/spec/composite-chart-spec.js
@@ -327,8 +327,8 @@ describe('dc.compositeChart', function() {
 
                 dateValueGroupLine2.on("click")(dateValueGroupLine2.datum());
                 expect(dateValueGroupLine2.text()).toBe('Date Value Group Line 2');
-                expect(d3.select(chart.selectAll('g.dc-legend g.dc-legend-item')[0][3]).classed("fadeout")).toBeTruthy();                
-                expect(chart.selectAll("path.line").size()).toEqual(3);
+                expect(dateValueGroupLine2.classed("fadeout")).toBeFalsy();
+                expect(chart.selectAll("path.line").size()).toEqual(1);
             });
         });
     });
--- a/spec/core-spec.js
+++ b/spec/core-spec.js
@@ -80,6 +80,20 @@ describe('dc.core', function() {
             dc.deregisterChart(chartGrouped, chartGroup);
             expect(dc.hasChart(chartGrouped)).toBeFalsy();
         });
+
+        it('should have switched to an existing group', function () {
+            chart.chartGroup(chartGroup);
+            expect(dc.hasChart(chart)).toBeTruthy();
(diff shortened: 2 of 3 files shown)


mtraynham:
Change should be directed toward the `master` branch.  Can't seem to change it which once you create the pull request :(

gordonwoodhull:
Great, thanks! Yes there is a lot of annoying redundancy. I think the best we can do is make it consistent and deprecate where we can. 

Np about the branch. It's not too hard to redirect these, with care. When the branches diverge more it might get tricky.

spec/composite-chart-spec.js
```diff
@@ -327,8 +327,8 @@ describe('dc.compositeChart', function() {
 
                 dateValueGroupLine2.on("click")(dateValueGroupLine2.datum());
                 expect(dateValueGroupLine2.text()).toBe('Date Value Group Line 2');
-                expect(d3.select(chart.selectAll('g.dc-legend g.dc-legend-item')[0][3]).classed("fadeout")).toBeTruthy();                
-                expect(chart.selectAll("path.line").size()).toEqual(3);
+                expect(dateValueGroupLine2.classed("fadeout")).toBeFalsy();
+                expect(chart.selectAll("path.line").size()).toEqual(1);
```
gordonwoodhull: ugh, how are these related to the bug being fixed here? please tell me they are not.

gordonwoodhull: @mtraynham, so.... if you have to make seemingly random changes to the tests like this, that just might a sign that the changes broke something. 

In this case, since the compositeChart `generateG()` calls `child.chartGroup(_chart.chartGroup())`, and the child charts have the same anchor as the parent, it ends up deregistering the composite chart and registering the child chart instead! So now all but one of the lines disappear when the chart rerenders.

I'll fix this up in another commit, once I figure out a good way. I guess this proves that testing is good, because it wasn't at all obvious that such a harmless-looking change could have such a disastrous effect. A little scary that this only broke one test, though.

mtraynham: Good catch!  I guess I was a little offset by the fact that only one test failed instead of all of them...

gordonwoodhull: the test is looking for a faded out legend item, and it also seems weird for there only to be one line remaining, so i was suspicious. i'm adding a flag for "this is a child chart", which seems like something we need to know anyway.

PR #835, opened 2015-01-13
Fixes Bug 810 - Update License

mtraynham:
Fixes Bug #810 

Pretty much took your suggestion verbatim and added the AUTHORS link as well.  Feel free to push back.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -175,18 +175,8 @@
 
    END OF TERMS AND CONDITIONS
 
-   APPENDIX: How to apply the Apache License to your work.
-
-      To apply the Apache License to your work, attach the following
-      boilerplate notice, with the fields enclosed by brackets "[]"
-      replaced with your own identifying information. (Don't include
-      the brackets!)  The text should be enclosed in the appropriate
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -175,18 +175,8 @@
...
-      boilerplate notice, with the fields enclosed by brackets "[]"
-      replaced with your own identifying information. (Don't include
-      the brackets!)  The text should be enclosed in the appropriate
-      comment syntax for the file format. We also recommend that a
-      file or class name and description of purpose be included on the
-      same "printed page" as the copyright notice for easier
-      identification within third-party archives.
-
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2015 Nick Zhu & the dc.js Developers
```
gordonwoodhull: According to https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html it should be "2012-2015" since every year has had significant contributions.

gordonwoodhull:
Thanks @mtraynham!

mtraynham:
Ahh good catch.  No problem :)

PR #844, opened 2015-01-21
NPM Dependecy Upgrades

mtraynham:
I went ahead and upgraded all dependencies to their latest releases.  The only thing that gave me an issue was `jscs` which I kind of expected (0.6 -> 1.2).  All of the grunt tasks and tests seem to work correctly.  Interested to see how the sauce-labs build goes.

**Edit** Oh, added the Gemnasium badge.  It's a nice to have, but warns of out-dated dependencies.


gordonwoodhull:
This is great.  However, running the tests in the browser, where I think I was seeing hanging with jasmine 2.* before, shows that 9 of our tests are not actually functioning.  (I'm not sure why they aren't color coded on the command line.) 

For example, the throttle test in event-spec.js is missing 
```
            jasmine.clock().tick(100);
```
after the test, to make it function, and some of the tests in the line-chart-spec.js apparently are selecting nothing and thus expecting nothing.

 I'm fine to check this in without it, but perhaps you'd like to take a look.

gordonwoodhull:
The last non-functioning test is kind of amusing:

```
        it('print simple string and a range', function() {
            expect(printer(["a", [10, 30]]), "a).toEqual([10 -> 30]");
        });
```

mtraynham:
Wow, that is incredibly hard to notice.  That yellow color is almost the same as the green!

mtraynham:
Yeah, looks like somebody mouse cut/drag/pasted it in the wrong place.

gordonwoodhull:
looks like that somebody was me, when i was porting the last of the tests from vows. :-(

glad that jasmine now detects tests without any expectations!

mtraynham:
:P I've fixed that one.  Looking into the others.

mtraynham:
Line chart's don't render dots if `brushOn` is `true`.  Still need to fix the event issue, looks a bit more involved though.

gordonwoodhull:
Should that test should set brushOn(false)?

mtraynham:
Correct.  I think it's good to go now.  All of them should be working properly now.

gordonwoodhull:
Just to be sure, and to enjoy how much faster the tests run after #825, I ran the tests on 5 browsers on 3 OS's. Looks great. Rebased to master and will push as 2.0.0-beta.5.

gordonwoodhull:
merged 

(hmm, gitflow doesn't always seem to maintain the commit messages that make closing PRs automatic)

mtraynham:
Which branch did that make it into?  Don't see that commit any where.

mtraynham:
Ahh, it seems the merge from master -> develop didn't pick up the `package.json` changes.

gordonwoodhull:
Sorry about that!  So used to that conflict that I was careless there.  Fixed now.

PR #845, opened 2015-01-21
Update changelog issue references with links

mtraynham:
Simple regex find and replace:

`#(\d+)`

`[#$1](https://github.com/dc-js/dc.js/issues/$1)`

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,21 +1,21 @@
 ## 2.0.0 beta 4
- * make barChart.onClick a proper override (fixes the complaint in #168)
+ * make barChart.onClick a proper override (fixes the complaint in [#168](https://github.com/dc-js/dc.js/issues/168))
 
 ## 2.0.0 beta 3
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
I had hoped that since the changelog is hosted on github, it would use github flavored markdown, but I guess not.

I guess we really are going to need to generate it automatically.

mtraynham:
You can use relative paths, but the issues page is not really a relative path...  There might be a to use markdown link references, but I don't know if you can concat those with strings...

http://daringfireball.net/projects/markdown/syntax#link
```
  [google]: http://google.com/        "Google"
  [yahoo]:  http://search.yahoo.com/  "Yahoo Search"
  [msn]:    http://search.msn.com/    "MSN Search"
```

gordonwoodhull:
Yeah, apparently not:

[this ticket][issue]/845

[issue]: https://dc-js/dc.js/issue

PR #846, opened 2015-01-21
Color domain ordinal #816 fix

### LINKED ISSUE #816: calculateColorDomain doesn't make sense for ordinal domains
While reviewing #813 I noticed that this function of the color mixin is only calculating the min and max color values. An ordinal scale's domain is a list of values, not an extent.

QuinnLee:
This is to fix #816, but not ready to be merged.  I want to get feedback first.

`d3.scale.ordinal()` has a `rangePoints` while `d3.scale.linear()` do not.  This checks it and sets the color domain.

One problem I realized is how to handle sorting when generating the color domain for ordinal scales.  @gordonwoodhull What do you think should happen?  
Should the set of `_chart.data().map(_chart.colorAccessor());`?
What about the order?  Or should this option not set the domain when scale is ordinal?

Thanks!


gordonwoodhull:
Hmm, I wonder if we should use the ordering that we use for the x domain:

https://github.com/dc-js/dc.js/blob/a2e765fe72d0cb3101229ac03323c79bfdd6a86f/src/coordinate-grid-mixin.js#L377

I guess that would be consistent, anyway.  Maybe in an ideal world there would be an ordering function for each ordinal domain, but right now we order the data, not the keys.

QuinnLee:
@gordonwoodhull I pushed new changes. I've added 
``` Javascript
_chart._ordinalColorDomain = function () {
    var groups = _chart._computeOrderedGroups(_chart.data());
    return groups.map(_chart.colorAccessor());
 };
```

Issue #847, opened 2015-01-27
Change to d3.style()


lord-xeon:
When setting the fill, stroke, text-anchor, & opacity, d3.style() should be used instead of d3.attr().

Using d3.style() allows the values to be overridden by the user in with CSS if necessary, whereas d3.attr() sets the style attributes on the SVG element itself.  This makes customizing the charts easier for the developer/designer.

mtraynham:
The opacity, text-anchor and maybe the stroke seem somewhat reasonable as styles.  They rarely change.  

The fill on the other hand, I would argue that we should not use styles.  Fills are constantly changing based on the color-mixin, and typically you'll have multiple fills per chart.  So there is really nothing to gain using CSS.  If you override it, are you expecting many colors to be come one color?

lord-xeon:
Keeping the style attributes on the style element allows better debugging in the inspector, and better setting of colors.

Also, when using a brush on bar charts, the bars get the .deselected class, which changes their color, but perhaps I'd like to reverse that.  Granted, I'd have to use !important, but it would be possible via a style sheet.

I agree though, for the most part, fill isn't changed as much as the other 3 styles would be.

mtraynham:
I always seem to forget how CSS overrides work (especially in conjunction with SVG attributes). For example, dc.js usually set's the attribute `fill`.  The `deselected` class get's added and it's represented in the CSS as the style `fill`.  So does the style get applied after the attribute, and therefore has precedence?

gordonwoodhull:
Yes, styles should override attributes - the attributes sort of become an invisible stylesheet which is overridden by any other styles.

http://stackoverflow.com/questions/24293880/svg-why-does-external-css-override-inline-style-for-text

So is this a clarity issue, seeing something in the tag which is not used, or is there something missing functionally?

gordonwoodhull:
I just tested changing dc.css live in the browser. I can use styles and override attributes just fine.

For example, adding

```{css}
circle {
    fill: black;
}
```

turns the bubbles black, and adding
```{css}
circle {
    opacity: 0;
}
```

makes them disappear. Etc. with `path.line`. This even though there are still attributes on the circles/lines.

@lord-xeon, I think there must be a different problem if you were unable to use style sheets. I'm closing this, please comment if there is still an issue.

PR #848, opened 2015-01-27
#847 fix

lord-xeon:
This is a fix for #847.

I just did a find & replace for all instances of:

    .attr('fill
    .attr('stroke
    .attr('text
    .attr('opacity

Replace with:

    .style('fill
    .style('stroke
    .style('text
    .style('opacity

under the /src directory

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ spec/index-browserify.html
 # font noise from some doc tool
 web/docs/public
 
+dc.sublime-workspace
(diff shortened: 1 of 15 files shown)


gordonwoodhull:
This breaks a lot of tests because the tests are still verifying the attribute. Let's figure out if this is really needed.

gordonwoodhull:
Style sheets override attributes just fine as described in #847, so I'm closing this as invalid.

PR #856, opened 2015-02-06
Responsive X-Axis

robhardwick:
I'm trying to use the dc.js line chart in a context where the chart can be resized by the user. This seems to be working well when the chart width is updated but when the height is changed the x-axis doesn't have its y position updated.

I've created a JSFiddle where you can see the problem (just click "Update"):

http://jsfiddle.net/bsbwjepf/

Looking at the code, the issue seems to be on [lines 417-421 of coordinate-grid-mixin.js](https://github.com/dc-js/dc.js/blob/b61dc54d36d399361643465b7db0fb67cd0e1b25/src/coordinate-grid-mixin.js#L417-421). The x-axis is only translated to the correct position on the first render and not during a redraw (which is what I'm looking for).

I'm happy to submit a pull-request to change this behaviour but would appreciate confirmation that this is a real bug and I'm not just missing something.


robhardwick:
Not sure why the tests have failed, any ideas?

robhardwick:
It looks like the tests were broken by the upgrade from grunt-contrib-jasmine v0.7.0 to v0.8.1 in commit d467386. If I revert the package to the old version and run the tests locally they all pass.

I can add this to my pull request if that makes it easier but maybe you would prefer to deal with it yourself? I'm still not sure why the upgrade to jasmine seems to be breaking date comparisons.

gordonwoodhull:
It's okay, we'll fix the tests rather than reverting Jasmine. There is a certain amount of churn simply because of updates to dependencies. Don't worry about it.

gordonwoodhull:
Hmm, this causes the axis to transition in from the top of the chart at first render, kind of extravagant. I'll figure out how to fix that.

gordonwoodhull:
Merged in 2.0 beta 15, with some other fixes for resizing charts. Thanks @robhardwick !

PR #859, opened 2015-02-09
Fixes Jasmine tests for Date Array expectations

mtraynham:
Seen in #856.  Looks like the newer version of Jasmine tests Arrays a bit differently.  My assumption is that it might have previous called a toString on each array, or checked the size and each value of the arrays independently.  Now it seems to do a strict equality comparison (not really a bad thing).


mtraynham:
Closing in favor of @gordonwoodhull 's changes.

PR #860, opened 2015-02-14
Make dc.js possible to be used as a component.

shobhitg:
Adding a component.json will make dc.js compatible with component package manager.
Component eco-system is as big as bower. And most libraries already make themselves component compatible including d3 and crossfilter.

### DIFF
--- a/component.json
+++ b/component.json
@@ -1,9 +1,14 @@
 {
   "name": "dc.js",
-  "repo": "dc-js/dc.js",
+  "repository": "dc-js/dc.js",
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
merged. thanks!

gordonwoodhull:
I see, so we added this just in time for componentjs to get deprecated [on June 21, 2015](https://github.com/componentjs/component/issues/639).

I'm going to leave it in 2.0.* but remove it in 2.1.*

Issue #855, opened 2015-02-05
dataTable group() means something different


gordonwoodhull:
This is one of those inconsistencies which to fix would break a lot of code.

But why oh why does `dataTable.group()` take a function returning a grouping key, rather than a group? If it is doing something completely different, it should not use the same name.

emiguevara:
exactly, very confusing indeed... took a while to figure how to use gouping with tables

emiguevara:
I started looking at the data-table code, and one of the things that really bother me is the group-rows being shown even if there is just one group. I used to hide them with CSS. Today I added a getter/setter to control this as a chart parameter, without breaking back-compatibility. Do you think this could be useful? Changes in this commit: https://github.com/emiguevara/dc.js/commit/e7a60c082853c059abbef200e82ef4aa158ae32c

gordonwoodhull:
Yes, @emiguevara, please submit a PR.  That is something many people have been working around.

Please call it `showGroups`, plural, and if you can add a test for the option that would be great.  It looks like there is already a test that verifies there is a group `tr` in the default case:

https://github.com/dc-js/dc.js/blob/develop/spec/data-table-spec.js#L54

So I'd just need one verifying that there are no group `tr`s when this option is set.

emiguevara:
Hi @gordonwoodhull , that is good. I'll give it a go later today, I've never written any tests in JS but it shouldn't be too difficult. Thanks.

emiguevara:
PR submitted.

gordonwoodhull:
At least we have a warning for this API inconsistency in 2.0 beta 16, thanks to @calvino.

gordonwoodhull:
Also documented `dataTable.group()` and `dataGrid.group()` separately.

gordonwoodhull:
As of 3.0.12, `dataTable` and `dataGrid` have new functions called `section`, `showSections`, and `htmlSection`.  This follows the plan set forth in https://github.com/dc-js/dc.js/issues/1499#issuecomment-467041136

The old functions are still supported, but they emit a suggestion to use the new functions for clarity. They could be deprecated in the future (but there's really no need).

Additionally, the `section` function is no longer required for `dataTable`, since it's common to use a constant string and put everything in one group (and not show the header). The default is to use the blank string.

PR #863, opened 2015-02-25
dataTable showGroups getter/setter

emiguevara:
As discussed in https://github.com/dc-js/dc.js/issues/855 I implemented a showGroups getter/setter and its relative Jasmine tests (took me a couple days to figure Jasmine out).


gordonwoodhull:
Thanks @emiguevara, merged for 2.0 beta 16

PR #864, opened 2015-02-27
Fix a bug with bar charts not returning the instance on render

jpmunz:
Noticed a small issue, the bar chart render function was overridden but the result was not returned so method chaining was broken


gordonwoodhull:
Thanks @jpmunz, fixed this and a couple more in 2.0 beta 15. 

There are probably more, it's easy to forget.

PR #865, opened 2015-02-28
automatic tracking of latest release in web page.

espinielli:
A small contribution to automatically track the latest version of the repo as reported in the web page (using stock.js)

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -37,7 +37,7 @@
 </p>
 
 <h4>Version Status</h4>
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
That is pretty cool.

gordonwoodhull:
Like having one less thing to update on release! Now I just need to remember to always update the release page.  :-)

Thanks @espinielli!

PR #867, opened 2015-03-01
protect against division by zero.

espinielli:
Added the same division-by-zero protection to `yearlyPerformanceGroup` as was already done for [`indexAvgByMonthGroup`](https://github.com/dc-js/dc.js/blob/develop/web/stock.js#L135).

### DIFF
--- a/web/stock.js
+++ b/web/stock.js
@@ -76,8 +76,8 @@ d3.csv('ndx.csv', function (data) {
             p.fluctuation += Math.abs(v.close - v.open);
             p.sumIndex += (v.open + v.close) / 2;
             p.avgIndex = p.sumIndex / p.count;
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Thanks @espinielli. The test failure here means that the initial output has changed, which is somewhat surprising since nothing is filtered initially, so I wouldn't expect zeroes. 

I will take a look for myself when I have time, but can you characterize the difference in the display?

espinielli:
The test fails because it compares the doc file generated from `stock.js`: there is a difference because there is additional code.

gordonwoodhull:
Actually, it doesn't compare the code or the docs, only the output, so I wouldn't expect any difference.  There were a couple of syntax errors in your code which caused the charts not to render.  Fixing those in 73a15a4, it looks good.  Thanks!

PR #837, opened 2015-01-16
Heat map sort

QuinnLee:
Added getters and setters for sorting `cols` and `rows` in a heat-map

PR #869, opened 2015-03-02
Refactored heatmap to allow for custom sorting.  Defaults to ascending.

mtraynham:
The `rows`/`cols` setter/getters were doing a bit too much.  They should only allow the user to specify a custom domain for the specific axis.

Added a `rowOrdering` and `colOrdering` setter/getter to specify a custom sort for each axis.  This can be null for no sort.  Thus if a user supplies their own domain with no sort, it will use the order they provided in their custom domain.

Also, we can reuse the scale instead of recreating it every time a redraw happens.

### DIFF
--- a/spec/heatmap-spec.js
+++ b/spec/heatmap-spec.js
@@ -154,6 +154,64 @@ describe("dc.heatmap", function() {
 
     });
 
+    describe('override scale domains', function () {
+        beforeEach( function () {
+            chart.rows([1]);
+            chart.cols([1]);
+            chart.render();
+        });
+
+        it ('should only have 1 row on the y axis', function () {
+            var yaxisTexts = chart.selectAll(".rows.axis text");
+            expect(yaxisTexts[0].length).toEqual(1);
+            expect(yaxisTexts[0][0].textContent).toEqual('1');
+        });
+
+        it ('should only have 1 col on the x axis', function () {
+            var xaxisTexts = chart.selectAll(".cols.axis text");
+            expect(xaxisTexts[0].length).toEqual(1);
+            expect(xaxisTexts[0][0].textContent).toEqual('1');
+        });
+
+        it ('should reset the rows to using the chart data on the y axis', function () {
+            chart.rows(null);
+            chart.redraw();
+            var yaxisTexts = chart.selectAll(".rows.axis text");
+            expect(yaxisTexts[0].length).toEqual(2);
+            expect(yaxisTexts[0][0].textContent).toEqual('1');
+            expect(yaxisTexts[0][1].textContent).toEqual('2');
+        });
+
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Good work.  I agree that it was odd that the accessors were constructing and returning a scale.

This raises the question whether it's possible (or interesting) to have other kinds of scales for a heatmap. I know for a geographic heatmap it's more like a linear scale (floating point with a certain precision). We can wait to see if anyone asks.

gordonwoodhull:
This supersedes #837.

mtraynham:
Interesting thought on the scales.  In some cases, depending on the domain of data, most of these charts could have any type of scale on their X & Y (Z?) axes.  I'm tempted to extend that issue beyond just heatmaps (even though they are hard-coded to ordinal).

It would be really cool to have a scale resolver routine, which generates a scale based on the type of data fed to it (configurable of course).  Allowing most charts to handle any variant of scale type would be really helpful, and my likely simplify chart creation.

gordonwoodhull:
Automatically resolving the scales is a great idea, however I see it as out of scope for dc.js.  For the most part, dc.js doesn't infer anything and you have to set everything.  

I think it's best to keep it that way and build more libraries on top, like dcplot.js, which does lots of defaulting and inference.  I plan to return to that library soon.

However, maybe that is something different from what you're talking about. I don't know if it's possible to create an adapter that makes any scale work with any chart, but ideas are welcome.

I think a heatmap is just a scatterplot with ordinal scales and square dots. ;-)

renoth:
When will this be available to the public? 2.0.0-beta10 seems to not have this feature.

mtraynham:
@renoth I targeted the PR at the `develop` branch as I saw it more of a enhancement rather than a bug.  Maybe @gordonwoodhull can cherry pick the commits back to `master` if desired.

gordonwoodhull:
Since this changes the interface it has to go into a new version. 

Please download [the develop version](https://github.com/dc-js/dc.js/blob/develop/dc.js). If you need to use a CDN I can see about starting to distribute 2.1.* but I am afraid that this might break a lot of people's work if they are lazy in how they specify their dependencies.

gordonwoodhull:
Any suggestions welcome about how to simultaneously distribute a develop and a master version.

mtraynham:
@gordonwoodhull Does a change in the interface that **is** compatible with pre-existing versions still suggest that it always needs to be in develop?

gordonwoodhull:
? Sometimes I will merge a change to develop if it is risky, but in general if a change does not change the interface then it can be merged to master. This one changes the interface for the better, but it returns the values unsorted instead of a scale.

mtraynham:
Ahh, right it does.  Overlooked that.

PR #876, opened 2015-03-09
Added in cap by percentage feature.

LewisCarey2:
This allows you to specify a percentage and any data values under the
percentage will not be displayed in the graph.

The cap percentage is simply applied over top of the regular cap function. It adds no overhead if you choose not to use the functionality, and all existing code will work (signatures kept the same).


LewisCarey2:
Will just write some quick tests later today and reopen.

PR #877, opened 2015-03-11
added xAxisMin and  xAxisMax functions for row-chart

gazal-k:


gordonwoodhull:
Sorry for the slow response. Where are `xAxisMin` and `xAxisMax` actually used? Seems like they should be called from `calculateAxisScale` where it is now using `d3.extent(...)`. Or are they only intended for chart users rather than for internal chart use?

Also it looks like they should probably use `cappedValueAccessor` instead of `valueAccessor`.

gazal-k:
I wrote this in because I needed xAxisMin and xAxisMax as a chart user rather than for internal chart user. I've changed calculateAxisScale now to use xAxisMin and xAxisMax.

gazal-k:
Decided against using the xAxisMin and xAxisMax functions in
calculateAxisScale. existing d3.extent function is faster.

PR #890, opened 2015-03-20
Allow brushing filters to only occur at brush end

mtraynham:
This is perf change.  Constantly calling brushes can create big issues in performance with multiple charts.  Granted the interval between filtering could also be changed.  This just allows you to turn off that intermediate brushing calls and only trigger at the end.

Also, if brushOn is turned on/off, we should clear the current state of filters.  Mixing range filters with click filters is probably bad.

### DIFF
--- a/src/coordinate-grid-mixin.js
+++ b/src/coordinate-grid-mixin.js
@@ -1,11 +1,4 @@
-/**
-## Coordinate Grid Mixin
-Includes: [Color Mixin](#color-mixin), [Margin Mixin](#margin-mixin), [Base Mixin](#base-mixin)
 
-Coordinate Grid is an abstract base chart designed to support a number of coordinate grid based
-concrete chart types, e.g. bar chart, line chart, and bubble chart.
-
-**/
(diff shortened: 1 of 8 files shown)


src/coordinate-grid-mixin.js
```diff
@@ -891,16 +898,24 @@ dc.coordinateGridMixin = function (_chart) {
         if (_chart.brushIsEmpty(extent)) {
             dc.events.trigger(function () {
                 _chart.filter(null);
-                _chart.redrawGroup();
+                redrawFn();
             }, dc.constants.EVENT_DELAY);
         } else {
             var rangedFilter = dc.filters.RangedFilter(extent[0], extent[1]);
```
mtraynham: Hmm seems one thing is kinda strange with this... After applying a filter, a click outside the brush range intermittently does not clear the filter.  A console log reports that the new click generated the previous range, and then it falls into this code path.

mtraynham:
I'm scrapping this.  It is truly too complicated without breaking every test.

PR #894, opened 2015-03-20
Added external radius padding

mtraynham:
If you have the width/height of the chart filling it's parent, there is no proper way to reduce the external radius of pie chart while keeping it centered (at least with some really hacky margin work).  This solves that issue.


gordonwoodhull:
Thanks @mtraynham, added a test and merged for beta 13

PR #896, opened 2015-03-23
Coordinate grid xDomain change

mtraynham:
Crossfilter always provides every row, even if the measure value is 0.  When not using Crossfilter, this will almost never be the case.

This check ensures that the xAxis is updated if the previous domain length is not equal to that of the new domain.

### DIFF
--- a/src/coordinate-grid-mixin.js
+++ b/src/coordinate-grid-mixin.js
@@ -393,7 +393,8 @@ dc.coordinateGridMixin = function (_chart) {
 
         // has the domain changed?
         var xdom = _x.domain();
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
This logic is getting really hairy. I wonder if there would be any risk to just always rescaling. Are there cases where we actually want the `_xUnitCount` not to change when the domain has changed?

mtraynham:
The only reason not to may be performance, but I believe that has fairly minimal overhead anyway.

gordonwoodhull:
Moved that hairy logic to a function, and compare elements by string value for good measure (to avoid date problems). Thanks @mtraynham!

mtraynham:
Thanks Gordon!

PR #898, opened 2015-03-25
Performance: Only filter on brush end

mtraynham:
Allows coordinate grid charts to only filter on brush end.


guilhermecgs:
I have tested this and it works fine...

If I keep the button pressed, the filter only occur when the brush ends... as expected...   This improve the performance and usability a lot.

However, it looks like there is a lag when I start pressing down the mouse button... There is a delay between the mousedown event and the feedback to the user. Probably something not related to the changes in this PR

gordonwoodhull:
That's great, thanks @guilhermecgs.

Does it help to set `dc.EVENT_DELAY = 0;`? I think that you should not need it since that's part of the throttle/debounce which this feature effectively disables.

Then again, it's possible that there is some expensive calculation going on between the mousedown and the brush showing. There shouldn't be, but it's possible.

guilhermecgs:
@gordonwoodhull 

Also, It would be great to add an event to get the preFilter event.. There is only the "filtered"event today

Between prefilter and posFilter, I could add a loading css

gordonwoodhull:
Hmm, good point - applying the filter can be slow in itself. Right now I guess you could do this by implementing the filterHandler as a pass-through:

```js
var oldHandler = chart.filterHandler();
chart.filterHandler(function (dimension, filters) {
  // start your loading css here
  return oldHandler(dimension, filters);
})
```

PR #900, opened 2015-03-26
Fixes filtering on render

mtraynham:
After a render, an ordinal coordinate grid doesn't retain its faded area.  Simple fix.


gordonwoodhull:
Thanks. I think there are more than a few places where dc.js assumes nothing will be filtered on render.

gordonwoodhull:
Thanks @mtraynham, merged.

Issue #781, opened 2014-12-03
sunburst or nested donut chart


PBrockmann:
It would be nice to have a sunburst or a nested donut chart directly from dc.js.
I call sunburst or nested donut chart a chart like 
![1-sunburst](https://cloud.githubusercontent.com/assets/5402758/5286151/329b2c0a-7b22-11e4-8ea0-de91e86e7dba.jpg)
Is it possible, already done by someone, I would be very interested to know.

PBrockmann:
Is there fine control on colors ? I would like to have colors of sub pie-slices derivated from colors of pie-slices from the 1st level ? I can control them from the css but then would need hierarchy in name of slice (for now pie-slice _1, pie-slice _2, ...).

IgorShch:
@PBrockmann Could you post the code you used the generate the sunburst chart? I struggle to understand how to get crossfilter to represent the hierarchical data. Thanks a lot

blairn:
I'll explain it.

We use an array for a key.

say we have these lines in a csv.

location, value
animal/moose, 5
animal/cat,2
animal/frog,3
location/house,3
location/fridge,4
location/island,2
ant/red/angry,3
ant/red/happy,2

You can see you have a hierarchy (filled with junk data, but a hierarchy
all the same)

we turn the data into the form....
{location:['animal','moose'] , value:5}
{location:['animal','cat'], value 2}
...

and push that into crossfilter (with other fields, since you will want to
actually be able to crossfilter it on something!).

When we get the data out, the key still contains the path in the
heirachy...

How we do the filters, is.... crossfilter lets you pass in a function
describing the filter. so... we can build a filter for ['animal'], which
filters for everything whos path starts with animal.

I'll post code later today. I'm unfortunately without net at home. and
doing this off the top of my head from a cell phone is pretty hard -
hopefully it is giving you an idea of what is going on.

if not, I'll post the actual code fragments which do things soon.

IgorShch:
@blairn Thanks a lot. I found the code in the examples folder, but I'm sure it will be useful for other people. This graph is just awesome. If you are going to post an example, could you please explain how to control the width of every doughnut?

blairn:
You don't yet :)

Currently it is using an sqrt function to set them, I'll make it so you can
set your own function.

nsubordin81:
[...]

PR #907, opened 2015-04-08
Sunburst

blairn:
I am not good at git, but... I think this is right.
It works, tests, lints. I'll do more clean up in the next few days, along with more charts being pull requested. 

Enjoy!


gordonwoodhull:
Examples are great, but we will need Jasmine tests as well. Please see if you can exercise the functionality with some tests similar to https://github.com/dc-js/dc.js/blob/develop/spec/pie-chart-spec.js

blairn:
I thought I did have tests o.O. I remember writing them, I'll go have a
look now

blairn:
ah, I do have tests, but forgot to commit them :( - me and git have a
difficult arrangement.

How do you want me to add them? commit it and build another pull request
replacing the one I did before? or send it to you so you can commit it in?

blairn:
ah... my tests fail LINT something awful, I'll work on getting them working
^.^ and make a new PR. Good night, and thanks for everything! dc.js
wouldn't be nearly as wonderful without you!

gordonwoodhull:
Yes, adding them to the same branch and pushing them is the right approach. A PR it's always a work in progress. Lmk if you get stuck and I'll help troubleshoot. The tests are often pretty challenging.

KatiRG:
I was able to get the sunburst chart working using the code here https://github.com/blairn/dc.js/tree/sunburst, but the weird this is that it turns all my rowChart bars dark grey. Anyone else find this?

KatiRG:
for now we have just hard-coded it in the css:

`.dc-chart g.row rect {
    fill: #1f77b4;
}`

gordonwoodhull:
@csymill26 asked in #1069 and #781:
> The sunburst is an awesome graph. Is there a way to make it zoomable? I have not been able to make it do so. Also, `.innerRadius` does not seem to work for me either. No matter the value I give it, the inner ring is the same size. I saw someone else mentioned having this problem. Is there a fix? Another thing, is it possible to add spaces between the levels? Thanks!

gordonwoodhull:
@csymill26, would you expect the sunburst also to "filter in" to the zoomed part, or is it just a view thing. Also what would be the interface, something like single click to filter, double click to zoom?

csymill26:
Zooming in would be a visual thing as well as a filter thing.  It doesn't matter to me if it's a single or double click to zoom.

csymill26:
Couple questions:  Is the inner radius value working yet?  Will you add zooming?  If it/they are not working yet, when will these become available?  Thank you!

gordonwoodhull:
Unfortunately I think @blairn is quite busy, since he hasn't gotten back to this PR for a while. I'd be willing to write the tests if someone else wants to implement these features. 

@csymill26, probably your best bet if you need this stuff immediately is to start from Blair's source and make the changes yourself. Then please file another PR with your updates!

gordonwoodhull:
Oops, I missed the tests being added. That's awesome. I guess it is on me to get them running properly.

staveris:
Hi regarding innerRadius, i noticed that in dc.js line 4548 [here](https://github.com/blairn/dc.js/blob/sunburst/dc.js#L4548) inner and outerradius are set as functions of d.y and d.dy. I replaced them with .innerRadius(_innerRadius) and it works. I am not sure if some else is lost though

gordonwoodhull:
Merged via #1388 for 3.0.0

Issue #785, opened 2014-12-07
pie charts legends not using label()


sylvinus:
Hello!

Newcomer to dc.js, sorry for the possibly dumb question.

When doing a pie chart, the legend is using d.key, which in my case is an ID and not a label. I'm using the label() function to translate them and the pie chart itself looks fine, but the legend shows the IDs. So for consistency, why not use the label function to generate the names in the legend?

Thanks!

digitalsanity:
Is this dead?

gordonwoodhull:
Hi @digitalsanity, dc.js is alive and well, but not every contribution has been merged. 

It looks like the associated PR #916 stalled because I wasn’t sure if the same change had to be made for other charts. I.e. should the `legendLabelAccessor` go on the pie chart or the base mixin?

Without that merged, a workaround could be to override `.legendables` and change labels. Messy though.

PR #914, opened 2015-04-14
add legendLabelAccessor piechart

agrass:
I added a method legendLabelAccessor to change the text on legends in piecharts. It's the same of issue #785


gordonwoodhull:
Hi @agrass, this will be a useful contribution, and the implementation looks straightforward. 

A few questions:
 * I am curious what you need `getLegendLabel` for? The other accessors don't pass the chart as `this`, so maybe this helper function isn't necessary.
 * Would you be willing two simple tests that the text is applied correctly when the function is used, and is just the key when the function is not used? There is a helper function `legendLabel` in spec/legend-spec.js that would be helpful.
 * Would you be willing to add a line of documentation?

In the bigger picture, I also wonder if the functionality belongs higher up, like in baseMixin. It seems like this must be an issue for other kinds of charts. I'll comment on the issue.

Thanks!

https://github.com/dc-js/dc.js/edit/develop/CONTRIBUTING.md

gordonwoodhull:
Hmm, I see you replaced this PR, but I think this all still applicable, will move my comments there.

PR #915, opened 2015-04-14
update forking instructions for git-flow

gordonwoodhull:


gordonwoodhull:
created this PR because i thought it would help get the changes into master, but it didn't. oh well, the instructions on develop are what matter.

Issue #906, opened 2015-04-06
remove renderlet from example and test code


gordonwoodhull:
If we've deprecated it, we shouldn't use it ourselves.

```
grodo:dc.js gordon$ grep -r --exclude=*.min.* --exclude=dc.js "renderlet(" .
...
./spec/bar-chart-spec.js:                        .renderlet(function(_chart) {
./spec/base-mixin-spec.js:            chart.renderlet(firstRenderlet);
./spec/base-mixin-spec.js:            chart.renderlet(secondRenderlet);
./spec/bubble-chart-spec.js:            chart.renderlet(renderlet);
./spec/composite-chart-spec.js:            chart.children()[0].renderlet(function(chart) {
./spec/coordinate-grid-chart-spec.js:                chart.renderlet(function (chart) {
./spec/core-spec.js:            dc.renderlet(function(group) {
./spec/data-grid-spec.js:            chart.renderlet(derlet);
./spec/data-table-spec.js:            chart.renderlet(derlet);
./spec/helpers/spec-helper.js:    dc.renderlet(null);
./spec/pie-chart-spec.js:            chart.renderlet(function() {
./src/base-mixin.js:    #### .renderlet(renderletFunction)
./src/base-mixin.js:    chart.renderlet(function(chart){
./src/base-mixin.js:    deprecated `.renderlet()` method.
./src/coordinate-grid-mixin.js:    chart.renderlet(function(chart){
./src/core.js:        dc._renderlet(group);
./src/core.js:        dc._renderlet(group);
./src/events.js:    chart.renderlet(function(chart){
...
./web/ep/index.html:    .renderlet(function (chart) {
./web/ep/index.html:    .renderlet(function (chart) {
./web/ep/index.html:    .renderlet(function (grid) {
./web/ep/list.js:  .renderlet(function (chart) {
./web/ep/list.js:        .renderlet(function (table) {
./web/examples/bar.html:    .renderlet(function(chart) {
./web/play-ground.html:        .renderlet(function (chart) {
./web/play-ground.html:        .renderlet(function (chart) {
./web/stock.js:        .renderlet(function (table) {
```

PR #917, opened 2015-04-15
removed tests for .renderlet as it is deprecated and replaced usage in e...

AlanKavanagh:
...xamples and tests with .on('renderlet'..)


AlanKavanagh:
As requested in #906

gordonwoodhull:
Thanks, we still need to test the renderlet event, but just change the syntax. I'm about to pull the rest of this, thanks!

PR #919, opened 2015-04-16
Using d3.svg.axis instead of custom axis implementation on heatmap

mtraynham:
This is probably the more desired change to the heatmap, which is to use `d3.svg.axis` instead of a one off implementation.  I imagine our original heatmap implementation was borrowed from: http://bl.ocks.org/tjdecke/5558084
which also doesn't use d3.svg.axis.

The domain line and ticks have been set to `opacity: 0` for consistency of look and feel.

Also as a one off, this properly uses the ordinal scales `rangeBand()` function to give the correct width/height of a heatmap box.

Broke some tests though :(

### DIFF
--- a/dc.css
+++ b/dc.css
@@ -305,6 +305,11 @@ div.dc-chart {
     stroke: none;
 }
 
+.dc-chart .heatmap .axis .domain,
+.dc-chart .heatmap .axis .tick line {
+    opacity: 0;
+}
+
 .dc-chart .heatmap .box-group.deselected rect {
     stroke: none;
     fill-opacity: .5;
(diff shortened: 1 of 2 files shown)


mtraynham:
Tests fixed.  Broke because:
- Rename of the axis class from `rows.axis` -> `axis.y`; `cols.axis` -> `axis.x`; Reverted this.
- Position is no longer checked by `x` & `y` attributes.  We validate the `transform` attribute now.
- And the last one I hate...  `dc.transition` can return the selection or a transition.  So the `each` call needs special handling as `transitions` have 2 parameters, the first being which state the transition is in.
https://github.com/mtraynham/dc.js/commit/10ed200f557533bb9e123d8b63ad06236f85719e#diff-7485f5da2b41a10b58398ae4afbb4f61R223

mtraynham:
Hmm don't know why the stock tests failed when I updated a spec test...

mtraynham:
Doesn't fail locally either...

mtraynham:
Passing again...

Issue #582, opened 2014-05-01
Hiding series in Series Chart


davejlong:
It would be really nice to be able to use the Stack Mixin with a Series Chart to make individual series hidable using the legend.

``` javascript
dc.seriesChart
    .dimension(dim)
    .group(group)
    .hidableStacks(true)
```

mikwat:
I've made an attempt to solve this item. Feedback appreciated. See #920

PR #920, opened 2015-04-16
Hidable support for series charts

mikwat:
In an effort to address #582, I've made some changes to "stack-mixin" to make it aware of series charts through the fact that the `_stack` has only one element. This is necessary because when series charts are redrawn, the `_stack` of each underlying chart is rebuilt. Therefore, any changes to the `hidden` attribute are not persisted.

I also updated "composite-chart" so that it filters hidden children in a few key locations.

I updated a few examples so that they demonstrate the hidable legends and added a stacked line (area) chart example.

If this seems like a reasonable solution, let me know and I'll add some specs. I've tried my best to follow the contribution guidelines, but this is my first time so please let me know if I've made any mistakes.

### DIFF
--- a/spec/series-chart-spec.js
+++ b/spec/series-chart-spec.js
@@ -133,4 +133,50 @@ describe('dc.seriesChart', function() {
             expect(dots[0].length).toEqual(6);
         });
     });
+
+    describe('legends with hidable stacks', function () {
+        beforeEach(function () {
+            chart
+                .chart(function(c) { return dc.lineChart(c).hidableStacks(true); })
+                .legend(dc.legend().x(10).y(10).itemHeight(13).gap(5))
+                .render();
+        });
+
+        describe('clicking on a legend item', function () {
+            beforeEach(function () {
+                legendItem(0).on('click').call(legendItem(0)[0][0], legendItem(0).datum());
+            });
+
+            it('should fade out the legend item', function() {
+                expect(legendItem(0).classed('fadeout')).toBeTruthy();
+            });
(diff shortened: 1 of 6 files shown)


mikwat:
@gordonwoodhull Any feedback appreciated, thanks.

gordonwoodhull:
Sorry for the slow response, I am absolutely swamped. This looks really good at a first glance. It does need tests.

mikwat:
Thanks, I understand. I'll add tests.

spec/series-chart-spec.js
```diff
@@ -133,4 +133,50 @@ describe('dc.seriesChart', function() {
             expect(dots[0].length).toEqual(6);
         });
     });
+
+    describe('legends with hidable stacks', function () {
```
mikwat: @gordonwoodhull Added specs. I copied these from `legend-spec.js` and modified slightly to work with `seriesChart`. Would it be better to consolidate these in `legend-spec.js`? If so, I'd need to duplicate the `seriesChart` setup. Not a big deal, just let me know which approach you prefer.

mikwat:
I know you're busy @gordonwoodhull Feedback welcome whenever you have a chance.

gordonwoodhull:
Thanks @mikwat. I tend to merge and fix any problems myself since I never know who will still be around to respond to code reviews.

This generally looks good and complete. I'll be sure to ping you on any revisions I make.

mikwat:
Sounds good. I'll be around.

PR #922, opened 2015-04-20
fix #897 Selector with brackets not working

### LINKED ISSUE #897: Selector with brackets not working
I use selectors for my charts like this:

    #dimensions li.list-group-item[property=foobar]

This worked until this commit:
https://github.com/dc-js/dc.js/commit/b2059e12676fafa3df2464a156b2805bd64c80de

Now I get an illegal selector string for the clipping path in the coordinate grid mixin.

It can be fixed with (brackets and equal signs are additionally replaced):

    function getClipPathId() {
        return _chart.anchorName().replace(/[ .#=\[\]]/g, '-') + '-clip';
    }

IwfY:
My implementation to fix this issue.

### DIFF
--- a/spec/coordinate-grid-chart-spec.js
+++ b/spec/coordinate-grid-chart-spec.js
@@ -208,6 +208,26 @@ describe('dc.coordinateGridChart', function() {
                 });
             });
 
+            describe('with a selector containing brackets', function() {
+                beforeEach(function() {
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Thanks @lwfY, merged for 2.0 beta 13

PR #923, opened 2015-04-22
added bullet chart

espinielli:
Added a new chart: bullet chart, a must for dashboards!

See it live [here](http://bl.ocks.org/espinielli/e7625ce617e4d9c87cae)

It is based on the [bullet chart d3-plugin](https://github.com/d3/d3-plugins/tree/master/bullet) with a fix as described in the d3/d3-plugins#130

*Note* I explicitly ignored the "original" bullet chart file in the `lint` grunt task: this is code copied from another repo and should stay equal to the original so as to ease future inclusion of newer version (and so be able to see what has changed. Just my humble opinion.)

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -61,7 +61,7 @@ module.exports = function (grunt) {
                 src: ['<%= conf.src %>/**/*.js', 'Gruntfile.js', '<%= conf.web %>/stock.js'],
                 options: {
                     jshintrc: '.jshintrc',
-                    ignores: ['<%= conf.src %>/banner.js', '<%= conf.src %>/footer.js']
+                    ignores: ['<%= conf.src %>/banner.js', '<%= conf.src %>/footer.js', '<%= conf.src %>/d3.bullet.js']
                 }
             }
         },
@@ -415,5 +415,7 @@ module.exports.jsFiles = [
     'src/heatmap.js',
     'src/d3.box.js',
     'src/box-plot.js',
+    'src/d3.bullet.js',
+    'src/bullet-chart.js',
     'src/footer.js'  // NOTE: keep this last
 ];
--- a/src/bullet-chart.js
+++ b/src/bullet-chart.js
@@ -1,0 +1,162 @@
+/**
+## Bullet Chart
+Includes: [Color Mixin](#color-mixin), [Margin Mixin](#margin-mixin), [Base Mixin](#base-mixin)
+
+Concrete bullet chart implementation.
+
+Inspired by:
+
+* [Mike Bostock's bullet chart](http://bl.ocks.org/mbostock/4061961)
+
+Examples:
+
+* ToDO
+
+#### dc.bulletChart(parent[, chartGroup])
+Create a bullet chart instance and attach it to the given parent element.
+
+Parameters:
+* parent : string | node | selection | compositeChart - any valid
+ [d3 single selector](https://github.com/mbostock/d3/wiki/Selections#selecting-elements) specifying
+ a dom block element such as a div; or a dom element or d3 selection.
+* chartGroup : string (optional) - name of the chart group this chart instance should be placed in.
+ Interaction with a chart will only trigger events and redraws within the chart's group.
+
+Returns:
+A newly created bullet chart instance
+
+```js
+// create a bullet chart under #chart-container1 element using the default global chart group
+var chart1 = dc.bulletChart('#chart-container1');
+// create a bullet chart under #chart-container2 element using chart group A
+var chart2 = dc.bulletChart('#chart-container2', 'chartGroupA');
+// create a sub-chart under a composite parent chart
+var chart3 = dc.bulletChart(compositeChart);
+```
+
+**/
+dc.bulletChart = function (parent, chartGroup) {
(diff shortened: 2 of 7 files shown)


mtraynham:
Very, very cool.  Three things:

1. I wonder if we can make this work with the color mixin, so the colors aren't [static](https://github.com/espinielli/dc.js/commit/a621c4679c7c4e17430226cb386819e347603fc5#diff-957dbafaff86e402caa51802e300ac40R26), but user defined.  I imagine [d3.scale.quantize](https://github.com/mbostock/d3/wiki/Quantitative-Scales#quantize) would be the proper option for this setup. 
2. I'd prefer if you revert the following line and get it working with jshint properly: https://github.com/espinielli/dc.js/commit/a621c4679c7c4e17430226cb386819e347603fc5#diff-35b4a816e0441e6a375cd925af50752cR64
3. Is this a Crossfilter groupAll only chart, i.e. you wouldn't use dimensions?

@gordonwoodhull might ask you for some tests :)

gordonwoodhull:
This is great. An entirely different approach from #692, both in terms of data and implementation. I like relying on d3.plugin rather than cramming this into a composite chart. 

I don't have a strong opinion about using more complex data items, as this PR does, versus using multiple data series as #692 does. They both have their advantages.

 1. Agree
 2. I think the reason for skipping jshint on that one file is that it is copied directly from the d3-plugin repo, which seems reasonable enough to me.
 3. I'd like to see the example actually use crossfilter rather than faking the `.data()` like this. It looks like this will work with regular groups, with a rather complex reduce function that creates ranges, measures, and markers.
 4. Yes, tests. I know, it's hard. But I can merge it a lot sooner (to 2.1) if I don't have to write the tests myself.

mtraynham:
I have another idea for 2.  I was originally thinking we add it as a bower dependency, but that idea fell over pretty quickly.  https://github.com/d3/d3-plugins/issues/122

As a second thought, can we just add it as a [git-submodule](http://git-scm.com/book/en/v2/Git-Tools-Submodules) instead of copying the code here directly.  This seems much more appropriate for external code that we have decided to not maintain directly (jshint or otherwise).

The issue reported in https://github.com/d3/d3-plugins/issues/130 could be fixed with a PR and temporarily we just submodule the PR fork.

gordonwoodhull:
@mtraynham, can you submodule just one file, or are you suggesting bringing in the whole repo for the one file? I've found submodules to interact in annoying ways with switching between branches, which I do all the time with git flow. Although technically correct it might be easier just to pull in the file.

gordonwoodhull:
We also do use box.js from d3-plugins so maybe the submodule would be appropriate. I'm not sure.

espinielli:
git submodule does not have the granulrity of a file.
Maybe `git subtree` is an option but I do not know much.

For now I just copied `bullet.js` and renamed to `d3.bullet.js`  as was done similarly for `box.js`.
(Ok, plus the fix) So for easy of merging any newer version of it I avoided changing it due to jshint.

espinielli:
Note that d3/d3-plugins#130 has been fixed in the meanwhile.

SofiaGR:
So I added this chart to dc, but I'm not sure how to make it responsive. Everytime I try, the svg + bullet tends to return null...or I crash things. Any suggestions on how to make this responsive to window resize?

gordonwoodhull:
@SofiaGR, thanks, that's excellent feedback. I really appreciate community review on all PRs.

Currently there is not consistent support for resizing in dc.js charts, so I'm not surprised this contribution does not have the feature either.

The first step to getting this fixed is to contribute a test for it. So far I have not automated the testing for resizing and there are instead an [incomplete set of examples](http://dc-js.github.io/dc.js/resizing/) which you can use to test the feature on each chart by eye.

Would you be willing to copy one of the examples and modify it to show and resize a bullet chart?

You could open a PR with the test or just email it to me.

SofiaGR:
@gordonwoodhull I'll look into it, I figured out the resize mostly, but need to see why the text is dangling around when the window resize occurs.

SofiaGR:
Ok, I have the bullet charts responsive in my app, but I had to make a few changes to the bullet-chart.js source-code for that. 

I also altered the d3.bullet.js merely to allow the user to choose the number of axisTicks since it was a request on my side (smaller screens meant that 10 axis ticks was too crowded). 

Plus I separated the title_orient from the bullet_orient. I wanted to keep the bullet chart thin in height but large in width while having the text  switch around (bottom of bullet chart or left of it in my case) depending on screen size. Text is at the bottom if there's not enough space to show it at the left of bullet chart. By not enough space, I mean a width value that is actually adjusted to my bullet chart widths, so that is not in the source-code.

How do you suggest I approach this now?

gordonwoodhull:
Hi @SofiaGR, sorry I missed this! 

If you're still interested in contributing, you could start another PR based on this one plus your changes. 

dc.js 3.0 will be released soon with support for d3v4, and then we are porting some of these PRs to release them after that.

SofiaGR:
@gordonwoodhull Aaah, 10 days later after I posted that, I started working for a new company. Not sure if I still have that code, need to check :/

gordonwoodhull:
Ha, okay @SofiaGR, thanks for responding. Life is flux. 

Please let us know if you find anything.

PR #925, opened 2015-04-29
Simplifying grunt and bower

### LINKED ISSUE #769: .stack() with .options()
Hi,

Is it possible to use .options() with .stack()? Maybe in form of an array?

ruhley:
* Moving grunt tasks and config into seperate files
  * Minimising the files downloaded when installing through bower

And an old pull request
  * Allow the .options() method to take an array of arguments. Fixes #769


mtraynham:
I'm all for splitting things out, but is this a bit overkill?  I can see us breaking out the `registerGruntMultitask` portions into modules as that emphasizes a clear distinction between code and config.  But the rest of it is just configuration, which seems like it would really hard to understand split out in this way.

Not saying this is wrong, but would love to hear other opinions.

gordonwoodhull:
@mtraynham, I was waiting for your opinion on this. I agree that it makes sense to separate out code, especially the big blocks, but the configuration should stay in the Gruntfile. 

Some more of the big blocks can better be implemented using grunt plug-ins, too. (I'm looking at the examples index in particular.)

gordonwoodhull:
For minimizing the bower download, it looks like #935 is more specific. I intend to merge the parts splitting out the grunt code, and the options stuff, on master (2.0 branch).

ruhley:
For the bower.json ignore property you can either go with #935 listing everything you want to ignore. I prefer to ignore everything and then include only the couple of files I want.

I also prefer to split all grunt config into different files as each file is small and modular.

But as I said these are just preferences, that I thought you might like.

gordonwoodhull:
It does look a lot simpler to ignore everything and then "not-ignore" the couple of files that are needed; I think I'll go with that. Looks like we need to not-ignore LICENSE as well.

mtraynham:
I'm fine with either resolution, just make sure to update the bower version :)

gordonwoodhull:
Yup. Can't we generate it from package.json somehow? 

There is another PR to add component.json: more build steps, more error-prone redundancies.

Tim-Intellipharm:
Would you like me to close this pull request and create another with just the bower.json ignore updates? I can set it to something like ```["*", "!dc.*", "LICENSE"]. I can also look into creating a custom grunt task that sets the bower.json version to the same as in package.json.

mtraynham:
There is [grunt-bump](https://github.com/vojtajina/grunt-bump).  I'd be more inclined to use something that already exists than us maintain more build scripts.

gordonwoodhull:
Thanks @Tim-Intellipharm, but I can piece apart the changes easily enough.

Wow @mtraynham, grunt-bump is over the top. But I may be able to get it to do just what I want. Thanks!

Issue #927, opened 2015-04-29
using aggregated data in data table


yetanotherion:
Hello to all,

Let's take an example of 
- data = list of cars.
- car an object with attributes ("brand", "price_sold", "date_sold", "country_sold").

I'd like to display a "ranking" table that sorts the brands, from the more to the less sold.
Thanks to the group and sortBy functions, I'm currently able to make a table that displays something like
- most_sold_brand
 - ...
 - most_sold_brand, price_sold_i, date_sold_i, country_sold_i
 - ...
- second_most_sold_brand
 - ...
 - second_most_sold_brand, price_sold_j, date_sold_j, country_sold_j
 - ...
- ...

This table is however hard to read for ranking purposes, as all the *_i, *_j should be filtered manually
to have the ranking information.

What would be nice, would be to have:
- a table, like
 - most_sold_brand, number_of_most_sold_brand_cars_sold 
 - second_most_sold_brand, number_of_second_most_sold_brand_cars_sold
 - ...
- that is still refreshed by the different dimension selections performed in other charts (which I find by the way, to be an awesome feature :))

I currently don't manage to implement that, as the group in table does not 
support groups created by means of the dimension.group().reduce(add, del, init) pattern.

Do you have any suggestions on how to do so ?

AlanKavanagh:
If I am understanding you correctly then you just want to remove the headings? Why not  add a postRender event which  gets all sub  elements with class .dc-table-label and removes them? Also this is more suited for StackOverflow than the issues since it is a question not an issue with the library. Hope that helps.

gordonwoodhull:
@yetanotherion, my understanding of the request is to use a group instead of a dimension for the data for the data table. Because you want reduced values in the cells instead of raw rows.

I believe this is already supported, by accident. The coupling between dc and crossfilter is very light, so the data table is just calling `.all()` and `.top(N)` on the dimension and this also works with a group.

See this PR which depends on this: https://github.com/dc-js/dc.js/pull/697

In fact, the data table does not use "groups" like any other dc chart - instead it takes a `group` *function* which creates the values to nest by.

I delayed in answering because I wanted to create an example and update the documentation, but I don't have time at the moment. 

Could you please try this out and report if it works? (And if you are very kind, post a simple example.)

gordonwoodhull:
Not helpful to OP but just to create the link, here's an equivalent issue on crossfilter, really a documentation issue here: https://github.com/square/crossfilter/issues/150

[...]

PR #929, opened 2015-05-01
web/examples: add table-onaggregated-data

yetanotherion:
This commit shows how to create a table
on aggregated data. This currently works
even out of luck :), by setting a
dimension.group().reduce(...)
object instead of a dimension in the table.

More precisely, it works if the table's:
-  .group() is defined, and returns the same value for all elements
-  .order() is defined & defined to something else than the default value
   (which seems to be either d3.ascending or descending, a method that is
    undefined for .reduce()-d dimension groups is called else).

### DIFF
--- a/web/examples/table-on-aggregated-data.html
+++ b/web/examples/table-on-aggregated-data.html
@@ -1,0 +1,73 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>dc.js - Bar Chart Example</title>
+    <meta charset="UTF-8">
+    <link rel="stylesheet" type="text/css" href="../css/dc.css"/>
+</head>
+<body>
+
+  <table id="test">
+        <thead>
+            <tr>
+              <th>Expt</th>
+              <th>#Run</th>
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Thanks, this looks good. To clarify, this is for #927 and the idea is to also document that a data table can take a group instead of a dimension (and that it has a completely freakazoid idea of what a group is). 

I'll fix the docs at the same time I merge this.

yetanotherion:
Updated the PR to fix typo & useless modifications of index.html

gordonwoodhull:
Thanks @yetanotherion, merged for 2.0 beta 13

yetanotherion:
you're welcome :)

2015-07-30 9:59 GMT+02:00 Gordon Woodhull <notifications@github.com>:

> Thanks @yetanotherion <https://github.com/yetanotherion>, merged for 2.0
> beta 13
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dc-js/dc.js/pull/929#issuecomment-126215360>.
>



-- 
Ion

PR #930, opened 2015-05-05
Add the possibility to select specs with the command line

igncp:
Hi,

Correct if I am wrong but trying the tests I saw that each time you want to run a test with `grunt test`, you have to run all the spec files. The same happens with `grunt watch`. So, I propose this change in the Gruntfile that allows to specify the test pattern you want to run, e.g.:

`grunt test --specs spec/color-spec.js`

Thus saving time. Of course, if you don't give any option, grunt acts as usual. Hope it helps.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -102,7 +102,7 @@ module.exports = function (grunt) {
                 options: {
                     display: 'short',
                     summary: true,
-                    specs:  '<%= conf.spec %>/*-spec.js',
+                    specs:  grunt.option('specs') || '<%= conf.spec %>/*-spec.js',
                     helpers: '<%= conf.spec %>/helpers/*.js',
                     version: '2.0.0',
                     outfile: '<%= conf.spec %>/index.html',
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
The existing way to do this is `grunt jasmine --filter=color` - does this work for you?

igncp:
Great, I didn't know this feature. Right now it is throwing an exception between the coverage and browserfify phases, but I guess it should work so I'll dive more into it. Thanks!

gordonwoodhull:
Hmm, it runs okay for me, but I usually kill it before coverage finishes, because it's so slow. 

I guess the more exact command is `grunt jasmine:specs --filter-color`

Okay to close this?

igncp:
Of course, I tried the last command (the last hyphen as an equal, like the first command you mentioned) and it goes much faster than my suggestion, ~700ms to > 4 seconds (it also concatenated + uglified them), so better this way.

gordonwoodhull:
Great! Fixed the typo above. Thank you.

PR #932, opened 2015-05-07
add the css classes _0 _1 _2 _3 ... to table headers

bjelline:
add the css classes _0 _1 _2 _3 not just to td tags but also to the header th tags, 
making it possible to style all the columns.


gordonwoodhull:
Thanks @bjelline, this is more consistent and a good idea. It is potentially a breaking change for anyone who wrote selectors just on `_1` without specifying `td`, so I'll schedule it for 2.1.

PR #937, opened 2015-05-19
ES6 Modules

mtraynham:
Still packages everything into a single bundle, with `dc` as the dot notation namespace.

This PR is almost complete.  What is left to do:
- [x] Remap old globally named items for backwards compatibility in `index.js` (dc.events.trigger, dc.filters..., etc.)
- [X] Jshint errors with RangedFilter (other filters).  Wants us to `new` these as they are class notation named.
- [X] Fix bugs
- [X] Fix templates for Version as Webpack doesn't use templating.
- [x] Fix doc ordering and generation? Resolved with #978

What this brings:
- ES6 Module loading
  - Other module systems only need to package a subset of the code
  - Move away from the old Grunt concat sillyness
- ES6 in general, because why not?

Breaking changes:
- [dc.disableTransition](https://github.com/dc-js/dc.js/pull/937/files#diff-7eb52b366866677666470e019283c8eaR179), [dc.enableDebugLog](https://github.com/dc-js/dc.js/pull/937/files#diff-567859a4cc8b910de0454da79959bde6R2) and [dc.dateFormat](https://github.com/dc-js/dc.js/pull/937/files#diff-2b4ca49d4bb0a774c4d4c1672d7aa781R4) are now getter/setter's


gordonwoodhull:
This requires transcompilation right?

mtraynham:
It does through Webpack's `babel-loader`.  In this PR, the only ES6 feature is modules.  It could be replaced with CommonJS or AMD if necessary.  Either way, `window.dc` would still be available.

Webpack is nice because it's a bundler, that supports source maps and mini-fication through plugins and loaders.  

A few things are resolved by this:
- No more concatenation of header/footer code
- Browserify support is also obtained (Webpack is just nicer IMO)
- IDE support is much more intuitive.  Webstorm understands this much more than our Grunt config concatentation (I know you're on emacs though).
- jspm and future module bundlers, can install through `npm` and bundle a subset of the code.

gordonwoodhull:
So it's really about the IDE integration? The other features just seem "abstractly nicer" without delivering practical benefit. I am still wary of transcompilation because it can screw up the debugging experience, which is crucial. (We have a lot of bugs, if you haven't noticed..!) 

I admit my reluctance is largely because of some bad times with coffeescript.

mtraynham:
You're probably right, the IDE integration is probably the most impactful.  And don't get me wrong, you are not alone with the dislike of Coffeescript.  It has a small community, it's esoteric, and it writes like python.  That language is completely orthogonal to the future of JS.  In my opinion, it really serves no more purpose than being a speed bump in the road.

So I wanted to explore a module system.  There are plenty of them and I chose the one that, to me, makes the most sense.  They may be "abstractly nicer", but they do resolve problems.  Concatenating source files and uglifying the script is pretty unsafe if you ask me; especially that rely on two separate files to encase it into a code block.

jefffriesen:
+1
We've moved our apps all to Webpack & Babel ES6/7 transpiling. I love working with Webpack and I think it's a significant upgrade from Gulp and Grunt. Hot reloading and simple setup, for both transpiling and building are some reasons.

I actually like coffeescript and most of our previous codebase was in Coffeescript. The transpiling never bothered me because it was just part of the build system. We used source maps for a while, but turned them off just because it's not that hard to understand the JS generated by CS (other compile-to-JS languages that's not true). ES6/7 to ES5 is even less of a difference.  

Here's a good take on Babel that I just saw this morning: http://codemix.com/blog/why-babel-matters

mtraynham:
@jefffriesen Thanks for that article! I had no idea Babel supported that function duplication stuff.  Pretty awesome.  On your builds, did you guys completely move away from the build systems (i.e. Grunt/Gulp) or just add Webpack as a step?

In the end, I didn't want to go overboard here with a move to ES6, as I know Gordon had some reservations not to.  But it sounds like he can be convinced otherwise :smile:

jefffriesen:
@mtraynham et al,
Yes, everything is done through webpack - no grunt or gulp steps. I've seen some hybrid repos that use both but I thought it just added complexity. There are plenty of examples out there and you probably have some you've built yourself. 

Here's a repo I built for myself to get apps off the ground: https://github.com/jefffriesen/webpack-react-quickstart

It's uses React and React Router, but you could rip that out and it would work fine for a non-react app. This is all there is to the main config file: https://github.com/jefffriesen/webpack-react-quickstart/blob/master/webpack.config.js

```
var webpack = require('webpack');
module.exports = {
    entry: [
      'webpack/hot/only-dev-server',
      "./src/app.js"
    ],
    output: {
        path: __dirname + '/build',
        filename: "bundle.js"
    },
    module: {
        loaders: [
            { test: /\.js?$/, loaders: ['react-hot', 'babel'], exclude: /node_modules/ },
            { test: /\.js$/, exclude: /node_modules/, loader: 'babel-loader' },
            { test: /\.css$/, loaders: ["style-loader", "css-loader"] },
            { test: /\.less$/, loaders: ["style-loader", "css-loader", "less-loader"] },
            { test: /\.(png|jpg|ttf|woff|svg|otf|eot|svg).*?$/, loader: "file-loader" },
        ]
    },
    plugins: [
      new webpack.NoErrorsPlugin()
    ]

};
```

One thing I really like with webpack is being able to import assets like css and less into the module just like any other JS file. 

There are lots of other examples on the web. It can get a little tricky for importing unfriendly-ES6 modules. D3 and crossfilter work great. (D3 finally dropped jsdom in their dependencies!) Libraries like Vega.js I had problems with. Webpack has ways to deal with this:
http://webpack.github.io/docs/shimming-modules.html
http://stackoverflow.com/questions/28969861/managing-jquery-plugin-dependency-in-webpack/28989476#28989476 (great SO answer, btw)

But dc.js only depends on D3 and Crossfilter, so it should be really simple. 

Here's a great overview for reference: http://dontkry.com/posts/code/single-page-modules-with-webpack.html

mtraynham:
Alright, I think this is actually going to pass the tests... but it fails for me locally.  There are some test cases that use the `makeDate` function.  It generates a UTC date and the new Date's it.  

It looks like the ES5Shim provided by Jasmine is actually breaking that, because I'm East Coast (-0400) and it's giving me wrong times... But I'm thinking Travis is probably set to UTC time, because the last build didn't fail those.

index.js
```diff
@@ -1,5 +1,65 @@
 // Import DC and dependencies
 
-d3 = require("d3");
-crossfilter = require("crossfilter");
-module.exports = require("./dc");
+export * from './src/core';
```
mtraynham: Most of this is for backwards compatibility.

mtraynham:
Alrighty, since everything seems to be running accordingly (minus the [new Date(Date.UTC(...))](https://github.com/dc-js/dc.js/blob/develop/spec/helpers/spec-helper.js#L33); which I'll open a bug for), the last thing to fix is docs and ordering of those docs.

Now the generation is unchanged, but the ordering is incorrect.  I wonder if there is a way around this with docco or markdown...

gordonwoodhull:
I'm still not real eager about this. I don't want to risk breaking a library that thousands of people use, and I doubt the debugging experience is great.

Again, if you really want an ES6 dc.js, I encourage you to go ahead with your own fork. 

And if browser support is really just around the bend, then why not just wait?

gordonwoodhull:
I am excited about ES6 too, and thanks @jefffriesen for the Babel article. 

It is just that dc.js needs a mountain of debugging all the time. And for that to work, I feel that the language needs to be supported natively by the browser debugger. Unfortunately, not just one browser debugger either, but pretty much all 5 browsers.

I promise I will try out this PR and see just how bad the debugging gets. If people want to move on to a fork of dc.js I certainly understand and will support your decision.

mtraynham:
@gordonwoodhull Forgive me for the late response, got tied up with kids being sick :frowning:... freaking day cares.  

I completely understand your hesitance to move to a transpiler.  dc.js is not a simplistic library by any means, and testing code in the browser should obviously not become a harder thing to do.

Not completely related to ES6, but modules are, without a doubt, probably the best thing that has happened to JavaScript in a long time.  Breaking away from complete global scope to code that can be interpreted without interference, I consider a big win.  I always go back to the saying, "Don't modify what you don't own."   Unlike appending everything to the `dc` namespace, modules definitely provide that permissible scope.  We should distinctly import portions of the code for testing, and break apart things that violate the single responsibility principle.  At the same time, the code base would greatly benefit from an easier code reuse and abstraction pattern.

So I do implore you to try it out.  As I said before, I'm not really interested in maintaining a fork, but I also don't mind giving leeway to other alternatives.  I saw that this as a long-term goal and went with a solution that is unlikely going to change.  Frankly, I find this stuff fun to try out :smile:.

mtraynham:
Updated last step, in the process as #978 is complete and awaiting acceptance.

mtraynham:
So, I think I'm going to close this and start anew... I attempted a rebase, but so much has changed, I'm almost certain I would likely break something in the mix.  On top of this, I think using the newer d3 modules would be better suited for something like this (which none of this has).

mtraynham:
Closing in favor of #1175.

Issue #510, opened 2014-01-28
Bidirectional two-stack row chart


mmmdreg:
See [this image](http://www.insidefacebook.com/wp-content/uploads/2009/05/newinsights1.png) for an example.  
These charts are similar to stacked charts for paired data (e.g. male/female). Whereas a usual stacked chart is good at showing the _aggregate_ of two or more stacks (or the composition of the total i.e. a + b + c = total), a bi-directional chart more effectively shows the _difference_ between two stacks across different bins.

An ideal implementation would use the same scale on both halves and be implemented as a stacked row chart  (issue [#397](https://github.com/dc-js/dc.js/issues/397)) with 2 and only 2 stacks, with an option 'bi-directional' which only takes effect if there are only 2 stacks.

In the meantime, it may be simulated by firstly allowing the regular row chart to be flipped horizontally (i.e. vertical axis and origin move to the right side of the graph) and then displaying the graph adjacent to a regular row chart.

moshequ:
@jefffriesen Your jsfiddle doesn't work anymore I've created a working version http://jsfiddle.net/moshequ/RWs67/111/

jefffriesen:
@moshequ cool, thanks for updating it.

gordonwoodhull:
Cf. #1053 #1068

PR #943, opened 2015-06-04
Paired Row Chart and Row Chart useRightYAxis

ruhley:


gordonwoodhull:
Coooool. One could implement #510 bidirectional row chart with this.

ruhley:
I am working on a Paired Row Chart like [https://reference.wolfram.com/language/ref/PairedBarChart.html](https://reference.wolfram.com/language/ref/PairedBarChart.html) if that is what you mean by a bidirectional row chart.

This is a precursor to getting the paired row chart (or whatever you want to call it) to work.

ruhley:
I have just noticed that the animation when rendering is still going from right to left. If you hold off for a bit I can get that to work from left to right as well.

gordonwoodhull:
Yeah exactly, see the link to #510 above. Usually used for gender vs age distribution.

Thanks for the row example, that's been missing.

I'm going to have to wait for tests (nudge, or write them myself), so no rush...

ruhley:
I have fixed up the animation for row charts that are right aligned and also added tests. How does it all look?

ruhley:
@gordonwoodhull I have implemented the paired row chart, included examples and tests. I feel it is a little hacky by having two sub row charts, but all in all it works great for me. I had to change the row chart slightly to get it to work.

Take a look at the [example](https://github.com/dc-js/dc.js/pull/943/files#diff-fcc50b70a2594460c1d2f27a1091a824) for how it works.

gordonwoodhull:
Ooops, usually the best practice is to create a feature branch when contributing a PR. This has a ton of other changes in it now. I think I can sort through and find the changes that are relevant to this feature. 

Sorry for the extreme lag in merging this! It's hard to find the time for a careful review.

ruhley:
Yea, I realised that I probably should have done that a bit late, which is why I updated the [changelog](https://github.com/ruhley/dc.js/blob/develop/Changelog.md#ruhley-changes), so that anyone could find the commits for each feature easier.

gordonwoodhull:
Yes, that is very helpful! Probably what needs to happen next is for someone (could be me or anyone with decent git fu) to create a PR with just those changes. I have to review changes mostly independently.

gordonwoodhull:
Replaced by #1068, with just the relevant commits.

PR #946, opened 2015-06-10
Added ability to slice data-grid results for pagination; fix data-grid docs

calvino:
There was no way to create pagination for datagrid results with the current API. This was a huge feature request as grid items are based on html and with large datasets dc.js would be rendering a huge number of DOM nodes that were not relevant to the user. Using .top() to filter was not enough as most users expect to be able to see the next available page.

Because of this limitation I added the ability to specify a slicing range. I am slicing the entries array to achieve this. With the beginSlice and endSlice attributes it is fairly simple to implement pagination using datagrids. Updating these parameters and then redrawing the graph also for very simple and efficient pagination, eliminating the need to draw possible hundreds of thousands of DOM nodes.

I also added a comment that notes the difference between how dataGrid uses .group() versus the other charts. DataGrid actually wants a keying function, not a crossfilter function. This was very confusing to myself and other developers, so I explicitly call out this discrepancy in the documentation.

### DIFF
--- a/spec/data-grid-spec.js
+++ b/spec/data-grid-spec.js
@@ -63,6 +63,24 @@ describe('dc.dataGrid', function() {
         });
     });
 
+    describe('slicing entries', function() {
+        beforeEach(function() {
+            chart.beginSlice(1);
+            chart.redraw();
+        });
+
+        it('slice beginning', function() {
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Thanks @calvino! Merged for 2.0 beta 16.

PR #948, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3


gordonwoodhull:
Nice, one less thing to screw up when doing a release!

kkirsche:
Yep :+1:

gordonwoodhull:
Thanks @kkirsche, merged for 2.0 beta 13

PR #950, opened 2015-06-20
API doc header fix

ahammer-:


gordonwoodhull:
Thanks @ahammer- these docs are generated, and I guess a newline must have gotten lost at some point.

PR #953, opened 2015-06-25
fix calculateBarWidth

nreese:
The function calculateBarWidth calls xUnitCount() to populate the variable numberOfBars. When using a non-ordinal scale, xUnitCount does not return the number of bars. Instead, xUnitCount returns the range of the scale. This causes all of the math with the gap and barPadding values to be funky. 
The pull request changes the function calculateBarWidth to take a parameter specifying the number of bars in the layer. All of the demos work great but the changes break 12 tests and I am not sure how to fix those.

### DIFF
--- a/src/bar-chart.js
+++ b/src/bar-chart.js
@@ -63,8 +63,6 @@ dc.barChart = function (parent, chartGroup) {
         var layers = _chart.chartBodyG().selectAll('g.stack')
             .data(_chart.data());
 
-        calculateBarWidth();
-
         layers
             .enter()
             .append('g')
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Hi @nreese, glad you found a fix for your use cases. The errors in the tests indicate this doesn't work in all cases. 

In general, not all the data will necessarily be shown, for one thing. So it's necessary to look at the bars that are within the domain of the scale. And running the stock example (the main page), the Days by Fluctuation chart has bars overlapping each other.

Hopefully your investigation will us narrow in on a solution that works in all cases.

gordonwoodhull:
I'm not convinced this doesn't indicate a misuse of `.xUnits()` - it should return the number of bars which should fit at the current scale. Granted, this is not necessarily the same as the amount of data. And I know there are problems with bar width, but I don't think this is the solution.

I know this PR is ancient, but if you're still having issues @nreese, please file a more detailed issue (or comment with a link to previous conversations on this issue). As always, a fiddle or block demonstrating the issue really speeds things up. Thanks!!

Issue #928, opened 2015-04-30
getting chart.data() might fail


Sumbera:
in rare cases where there are no data nor group/dim yet specified (and graph 'skeleton' have to be rendered, for example by custom subclass), retrieving chart.data() will fail on group.all as group is undefined. (but should return null). 

dc.baseMixin :

```
 var _data = function (group) {
     return  group.all();
};
```

should be :

```
 var _data = function (group) {
    return group ? group.all(): null;
};
```

gordonwoodhull:
Sorry for the slow reply.  

All the charts expect to be completely initialized before they are rendered, so I'm a little hesitant to accept this change, because it might mask other bugs. (The common case where people think they have properly initialized, for example.)

Sumbera:
no problem,thanks for explanation. I found it as issue on my case where (custom chart) choropleth map (or possibly any other geodata chart) is first loaded (as geojson) and rendered with no values (greyed), just keys and   geometry and once other  (attribute) data arrives asynchronously, it sets up a dimension and group and  redraws the rest. So in my case I prefer progressive rendering and not wait until after all data are available.

gordonwoodhull:
Hmm, yes there are many inconsistencies in dc.js. I'd consider changing this policy but I'd want a comprehensive set of tests for it, i.e. a test for every chart that it renders without crashing if the dimension and group are null.

I can see how this would be useful for getting charts up on the screen quickly, while waiting for the data to load.

gordonwoodhull:
I don't see how this would get past the mandatory attributes test, which makes sure that `.group()` has been specified. If anyone wants this, I'd suggest faking the group until the data is set:

```js
var emptyGroup = {
  all: function() { return [] }
};
chart.group(emptyGroup)
```

I like this better because it's explicit what's going on. Yes, it's a little awkward, but I'd suggest having better data tools for dc.js, instead. In this case an `empty_group()` function would fit the bill.

PR #959, opened 2015-07-13
fix group.all()  if group is null

Sumbera:
fixing issue:  getting chart.data() might fail #928


gordonwoodhull:
Rationale for the current behavior and a workaround in #928

PR #965, opened 2015-07-27
Customizable Legend Name Function

renoth:
This would be useful


renoth:
nevermind it is not well thought out.

PR #968, opened 2015-07-30
Fix merge issue

austinlyons:


gordonwoodhull:
Ahhh, only the uglified libraries and maps were broken, which we don't test, as well as the docs, which are published from master. Need. More. Tests.

gordonwoodhull:
Thanks for the heads up @austinlyons!

PR #975, opened 2015-08-07
Update dependencies to latest, fix spec linting errors

mtraynham:
Updates all npm dependencies to latest.  Update jscs & jshint options.


.jshintrc
```diff
@@ -1,28 +1,68 @@
 {
-    "browser": true,
+    "bitwise": false,
```
mtraynham: FYI, we do have bitwise operators in portions of the code.  Turn this off to find them...

.jshintrc
```diff
@@ -1,28 +1,68 @@
...
-    "sub": true,
+    "noempty": true,
+    "plusplus": false,
     "undef": true,
     "unused": "vars",
-    "globals" : {
+    "strict": false,
+    "maxparams": 5,
+    "maxdepth": 4,
+    "maxstatements": false,
```
mtraynham: Some files exceed 100 max statements...

mtraynham:
Not sure what your preference is on "`else` on separate lines"

```
if (...) {
} else {
}
```
vs
```
if (..) {
}
else {
}
```
but I corrected some of those cases.

mtraynham:
One last thing, I have to recommend this global npm dependency, because it works very well.

https://www.npmjs.com/package/npm-check-updates

At the CLI:
`npm-check-updates` or `ncu` lists all upgradable packages
`ncu -u` updates the package.json
Then call `npm update` to get the new depedencies

mtraynham:
Alrighty, I fixed almost all of the linter errors in `./spec`.  That was achieved with the new jscs CLI options `jscs --fix spec/**.js`.  It isn't hooked up in the Gruntfile though.

.jshintrc
gordonwoodhull: interesting. all of those look legitimate except for this one:

```{js}
   src/data-table.js
    106 |            bAllFunctions = bAllFunctions & (typeof f === 'function');
```
which i'll fix when i merge this for the next beta release. it's harmless, but one still shouldn't use bitwise for booleans.

then there's the annoying but apparently very fast [`n  | 0`](http://jsperf.com/javascript-trunc)...

gordonwoodhull:
Neat, thanks! I'm not crazy about `} else {` but I can live with it.

.jshintrc
mtraynham: http://jsperf.com/floor-vs-parseint-vs-bitwise

25% faster than Math.floor, but likely negligible performance hit.

gordonwoodhull: you're right, it's probably not helping anything. our bottleneck is the DOM, not math.

mtraynham:
Last update fixes the remaining issues with the spec directory and adds /spec to the jscs/jshint paths to lint.  I want to point out as well, we have a script at `./spec/helpers/jasmine-jsreporter.js`.  It is a duplicated copy of the file at https://github.com/detro/jasmine-jsreporter/blob/master/jasmine-jsreporter.js, but it's out of date.

I'd like to get this removed from the source and retrieved using npm instead.

gordonwoodhull:
Can't find an npm module for that - is there a standard, safe way to pull sources from github?

It seems to be used by grunt-saucelabs, which just says pull it from github.

https://www.npmjs.com/package/grunt-saucelabs#test-result-details-with-jasmine

gordonwoodhull:
Actually, there seem to be two copies within grunt-saucelabs itself:

```
-rw-r--r--  1 gordonwoodhull  staff  12355 May 30  2014 node_modules/grunt-saucelabs//examples/jasmine/lib/jasmine-jsreporter/jasmine-jsreporter.js
-rwxr-xr-x  1 gordonwoodhull  staff   6468 Jan  3  2014 node_modules/grunt-saucelabs//test/jasmine/lib/jasmine-jsreporter.js
```

mtraynham:
Ya know for the life of me, I can't figure out how to install it directly from github...  I for certain have done this before.  I'll create a PR on that repo for a package.json and a request publish to npm...

gordonwoodhull:
I think it's probably fine to use the copy from saucelabs, since that's what we're using it for.

mtraynham:
sounds good to me

PR #977, opened 2015-08-11
JSDoc migration

mtraynham:
A WIP PR for the JSDoc migration.  It uses [jsdoc-to-markdown](https://www.npmjs.com/package/jsdoc-to-markdown), which unlike the standard jsdoc suite, only generates a single file.  I went ahead and migrated the bar chart as a quickstart.  There are some things that still need to be resolved:

- Table of contents
- Namespacing

I added a grunt watch task to easily see the differences between each save.

Example output: https://github.com/mtraynham/dc.js/blob/jsdoc/web/docs/api-latest.md

May need to revise some of the tags, JSDoc likes the use of classes, so I'm still trying to understand what all is available.


mtraynham:
Well, I didn't change that many files... I wonder if there was an issue when I merged.

gordonwoodhull:
Looks like you need to update to latest master, it's about 5 betas behind...

mtraynham:
Sounds good, I'm about to commit the `baseMixin.js` documention changes.  I'd like to stop at that point and get some feedback, figure out what you like/dislike.

mtraynham:
Actually, I only did a few functions.  Let me know what you think, and I'll continue.  The nice part of this, there's no real issues of ordering the files here, you just tell it who the parent function/object is. I think it would be more straight-forward if we did use a prototypal OOP structure, but this works too.

mtraynham:
Closing this, my rebase got jacked up because of some stashed changes.

PR #979, opened 2015-08-11
Alternate bar chart implementation and layer-mixin

alan-unravel:
courtesy of @mtraynham


mtraynham:
Thanks for putting this together @alan-unravel.  Just wanted to post the jsfiddle's for some additional resources:
Bar - http://jsfiddle.net/st7n2ocs/
Line - http://jsfiddle.net/dL5o2ygc/

alan-unravel:
No problem @mtraynham, I'm going ahead with the short term solution now, but this should probably be left open with a view to bring it in in v3 maybe ?

PR #982, opened 2015-08-11
Added ability to render custom legend text per item

calvino:
I needed the ability to create legends with more data than just the group name so I edited the legends widget to support that functionality. You can now dynamically generate legend text much like you would for labels or titles. You pass in a legendText function that returns the text you want to render, e.g.  chart.legend(dc.legend().legendText(function(d) { return d.name + ': ' + d.data })) to create labels with counts on them if you have grouped data reduced by count.


gordonwoodhull:
Thanks @calvino!

gordonwoodhull:
Merged in 2.0.0 beta 20

PR #984, opened 2015-08-12
Grouped visualization for stacked bar charts [WIP]

alan-unravel:
Allows for passing of a renderType for bar-charts rendering too a grouped bar chart if "group" or a stacked bar chart if "stack"

### DIFF
--- a/dc.js
+++ b/dc.js
@@ -4195,8 +4195,10 @@ dc.barChart = function (parent, chartGroup) {
     var _gap = DEFAULT_GAP_BETWEEN_BARS;
     var _centerBar = false;
     var _alwaysUseRounding = false;
-
+    
     var _barWidth;
+
+    _chart.renderType = "stack";
 
     dc.override(_chart, 'rescale', function () {
         _chart._rescale();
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
@Sid0403, @MsJR, I've also created a branch for this alternate candidate implementation for grouped bars. 

Branch: https://github.com/dc-js/dc.js/tree/kavanagh-grouped-bar
Zip file: https://github.com/dc-js/dc.js/archive/kavanagh-grouped-bar.zip
Example: https://github.com/dc-js/dc.js/blob/kavanagh-grouped-bar/web/examples/grouped-bar.html

The interface is slightly different: you set `.renderType('group')` instead of `groupBars(true)`, and there doesn't seem to be a way to set the gap between groups. [Documentation](https://github.com/dc-js/dc.js/blob/kavanagh-grouped-bar/web/docs/api-latest.md#dc.barChart+renderType)

I notice that bar labels (which were implemented after this PR) don't work - they should be easy to fix but I couldn't figure it out with a quick look.

Trying out this implementation and reviewing it here will be of great help. We'll merge the best features from each of the implementations.

sunpcm:
I don't it is a bug or not,When I change the barChart by a pieChart,the layer don't change. for example,when I click the pie ,the layer should changed from 15.6% to 45.% ,but ,It dosn't work.However,when I remove '.renderType('group')',it will be fine.But , Can you help me?Best wish!

PR #985, opened 2015-08-13
Y Axes alignment option

gazal-k:
When there are two y axes and at least one of them has values in the negative range, the 0 level on both axes don't line up.
Having an option to line them up seemed like a nice to have in my opinion.

### Example:

Without Setting alignAxes option:

screenshot from 2015-08-13 10 01 59

Setting alignAxes option: 
```js
chart.alignAxes(true);
```
screenshot from 2015-08-13 10 03 40

### Pending
 - tests
 - docs (couldn't find how to generate the docs)


gordonwoodhull:
I agree this is a good feature. However, I find the implementation excessively complicated.

 * I think you can implement this with much simpler, much less code by adapting the two places where `yAxisMin`, `rightYAxisMin`, `yAxisMax`, `rightYAxisMax` are *used*, rather than changing how they are *defined*.
 * You have made `yAxisMin` and `rightYAxisMin` mutually recursive, each one calling the other with the alignment flag implicitly false. This is inefficient, because all the data will be evaluated twice. I also find this implementation confusing.
 * At least to me, it seems more intuitive to leave the meaning of these functions alone, to let them still return the actual minimum and maximum for each axis. Then, after each has been calculated, adjust all four values and set the ranges for both left and right axes. Of course this means combining `prepareLeftYAxis` and `prepareRightYAxis`, which is fine, because they are only ever called at the same time, from `_prepareYAxis`. The combined function could be `prepareYAxes(left, right)`, where `left` and `right` are boolean parameters.

Also, this definitely needs tests in spec/composite-chart-spec.js, showing that the axes are not aligned with `alignedAxes` false, and are aligned with `alignedAxes` true. IMO the flag should also be called `alignYAxes`.

gazal-k:
@gordonwoodhull sorry about that. I'd been using the composite chart in an application we're developing. Originally this axes alignment feature was developed outside `composite-chart`'s source which actually was more optimal. I kinda messed it up when I tried to put that into `composite-chart` code.
So, now I've used the original implementation and introduced it into composite-chart as per your recommendations.
Just a minor difference though, because of cyclomatic complexity, I'd to refactor it again. There's no unified `prepareYAxes` function. Instead the calculation for axes alignment is done in a separate function and its result fed to `prepareLeftYAxis` and `prepareRightYAxis`.

I haven't added tests in spec/composite-chart-spec.js yet. Before that, please let me know if the refactor is ok.

dc.js
```diff
@@ -6307,8 +6308,11 @@ dc.compositeChart = function (parent, chartGroup) {
     };
 
     _chart._prepareYAxis = function () {
-        if (leftYAxisChildren().length !== 0) { prepareLeftYAxis(); }
-        if (rightYAxisChildren().length !== 0) { prepareRightYAxis(); }
+        var left = (leftYAxisChildren().length !== 0), right = (rightYAxisChildren().length !== 0),
+        ranges = calculateYAxesRanges(left, right);
```
gordonwoodhull: Needs `var` otherwise `ranges` will end up in the global scope.

gordonwoodhull:
Ah, this looks much more clear! And now I understand why you took the prior approach, modifying these classes from outside is indeed tricky. I like that the new approach doesn't change existing interfaces.

This can go into the 2.0 branch, once you add tests. Thanks!

dc.js
gazal-k: @gordonwoodhull looks like I mislead you with the indentation. I'll just change the multivar declare for better readability.

gordonwoodhull:
Duh, right, fooled by the indent. Thanks!

gordonwoodhull:
Hi @gazal-k.

Thanks again! I'm currently merging this. The design and code structure is very good, but I'm not sure about the actual calculation. It can cause one of the axes actually to flip over, in the simple Jasmine test case I built where one of the charts is all negative.

Could you explain the rationale behind taking the ratio of minimum and maximum on one axis, and then applying this ratio to the other axis? It seems like it does cause the zero lines to match up, but might not  always keep both plots with the chart boundaries or properly oriented.

gordonwoodhull:
Merged in 2.0.0 beta 20

gazal-k:
@gordonwoodhull sorry for the delayed reply. I don't remember exactly how I applied the math for it. Let me try the test you mentioned.

gazal-k:
@gordonwoodhull thanks for that. I had totally missed that case. And sorry that I made you write the tests :cry:
#1033 should fix the issue. Please check and let me know

gordonwoodhull:
Note: I think it was implied that `elasticY` must be true to enable this feature. In the first pass we didn't account for this, but that caused breakage in #1056. I'm enforcing that now.

In other words, you can't manually apply the Y domains and use this feature, because this feature expects to set the Y domains for you.

gazal-k:
@gordonwoodhull sorry about that. so alignYAxes should have no effect when elasticY is false, correct?

Shall make the change here: https://github.com/dc-js/dc.js/pull/1033

gordonwoodhull:
You agree this is correct, right? I think I have fixed it in the commit referenced above. 

It may have been me who broke this in the first place, since I made some additional changes after testing your code. 

Please review the fix and then document this requirement in #1033. Thanks!

gazal-k:
Hadn't noticed https://github.com/dc-js/dc.js/commit/3d13e68cf35f56879828148f39f07c79f4748175.
But that only changes the spec, right? I didn't see any change in https://github.com/dc-js/dc.js/blob/develop/src/composite-chart.js that limits alignYAxes from being applied when elasticY is false.

And yes, I do agree that alignYAxes need only be applied when elasticY is true.

gordonwoodhull:
Whoops I linked one commit but not the other. Here is the fix: d1c087dcaae5babb7a290ce86eeb772b057d8a82

PR #986, opened 2015-08-13
added simple path to external labels functionality on pie charts

alan-unravel:
Requires externalLabels to have a value. Should probably also require radius otherwise the labels will get cut off very frequently. Not as fancy as the initial example, will have to work on that
pathed_pie


dc.js
```diff
@@ -3884,9 +3885,46 @@ dc.pieChart = function (parent, chartGroup) {
...
+        .attrTween('points', function(d){
+          this._current = this._current || d;
+          var interpolate = d3.interpolate(this._current, d);
+          this._current = interpolate(0);
+          return function(t){
+            var centroid, centroid2;
+            centroid = d3.svg.arc()
+                .outerRadius(_radius + _externalLabelRadius)
+                .innerRadius(_radius + _externalLabelRadius)
+                .centroid(d);
```
gordonwoodhull: `centroid` is never used.

gordonwoodhull:
Thanks @alan-unravel, I am merging this and making a few improvements:
 - implementing the transition cases
 - removing the unused centroid noted above
 - applying the arc to the interpolated angle rather than the final angle
 - making lines to zero-sized wedges invisible

As you can see on this test page, there are many problems with pie transitions:
http://dc-js.github.io/dc.js/transitions/pie-external-label-transitions.html
.. but these path transitions are mostly better behaved at this point.

I think the position of the line could still be improved. Right now it's drawing from the center of the wedge to the center of a ring segment with the label on the outer radius. IMO it would be nicer to draw it directly to the label, but clip it at the label or a slight margin away from the label. I have code for that elsewhere but this is good enough for now.

gordonwoodhull:
Merged in 2.0.0 beta 20

alan-unravel:
Thanks @gordonwoodhull let me know if I can be of any help, the original lines from the example I put in gitter were nice but I didn't have the time to generalize the concept.

Issue #989, opened 2015-08-18
Default filterHandler is inefficient


esjewett:
This is the default filterHandler in the base mixin:

```
var _filterHandler = function (dimension, filters) {
        dimension.filter(null);

        if (filters.length === 0) {
            dimension.filter(null);
        } else {
            dimension.filterFunction(function (d) {
                for (var i = 0; i < filters.length; i++) {
                    var filter = filters[i];
                    if (filter.isFiltered && filter.isFiltered(d)) {
                        return true;
                    } else if (filter <= d && filter >= d) {
                        return true;
                    }
                }
                return false;
            });
        }
        return filters;
    };
```

Every time a filter is applied to a chart, the current filter on the dimension is removed and then a new one is added. This is extremely inefficient. I believe it renders all of the benefit of Crossfilter's optimization for incremental filter change moot. Instead, the filter should be removed *or* a new filterFunction should be applied. Depending on the filter type, we could even optimize further and use dimension.rangeFilter or dimension.exactFilter.

Can we start by removing the first line where ```dimension.filter(null);``` or is this doing something somewhere that I don't understand?

gordonwoodhull:
That makes sense. Someone must have thought that filters were cumulative at some point.

Please try it out and run the test suite. Hopefully that will catch any issues. The suite is extensive but never complete.

Related: #478, not using range filters is also shredding crossfilter performance.

esjewett:
Digging in a little now. Hopefully I can move to using filterExact or filterRange when appropriate as well.

esjewett:
BTW - which branch should I send a pull request against?

gordonwoodhull:
Thanks @esjewett, for straight bug fixes it's master/2.0. This doesn't present any interface changes, and I don't think it's very risky because any problems will be easy to detect.

esjewett:
For some reason the filterRange part of this fails tests, but can't see why. I will submit a pull request with just the removal of the extra null filter and the filterExact part.

```
if (filters.length === 0) {
            dimension.filter(null);
        } else if (filters.length === 1 && !Array.isArray(filters[0])) {
            dimension.filterExact(filters[0]);
        } else if(filters.length === 1 && Array.isArray(filters[0])
            && filters[0].isFiltered
            && typeof filters[0][0] === typeof filters[0][1]) {

            dimension.filterRange(filters[0]);
        } else {
            dimension.filterFunction(function (d) {
                for (var i = 0; i < filters.length; i++) {
[...]

PR #990, opened 2015-08-19
Issue #989 - filter handler is inefficient

esjewett:
Addresses issue #989 by removing the unnecessary and inefficient initial filter(null) and introduces logic for using filterExact in situations where it will work.


esjewett:
I suspect I screwed up this pull request. Anything I can do to help it be merged? Back out and redo without the generated files?

gordonwoodhull:
Don't worry about it, I can handle that as long as the changes to the non-generated files are all intentional.

I just was in a rush before vacation and didn't have time to review and see whether more tests are necessary, etc. I'm back next week and hope to merge then.

My turnaround tends to be frustratingly slow because I try to seriously review everything. There was a time when stuff was getting merged in too fast and it led to a lot of bugs, some still not fixed. :-(

esjewett:
Sorry, didn't realize you were on vacation! Enjoy :-) And +1 for reviewing thoroughly. I'm still not completely clear on all the different filter structures that various dc.js components use, so it would be good if you can take a second look and make sure this won't break anything that's not in the test suite.

gordonwoodhull:
Wow, this "feels" way, way faster on the main demo page! (Admittedly, I haven't benchmarked.)

Next I'll see if I can correctly detect dc.filters.RangedFilter

Issue #554, opened 2014-03-31
.title not working on composite of line charts


xavier-d:
According to the example provided by the documentation https://github.com/dc-js/dc.js/blob/master/web/docs/api-latest.md#composesubchartarray title method should work on Composite Chart.

I've updated the example provided:
https://github.com/dc-js/dc.js/blob/master/web/examples/composite.html

.compose([
            dc.lineChart(composite)
                .dimension(dim)
                .colors('red')
                .group(grp1, "Top Line")
                .dashStyle([2,2])
                .title(function(d) { return 'Custom function prints this\n' + d.value; }),
            dc.lineChart(composite)
                .dimension(dim)
                .colors('blue')
                .group(grp2, "Bottom Line")
                .dashStyle([5,5])
                .title(function(d) { return 'Custom function prints this\n' + d.value; })
            ])

and it is not working 

A bug ? A limitation of composite chart ? If yes the documentation should be updated

RickWalker:
I get this problem too, on series charts - titles are fine when I create the chart, but don't update when I filter.

RickWalker:
Actually, turns out I was on 2.0.0-alpha1 and it's fine on 2.1.0-dev. I'll make more effort to stay up to date!

Kylir:
Thanks @jeenu6 ! I had the same issue and set `chart.shareTitle(false)`. I'm using the master branch.

DougMair:
I had the same problem and found a solution.

Make sure to call .shareTitle(false) BEFORE calling the .compse( ...) method.

daniel-gall:
Submitted pull request for this issue:

https://github.com/dc-js/dc.js/pull/1198

gordonwoodhull:
I agree, we should document it. I wouldn't upgrade a bug to a feature, though. [Looks like](https://github.com/dc-js/dc.js/blob/71ed43213d4068ab395410a60337d4d3a5c93180/src/composite-chart.js#L341) we'd need to document this for `width`, `height`, `margins`, and `childOptions` as well, unfortunately.

Action items for this issue:
 * Fix compose not to immediately replicate settings from parent to children (this should happen in render/redraw instead)
 * Figure out whether `.shareTitle(true)` ever works, and how, or fix it so that it has a better chance of working

gordonwoodhull:
Fixed by #1365 in 3.1.5

PR #995, opened 2015-08-25
Fix heatmap x axis label position on redraw

crazycodingcat:
Small issue with heatmap: when heatmap is updated using _doRedraw, the 'y' position of x axis label does not get updated in transition. E.g. If I resize the heatmap and make it taller, the x axis label should move down.


gordonwoodhull:
This is great. Resizing of charts is definitely a work in progress; this was previously reported in #554

PR #999, opened 2015-08-27
Update all dc.mixin docs

mtraynham:
This adds a lot of missing documentation to the baseMixin.  Added default implementation examples, `@see` links (notably to other functions of interest & crossfilter/d3/HTML/SVG docs), and after a review of the current JSDoc implementation, updates references that returned `Chart` to just return `dc.baseMixin`.  Using `Chart` seems a bit unnecessary, since we should just be linking back to the `dc.baseMixin` (plan to update the other mixin/charts with this as well).

Depending on timeframe, I may get to the other mixins/charts this week as well, so I may rename the title and the changeset will grow.  

Interesting to document some of these things I didn't know existed or only work under some circumstances; guess I just don't use some of these things that often or take the functionality for granted...


mtraynham:
woops reopen.

gordonwoodhull:
Thanks @mtraynham!

PR #1001, opened 2015-08-28
Fix row chart x-axis position on resizing

crazycodingcat:
Another small problem in resizing:
When row chart is resized and redrawn, the x-axis position should get updated accordingly.

### DIFF
--- a/src/row-chart.js
+++ b/src/row-chart.js
@@ -66,9 +66,9 @@ dc.rowChart = function (parent, chartGroup) {
         calculateAxisScale();
 
         if (axisG.empty()) {
-            axisG = _g.append('g').attr('class', 'axis')
-                .attr('transform', 'translate(0, ' + _chart.effectiveHeight() + ')');
+            axisG = _g.append('g').attr('class', 'axis');
         }
+        axisG.attr('transform', 'translate(0, ' + _chart.effectiveHeight() + ')');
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Thanks @crazycodingcat - is there a reason you don't want the moving axis to be animated/transitioned as in your last fix?

crazycodingcat:
Yes, the moving axis should be animated/transitioned.  

I'm making a UI where user can drag the corner of the chart to resize it, and found this bug with the row chart - the x-axis position does not get updated when the height of the chart is changed. I'm using dc.redrawAll() to redraw the charts after the width and height are changed (not sure if this is the best way to do it though). So for the row chart problem, I found this PR as a quick fix without changing too much of the current implementation..

gordonwoodhull:
Cool, I'll just move the line to the transition part, as you implemented the other one and as I implemented this for bar charts. Thanks!

gordonwoodhull:
Oh, I see, actually the changing of the height of rows is not animated, so it's probably better not to animate the axis move for now. It's different for the coordinate-grid based charts [like the bar chart](http://dc-js.github.io/dc.js/resizing/resizing-bar.html).

Taking this as-is for now, but it would be nice to animate the height of rows the way the width of bars is already animated.

PR #1012, opened 2015-09-19
Fix non-mixin docs

mtraynham:
This updates the rest of the docs to be in line with the mixin docs.  Lots of links have been added/corrected.  We no longer use the Markdown notation for links, instead using the jsdoc notation.

### DIFF
--- a/src/bar-chart.js
+++ b/src/bar-chart.js
@@ -2,8 +2,8 @@
  * Concrete bar chart/histogram implementation.
  *
  * Examples:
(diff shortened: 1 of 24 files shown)


mtraynham:
Here's the documentation after this PR:
https://github.com/mtraynham/dc.js/blob/fix_chart_docs_and_links/web/docs/api-latest.md

mtraynham:
Hopefully this is the last PR for docs :smile:

gordonwoodhull:
Thanks @mtraynham!

PR #1017, opened 2015-09-24
linechart-per-dimension: one line chart per one dimension attribute

yetanotherion:
Create one line chart per experiment with morley.csv data.

Are there suggestions on how to implement that in another way ?

### DIFF
--- a/web/examples/chart-per-dimension.html
+++ b/web/examples/chart-per-dimension.html
@@ -1,0 +1,57 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>dc.js - Chart per dimension Example</title>
+    <meta charset="UTF-8">
+    <link rel="stylesheet" type="text/css" href="../css/dc.css"/>
+</head>
+<body>
+
+<div id="exp_1"></div>
+<div id="exp_2"></div>
+<div id="exp_3"></div>
+
+<script type="text/javascript" src="../js/d3.js"></script>
+<script type="text/javascript" src="../js/crossfilter.js"></script>
+<script type="text/javascript" src="../js/dc.js"></script>
+<script type="text/javascript">
+
+function createChart(number) {
+  return {id: number,
+          chart: dc.lineChart('#exp_' + number)};
+};
+
+var charts = [1, 2, 3].map(createChart);
+d3.csv("morley.csv", function(error, experiments) {
+
+  experiments.forEach(function(x) {
+    x.Speed = +x.Speed;
+    x.Run = +x.Run;
+    x.Expt = +x.Expt;
+  });
+
+  var ndx                 = crossfilter(experiments),
+      expDimension        = ndx.dimension(function(d) {return d.Expt;});
+
+  charts.forEach(function (c) {
+    var xf = crossfilter(expDimension.filter(c.id).top(Infinity)),
+        runDimension = xf.dimension(function(d) {return d.Run}),
+        speedGroup   = runDimension.group().reduceSum(function(d) {return d.Speed;});
+    c
+    .chart
+    .width(768)
+    .height(480)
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Hi @yetanotherion - thanks for the contribution. 

It depends what you're trying to do. :-)

If you want the charts to react to each other then they should be on the same crossfilter.

gordonwoodhull:
Here is a simpler way to get the same result as you got, without filtering:

```{js}
  var ndx = crossfilter(experiments);
    runDimension = ndx.dimension(function(d) {return d.Run}),

  charts.forEach(function (c) {
    var speedGroup   = runDimension.group().reduceSum(function(d) {
        return d.Expt === c.id ? d.Speed : 0;
    });
    c
    .chart
    ...
```

gordonwoodhull:
And then simply move the dimension into the loop, so that each chart has its own, in order for them to observe filtering:

```{js}
  var ndx = crossfilter(experiments);
  charts.forEach(function (c) {
    var runDimension = ndx.dimension(function(d) {return d.Run}),
        speedGroup   = runDimension.group().reduceSum(function(d) {
            return d.Expt === c.id ? d.Speed : 0;
        });
    c
    .chart
```

yetanotherion:
Thanks !
That would be the ideal solution, (I have bug on filtering updates
with the other solution).

However, after applying it in the 'real' example, it seems I'm reaching
something like a maximum number of dimensions, on the same crossfilter
object.
"
Error: invalid array width!"
(I have 39 objects in the list, and getting the error after creating
the 22th dimension (there are other dimensions not related to the list))

Is there such a limit ? And a way to workaround it :) ?



2015-09-24 14:16 GMT+02:00 Gordon Woodhull <notifications@github.com>:

> And then simply move the dimension into the loop, so that each chart has
> its own, in order for them to observe filtering:
>
>   var ndx = crossfilter(experiments);
>   charts.forEach(function (c) {
>     var runDimension = ndx.dimension(function(d) {return d.Run}),
>         speedGroup   = runDimension.group().reduceSum(function(d) {
>             return d.Expt === c.id ? d.Speed : 0;
>         });
>     c
>     .chart
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dc-js/dc.js/pull/1017#issuecomment-142912394>.
>



-- 
Ion

gordonwoodhull:
Right, there is a limit of 32 dimensions with crossfilter. You can try the community fork of crossfilter v2.0-alpha.02 if you really need more dimensions.

I'm still not really clear what you're trying to do. If you don't need these charts to filter each other, you can also define your groups based on a single dimension.

gordonwoodhull:
For completeness sake, here's probably the most efficient approach, building a single group with all the values and then using valueAccessors to draw each chart:

```{js}
  var ndx = crossfilter(experiments),
      runDimension = ndx.dimension(function(d) {return d.Run}),
      speedGroup   = runDimension.group().reduce(
          function(p, v) {
              p[v.Expt-1] = (p[v.Expt-1] || 0) + v.Speed;
              return p;
          },
          function(p, v) {
              p[v.Expt-1] -= v.Speed;
          },
          function() {
              return [];
          });

  charts.forEach(function (c) {
    c
    .chart
    ...
```

But it appears this is really a support request, not a contribution, so I'm closing this. Can we continue this conversation on the user group or on SO?

yetanotherion:
Your last solution worked. Sorry for the spam I'll try first
stack overflow over pull request next time.



2015-09-24 15:46 GMT+02:00 Gordon Woodhull <notifications@github.com>:

> Closed #1017 <https://github.com/dc-js/dc.js/pull/1017>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dc-js/dc.js/pull/1017#event-418111113>.
>



-- 
Ion

PR #1018, opened 2015-09-25
Remove grunt-debug-task

mtraynham:
This removes the grunt-debug-task npm dependency:

Reasons:
- Not used within the codebase
- [Deprecated](https://www.npmjs.com/package/grunt-debug-task) - suggests to use node-inspector directly
- Prevents contributors from using Node V4 (caused by an old Nan dependency at [node-inspector](https://github.com/node-inspector/node-inspector/blob/master/package.json) -> [v8-debug](https://github.com/node-inspector/v8-debug/blob/master/package.json) -> [nan: "^2.0.4"](https://github.com/nodejs/nan).  See https://github.com/nodejs/node/issues/2798.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,6 @@
     "grunt-contrib-jshint": "~0.11.0",
     "grunt-contrib-uglify": "~0.9.1",
     "grunt-contrib-watch": "~0.6.1",
-    "grunt-debug-task": "~0.1.5",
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Great, I've had trouble installing this on some machines (windows, recent macbook) too.

gordonwoodhull:
Thanks @mtraynham!

PR #1022, opened 2015-10-12
Box Plot: Fix keyAccessor, extent filtering, x domain change

mtraynham:
This fixes the data binding between the `g.box` SVG nodes and datum keys.  It should be using the keyAccessor, but it just happens to work here because Crossfilter always has datums with a `key` field.  There are a few other places `dc` does this...

This also fixes filtering with X domains that are not ordinal and use brushing.  The issue can be seen in the examples page, [here](http://dc-js.github.io/dc.js/examples/box-plot-time.html).

### DIFF
--- a/src/box-plot.js
+++ b/src/box-plot.js
@@ -158,7 +158,7 @@ dc.boxPlot = function (parent, chartGroup) {
             .attr('transform', boxTransform)
             .call(_box)
             .on('click', function (d) {
-                _chart.filter(d.key);
+                _chart.filter(_chart.keyAccessor()(d));
                 _chart.redrawGroup();
             });
     }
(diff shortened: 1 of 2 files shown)


mtraynham:
Seems my own commit suffered from incorrectly accessing the datum's keys...

gordonwoodhull:
Yeah, *a few* places.

```
$ grep \\.key\\b src/* | wc
      33     204    2432
```

:frowning: 

Thanks, this is really helpful!

mtraynham:
33 that's not so bad. Maybe a few are in comments.

mtraynham:
One more change I've added to this.  If the x domain changes, the new width is only applied to the `enter` nodes.  Older nodes retain their old width.  Simple change.

gordonwoodhull:
Thanks @mtraynham! Merged for 2.0 beta 33.

PR #1023, opened 2015-10-14
CSS minified version

indrimuska:
Added grunt rule to create `dc.min.css` starting from `dc.css` using [grunt-contrib-cssmin](https://github.com/gruntjs/grunt-contrib-cssmin) plugin.


gordonwoodhull:
Doesn't seem like a huge win (3k vs 4k) but sure, seems like a good idea. 

(The annoying thing is it means doing another PR to cdnjs on the defunct 1.7 branch, but that's not your problem.)

gordonwoodhull:
Thanks @indrimuska! Merged in 2.0.0 beta 20

indrimuska:
:+1:

Issue #1008, opened 2015-09-17
Rowchart has Infinite height  when _rowData.length is 0


nordfjord:
See: https://github.com/dc-js/dc.js/blob/6e95388b9ae37d12eaf2009a4c8f28d48f71a698/src/row-chart.js#L169

when you filter such that a rowchart has no rows, its height becomes Infinity, which translates to a lot of errors along the lines of:

`Error: Invalid value for <g> attribute transform="translate(NaN,0)"`

gordonwoodhull:
Seems like a legitimate bug, reopening (?)

nordfjord:
It might be a legitimate bug.

But the errors I was receiving had nothing to do with this, I was way off base.

They were caused by a preRedraw function I had which calculated the xAxis domain and didn't check for infinity/NaN. I haven't seen any errors crop up around the height being Infinity, so I think it should be safe to close.

Regards,

Einar

alisabzevari:
I think this is related to a calculation in `updateElements` function (https://github.com/dc-js/dc.js/blob/develop/src/row-chart.js#L173). This error happened to me:
`Invalid negative value for <rect> attribute height="-1.3095238095238095"`
It happened when I had too many rows and the chart has a small height. So the calculation made the height of every row to become negative.

gordonwoodhull:
I looked and looked, but couldn't find any way to detect invalid DOM attributes being set in tests? Seems really weird that this can't be tested (via Jasmine). However, I did add a test and verified that it was creating errors before the fix (described in #1024) but not after.

gordonwoodhull:
@alisabzevari, that's actually a different error - the opposite error. This one has to do with having no rows, not lots of rows. 

If you'd care to file a separate issue, I'd appreciate it. Even better with a test case (fiddle or jasmine).

gordonwoodhull:
Fixed in 2.0 beta 32

PR #1024, opened 2015-10-14
make sure _labelOffsetY doesn't become Infinity

nordfjord:
make sure _labelOffsetY doesn't become Infinity and spew a ton of errors to the console.

height becomes Infinity when `n == 0`, meaning any time the chart is rendered with no data.

### DIFF
--- a/src/row-chart.js
+++ b/src/row-chart.js
@@ -172,7 +172,7 @@ dc.rowChart = function (parent, chartGroup) {
         }
 
         // vertically align label in center unless they override the value via property setter
-        if (!_hasLabelOffsetY) {
+        if (!_hasLabelOffsetY && height !== Infinity) {
             _labelOffsetY = height / 2;
         }
 


gordonwoodhull:
Thanks @nordfjord! Much appreciated.

gordonwoodhull:
Hmm, started to write a test case for this, but I can't figure out: if there are no rows, then how is `_labelOffsetY` used at all? I.e. maybe it's infinity but then what's it being applied to?

nordfjord:
Sorry for the late reply (it's hard to sift through all the emails I receive from github to find the relevant ones).

This bug is not really a bug in the sense that it doesn't break anything, the only thing that happens is it fills my console up with an annoying amount of errors, which I tend to try to avoid.

To be honest I don't actually remember what my test case was. I think I was using a rowchart whose group filtered out 0 values, and when I applied filters so that no values were left in the group my console started spewing tons of errors. I then added this check and like magic my console was clean again.

Fil:
My fix was this one:

```
--- a/js/dc.js
+++ b/js/dc.js
@@ -8487,7 +8487,7 @@ dc.rowChart = function (parent, chartGroup) {
     }

     function updateElements (rows) {
-        var n = _rowData.length;
+        var n = _rowData.length || 1;

         var height;
         if (!_fixedBarHeight) {
```
 

(Wish it had been merged. I've just spent 2 hours on this same issue :-/)

gordonwoodhull:
@Fil, I still don't see how Infinity can be applied to any bars if there aren't any bars. What am I missing? Do you have a test case?

gordonwoodhull:
Aha! I think I found it. `updateElements` is getting called at the end of redraw, but the variable is used in `updateLabels`, which is called **twice**, both in `createElements` and in `updateElements`. So the time **after** the chart is emptied out, it will update the labels wrong and then fix them. 

This also would explain why it spews errors but draws correctly.

So I think the correct fix is to remove the first, redundant call from `createElements`.

Sorry for being obstinate about not taking fixes without tests or explanations. I feel that it only makes the code harder to read.

Fil:
I can confirm that "to remove the first, redundant call from createElements" fixes it too for me.

Fil:
(And I totally agree with your being obstinate about getting to the real cause of any bugs rather than just routing around.)

gordonwoodhull:
Thanks @Fil! I'll put that in the next release.

nordfjord:
Shouldn't we close this PR since it's not the real cause of the error?

gordonwoodhull:
Thanks, I was looking for this issue to include the fix in the last release, but I didn't find it in time. 

I'm going to reopen - the description is rarely the root cause!

gordonwoodhull:
Ref: #1008

gordonwoodhull:
Fixed in 2.0 beta 32. Thanks @nordfjord and @Fil!

PR #1025, opened 2015-10-14
Fixes bubble chart filtering issues

mtraynham:
Bubble charts have a label, that is not always visible given a particular radius.  It's hidden by setting `opacity=0`.  Unfortunately this still allows you **click** the text, and filter the bubble.  This is problematic as a user can select a bubble that may be close to another bubble and filter the wrong one, because the label was long or the bubbles overlap...  Adding a `style='display: none/block'` update to the selection does work, but to prevent any interaction with the `opacity` working properly, it needs to happen **after** the transition.  Thus, `dc.afterTransition` was added.

The other issue with bubble charts, small bubbles may be hidden behind large bubbles.  Sorting the data descending by radius before hand, ensures that larger bubbles are first in the data and thus higher in the DOM, behind smaller bubbles.  A call to [d3.selection.order](https://github.com/mbostock/d3/wiki/Selections#order) is required for them to properly be sorted.

### DIFF
--- a/spec/bubble-chart-spec.js
+++ b/spec/bubble-chart-spec.js
@@ -232,10 +232,10 @@ describe('dc.bubbleChart', function () {
         it('creates correct label for each bubble', function () {
             chart.selectAll('g.node title').each(function (d, i) {
                 if (i === 0) {
-                    expect(d3.select(this).text()).toBe('F: {count:0,value:0}');
+                    expect(d3.select(this).text()).toBe('T: {count:2,value:77}');
                 }
                 if (i === 1) {
-                    expect(d3.select(this).text()).toBe('T: {count:2,value:77}');
+                    expect(d3.select(this).text()).toBe('F: {count:0,value:0}');
                 }
             });
         });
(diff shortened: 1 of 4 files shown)


mtraynham:
One thing I noticed... `keyAccessor` is a terrible function name for this chart.  It would be nice to call it `xAccessor`, and actually have a keyAccessor that corresponds to `d.key`.

gordonwoodhull:
I definitely agree with the small bubbles change, and `.dc.afterTransition` is a welcome addition.

But wouldn't [`pointer-events: none`](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events) be an easier way to make the labels non-clickable?

mtraynham:
Ahh, you know what, it probably would be.

gordonwoodhull:
Yep, that's one of those OO inheritance pitfalls. `keyAccessor` mostly makes sense for the other coordinate grid charts, but very little sense here. Especially horrible that, out of foolish consistency, there is no key accessor for this chart, as you point out.

I had to implement `dc.afterTransition` for dc.graph.js... Hmm.

mtraynham:
Updated with `pointer-events` instead.  

Yeah, not sure what to do about that `keyAccessor`.  I had to hack my own code's `data` function to loop the data and append a `key` field to every datum.  Not terrible, unless the data from the server already has a `key` field, then I'm screwed...

gordonwoodhull:
I think we could fix it with a breaking change of xAccessor, as you suggest. Technically I guess a bubbleChart is not a coordinateGridMixin but something else. Or maybe all coordinateGridMixins should have an xAccessor, just usually defaulting to the the keyAccessor.

gordonwoodhull:
I am making `bubbleSortSize` an option (defaulting to old non-sorted behavior) and merging for 2.0 beta 21

Thanks @mtraynham!

Issue #211, opened 2013-07-01
Feature Request: Labels above bars in bar chart


wolframarnold:
Hi Nick, Weidong,

I know that the label method isn't effective for bar charts, but I'd like to make a feature request to add labels above the bars in a bar chart. It would be really helpful to see, e.g. the total bar height as a label above the bar. Similar as the labels in the row chart or the pie chart, but above the bar.

I love your work. It's been very helpful.

Cheers,
Wolf

nreese:
What is the status of this issue? Did a pull request ever get merged into the baseline? If so, is there an example showing how to create a bar chart with labels?

gordonwoodhull:
Hi @nreese there's never been a PR on this. I'd go with @dimirc's second answer to the SO question above, and take @mr23's answer into account as well. If you want to file a PR that would be awesome.

nreese:
How should labels behave for stacked bar charts?

gordonwoodhull:
I think they would just show the total. It would be kinda messy to have them overlay the next bar up...

nreese:
I was thinking of placing the labels above the bars. What should happen if there is not enough space between the top of the bar and the top of the chart for the label? Should the label be placed below the bar top for these instances?

gordonwoodhull:
I would expect the domain of the chart to stretch a bit in that case. But I guess it's debatable.

nreese:
It looks like the domain can be expanded by setting yAxisPadding. Should dc.js try to automatically set yAxisPadding to an appropriate value when rendering labels or should this be left up to the developer to manually set yAxisPadding as they see fit?

gordonwoodhull:
Sure. It's not perfect because it will also expand the padding below the chart, but it will do.

I would just leave it to the user, since we do not even automatically adjust the margins for axes or axis labels. (It would be nice, but probably better to address the layout problem in a comprehensive way. Like the others, this one depends on text height, which is somewhat tricky.)

nitinsurana:
+1 for having labels for stacked bar charts too. It'll be nice to have configurable option to decide the placement position in the bar - top/bottom/center.

prashant22april:
Hi All,
I urgently require a error bar charts using dc.js. Any help in this area would be highly appreciated. Something similar to the below image I am looking out for:
![download](https://cloud.githubusercontent.com/assets/15922042/17547940/20b2a07a-5f07-11e6-8a8d-dce332b8fa79.jpg)

[...]

Issue #998, opened 2015-08-26
bubble chart padding


gordonwoodhull:
bubble padding is even more of an issue than scatter plot padding #440 because of the variable size of bubbles

![image](https://cloud.githubusercontent.com/assets/1366709/9500558/321264da-4bf4-11e5-88ed-e6d787e99475.png)

It can be managed manually with  `xAxisPadding` and `yAxisPadding` but that's a lot of futzing.

gordonwoodhull:
And padding doesn't work very well because it adds the same amount to same to both sides.

Here, adding an absurd x padding of 50 and y padding of 400:

![image](https://cloud.githubusercontent.com/assets/1366709/9500668/b2abf228-4bf4-11e5-9321-2da535022c01.png)

gordonwoodhull:
Should be calculated for each side independently. It can be put under a flag to make it backward compatible.

tttp:
Hi,

I've been experimenting with AxixPadding, and it isn't ideal, mostly because I end up having ticks that are under 0, when you display a number of observation, a negative number is confusing.

A better solution I've found is to remove the clip-path, so the bubbles are drawn under the axis, and works fine IMO.
![bubbly](https://cloud.githubusercontent.com/assets/128426/14987633/9f78b304-1150-11e6-8000-70023685132e.png)

tttp:
Is there any reason for the clip-path? if it's a workaround only for me, I can set a post render and remove the clip-path on chart-body

gordonwoodhull:
That's a decent workaround for your case. We should consider making the clip path optional. I think it mainly exists to allow zooming of line and bar charts (#108) but zooming of bubble charts is also possible.

I don't think eliminating the clip path is enough in all cases. If bubbles are really big they will still leave the chart area.

tttp:
It takes 3 lines with renderlet and no (visible) impact to add then remove clip-path. Not sure it's worthwhile adding an option to make the clip path optional.

I've documented the workaround on SO, I'd say it's good enough ;)

http://stackoverflow.com/questions/37042901/how-to-avoid-truncated-bubbles-when-elasticx-or-elasticy/37042902#37042902

PR #1031, opened 2015-10-27
Bar chart labels

nreese:
pull request for issue https://github.com/dc-js/dc.js/issues/211 adding support for bar chart labels


src/bar-chart.js
```diff
@@ -50,6 +51,13 @@ dc.barChart = function (parent, chartGroup) {
         return _chart._render();
     });
 
+    _chart.label(function(d) {
+        return d.y0 + d.y;
+    });
+    //_chart.label() sets _renderLabel to true.
+    //Set back to false to disable renderLabel by default
+    _chart.renderLabel(false);
+
```
gordonwoodhull: this seems like an odd default value, the Y coordinate? why not just leave `.label()` uninitialized?

gordonwoodhull:
I must be missing something about the label = Y coordinate default. 

Other than that this looks really good. Thanks for including tests and for making it backward-compatible; since it is, we can include it in 2.0.

nreese:
I figured that most users want bar labels to display the y-axis value. For my use case in particular, I had some bars that were order of magnitudes taller than others (150 vs 3). Since the small bars are so tiny, 2 or 3 pixels, I wanted a way to convey that they were not zero.
What would you recommend for a default value function?

gordonwoodhull:
Oh, I get it now. Yes, of course that's a reasonable default. 

I guess we'll probably have to default in some floating point formatter. Your comment about tiny bars reminds me that it would be nice to make the labels clickable.

You can take on these improvements if you want; otherwise I'll just address them when I merge.

I also see that I'll need to do something about the padding; the stacked bar example makes it painfully obvious how poorly `yAxisPadding` works here, because adding a gap below zero is quite wrong. It's pretty much the same change as is needed for #998 bubble chart padding. I can address this after merging this.

nreese:
I can take a look at making the labels clickable - that should be straight forward.

How many decimal places should the default label generator format floating points to? Maybe 2?

What do you have in mind for y axis padding?

gordonwoodhull:
Thanks @nreese! We can use the same default floating point formatting that we use for the filters. It's `.2f` but it's configurable. In fact, maybe it makes sense to use `dc.utils.printSingleValue` for consistency.

https://github.com/dc-js/dc.js/blob/develop/src/utils.js

gordonwoodhull:
It might be enough just to have padding on each side instead of duplicated on X and Y.

But I don't want to commit to a design yet, because there is too much in dc.js which was designed hastily and continues to haunt.

nreese:
I have made the labels clickable and used dc.utils.printSingleValue for format the label value. 

Let me know if anything else is needed.

nreese:
It looks like there is a merge conflict. How do I look at/fix the merge conflict? Are you waiting on action from me for this pull request?

gordonwoodhull:
Thanks @nreese, don't worry about the merge conflict, although in the future it is a best practice not to check in the build artifacts when doing a PR. (It usually doesn't matter but it makes life difficult for the maintainer in rare cases, so that's [why we put that in the contribution guidelines](https://github.com/dc-js/dc.js/blob/master/CONTRIBUTING.md).)

Sorry for the slow turnaround. I haven't had time to thoroughly review this. But it looks ready afaict.

gordonwoodhull:
This merged to 2.0 beta 21. Thanks!

@nreese, please let me know how to credit you in the AUTHORS file - your commits are marked with Reese and a bogus email address. 

(Probably a good idea to set [your name](https://help.github.com/articles/setting-your-username-in-git/) and [your email](https://help.github.com/articles/setting-your-email-in-git/) in git if you're contributing to projects on GitHub.)

PR #1032, opened 2015-10-29
Heatmaps seem to be empty after crossfilter reloads data

smee:
`selectAll` does not inherit data from its parent. This means that if the underlying crossfilter gets new data the selection of `g.box-group` will see the new data, the children `title` and `rect` will not. The optical result of that is a seemingly empty heatmap although the data has changed.

### DIFF
--- a/src/heatmap.js
+++ b/src/heatmap.js
@@ -228,10 +228,10 @@ dc.heatMap = function (parent, chartGroup) {
 
         if (_chart.renderTitle()) {
             gEnter.append('title');
-            boxes.selectAll('title').text(_chart.title());
+            boxes.select('title').text(_chart.title());
         }
 
-        dc.transition(boxes.selectAll('rect'), _chart.transitionDuration())
+        dc.transition(boxes.select('rect'), _chart.transitionDuration())
             .attr('x', function (d, i) { return cols(_chart.keyAccessor()(d, i)); })
             .attr('y', function (d, i) { return rows(_chart.valueAccessor()(d, i)); })
             .attr('rx', _xBorderRadius)


gordonwoodhull:
Thanks @smee! Good catch.

mtraynham:
@gordonwoodhull There seems to be a few other places where `selectAll` is being used where `select` is likely more appropriate.  

Bubble Chart:
https://github.com/dc-js/dc.js/blob/develop/src/bubble-chart.js#L117
https://github.com/dc-js/dc.js/blob/develop/src/bubble-chart.js#L133

Bubble Mixin:
https://github.com/dc-js/dc.js/blob/develop/src/bubble-mixin.js#L133

Geo:
https://github.com/dc-js/dc.js/blob/develop/src/geo-choropleth-chart.js#L158

Heatmap:
https://github.com/dc-js/dc.js/blob/develop/src/heatmap.js#L234
https://github.com/dc-js/dc.js/blob/develop/src/heatmap.js#L237
https://github.com/dc-js/dc.js/blob/develop/src/heatmap.js#L252
https://github.com/dc-js/dc.js/blob/develop/src/heatmap.js#L269

Line:
https://github.com/dc-js/dc.js/blob/develop/src/line-chart.js#L356

Row:
https://github.com/dc-js/dc.js/blob/develop/src/row-chart.js#L208

gordonwoodhull:
Thanks @mtraynham - this is confusing because in most of the code you refer to, the data is used right on the next line. And it doesn't seem to be nested data. 

I think we would know if all these locations were failing. So are there some cases where selectAll will in fact replicate the data to the child?

mtraynham:
Well I deemed these cases as bad, because it was using a `selectAll` parent selection which did recieve `.data()` and then using a `selectAll` for a single child item, such as labels or a `g` which contains a single `circle`.  It's not entirely problematic because, yes, the data is not nested, but it does introduce issues with enter/update/exit selections.

I believe some of these may be masked by the fact that we transition the same nodes twice:
https://github.com/dc-js/dc.js/blob/develop/src/bubble-chart.js#L117
https://github.com/dc-js/dc.js/blob/develop/src/bubble-chart.js#L133

gordonwoodhull:
Nice!

Hmm. Just when I thought I understood selections. I agree they should be `select`s but wonder why they are not failing as they are. (Or failing twice!)

gordonwoodhull:
See https://github.com/dc-js/dc.js/pull/1237#issuecomment-266952653 for the reason this doesn't break the whole world. (Hint: most people use crossfilter, but dc.js doesn't require you to.)

gordonwoodhull:
Yesterday Mike posted a detailed rationale for why `select` works this way: https://github.com/d3/d3-selection/issues/106#issuecomment-267223054

gordonwoodhull:
adding test and merging for 2.0. thanks @smee!

gordonwoodhull:
still need to check against the very helpful @mtraynham list above

PR #1033, opened 2015-10-31
Y Axes alignment option: addendum

gazal-k:
Follow up to #985


gordonwoodhull:
Just as a progress report, I'm merging this in the hotfix/2.0.0-beta.32 branch. Here is a matrix of the 25 cases which I'm adding as an example and visual test case. 

(There are at least 25 cases because left and right can each be all positive, mostly positive, even positive/negative, mostly negative, or all negative.)

image

Looks like the current heuristic gets many cases well, but severely squishes one series or the other in about 8 cases, and loses one or the other in two cases. 

I'm going to try a more comprehensive approach.

gordonwoodhull:
Improved algorithm:
image
image
http://dc-js.github.io/dc.js/examples/align-axes.html

gordonwoodhull:
Live in 2.0 beta 32. Thanks @gazal-k!

PR #1041, opened 2015-11-17
Add dc.util docs

### LINKED ISSUE #969: dc.pluck() not in API docs
I would have expected it to be listed here:
https://github.com/dc-js/dc.js/blob/master/web/docs/api-latest.md#utilities

mtraynham:
Surprising these weren't there before... Fixes #969


gordonwoodhull:
Thanks!!

Issue #1046, opened 2015-11-25
Vertical grid lines on a bar chart are black (not light gray)


mikemsq:
Vertical grid lines on a bar chart are black (not light gray) - see the image.
Probable cause - style override ???
![dc 2 0 0-beta 21](https://cloud.githubusercontent.com/assets/3779501/11401592/412dc3f6-9361-11e5-8641-3fd9f8be2494.PNG)

gordonwoodhull:
Good eye. I think we can just replace the selector instead of having both, since the new one is strictly more specific. Right?

mikemsq:
That would do.
I wasn't sure if the other one is used anywhere else.
I'll update the fix.

Thanks,
Mike

gordonwoodhull:
You were right the first time. Thanks, merging in beta 24!

PR #1047, opened 2015-11-25
Fix for #1046

mikemsq:
Added more specific css selectors


gordonwoodhull:
Link #1046

gordonwoodhull:
Good eye. I think we can just replace the selector instead of having both, since the new one is strictly more specific. Right?

gordonwoodhull:
Merged in 2.0 beta 24

PR #1053, opened 2015-12-03
adding a pyramid chart.

Sarah-W:
Useful for population demographics and where ever your data comes in pairs you want to compare.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -403,6 +403,7 @@ module.exports.jsFiles = [
     'src/geo-choropleth-chart.js',
     'src/bubble-overlay.js',
     'src/row-chart.js',
+    'src/pyramid-chart.js',
     'src/legend.js',
     'src/scatter-plot.js',
     'src/number-display.js',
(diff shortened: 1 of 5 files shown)


gordonwoodhull:
Thanks @Sarah-W! Looks like the canonical name for this kind of chart is "diverging stacked bar chart", or in this case "diverging stacked row chart". 

Of course, we don't currently support stacking in row charts, so maybe we should just call it a diverging row chart. A pyramid chart seems to be something else, used more for conceptual charts (like the food pyramid).

What I'm going to do here is merge the best from this PR and @ruhley's #943 "paired row chart".

Sarah-W:
Sweet! I have made a couple of tweaks since the first PR, will push them now (else it will be the new year before it happens)

gordonwoodhull:
Here's a branch with both versions included, for comparison and eventual combination into the best-of-both:
https://github.com/dc-js/dc.js/tree/pyramid-paired

I've run out of time for today!

Sarah-W:
Well I've run out of time for this year. :) I will be not so much AFK as AF-IDE till early January. I look forward to seeing what you do with it.

gordonwoodhull:
Screenshot
image

Issue #397, opened 2013-11-05
Add stacked row charts


jefffriesen:
Nick said that doing this would probably be more than a renderlet could do. 

I would love this feature. Especially if you could choose between a regular row chart like this:
![screen shot 2013-11-05 at 2 19 41 pm](https://f.cloud.github.com/assets/445733/1477689/035503c8-4660-11e3-9b92-3cc74449bbad.png)

Or normalized (100%) row chart like this:
http://dimplejs.org/examples_viewer.html?id=bars_horizontal_stacked_100pct
![screen shot 2013-11-05 at 2 22 44 pm](https://f.cloud.github.com/assets/445733/1477717/7a37e30c-4660-11e3-9949-71372fe1bf98.png)

For reference, here is a normalized vertical bar chart from Mike Bostock:
http://bl.ocks.org/mbostock/3886394
![screen shot 2013-11-05 at 2 24 39 pm](https://f.cloud.github.com/assets/445733/1477749/b5fe5948-4660-11e3-93b2-5ad6bb184fa8.png)

carlthuringer:
@RohanHart :+1: Just integrated [your fork](https://github.com/RohanHart/dc.js) in a project that called for stacked row chart. Works flawlessly, compatible with the other charts we were building.

I would really like to see this integrated. Thanks for the hard work.

RohanHart:
Glad it worked for for you.  Hopefully being a year behind the master wasn't too painful.

flxfxp:
Any news on this? I'd love to have stacked row bars :)

jennakwon06:
Do you have an example or demo for the stacked row chart?

rusosnith:
bump! what happened to this?

gordonwoodhull:
Hi @rusosnith. It would be great to have this feature, but no one has contributed it. If you want to help, you could take a look at @RohanHart's fork and see how hard it would be to bring this feature into the dc.js library. Or I think there have been other attempts, but none resulted in a PR.

renauld94:
Hi @gordonwoodhull  is it still in developement?

gordonwoodhull:
Hi @renauld94. When I looked at @RohanHart's fork way back when, I wasn't able to piece apart the changes for this feature versus other changes he made. It has been a while but if anyone wants to figure this out, the contribution will be welcome!

rusosnith:
Damn. I got here again. To ask what happened to this. And then realize I had already asked the same thing 2 years ago

[...]

PR #1068, opened 2015-12-17
Paired row chart, again

gordonwoodhull:
This replaces #943, pulling just the relevant commits and updating for code standards. 

Also includes `rowChart.xAxisLabel` since the example depends on that. 

- [ ] add tests for `rowChart.xAxisLabel`


gordonwoodhull:
@ruhley, I don't think I got quite all of the formatting and code from your branch, as I just took the following commits so far:

  * Row Chart ```useRightYAxis``` (ba8bbc0e1d5d203ec39a39f9f99a0ee83d4597a2 / 217e294651961981a6959fa7ce30f21afe100fc7)
  * Paired Row Charts (137213b566ae39e5f646016c9f13e30b0f6385a1 / 1267f7a82d4b8280d102460e1e85126aaffd0a60 / 5494b449f95a183bbc8416aa01b7bbc37efcf460 / 8c5ed0de8ac914c934da497fbfc54d7c85caa22f / 7e71309853c6a664f9a6827d8b291c5c56bd1db1 / 58c93c1b3773fd4a7b455bbc09f2725eeec55aef)
  * Row chart ```xAxisLabel``` (c6e473bf359f069f5d4e6883d16f25827084a7a0 / a8d72ae8b8a2d0d4a4dce11b377f044365cd58fc / 289ae4600a53aa7a3396e9d95257264535fc5cfc / 27c03a48d9328673c632064aa7c784bfd6a1411a)
  * Getting grunt server to watch and compile the main css file (60353aa2eca1ec32174a7dedf6aec7c2a2cadef6)

ruhley:
@gordonwoodhull That looks great. Can't wait for some of this stuff to make it into the core library. :+1:

gordonwoodhull:
So the formatting is a little bit messed up:

image

I also couldn't figure it out from looking at the demo on your fork, because it has different formatting problems:

image
http://ruhley.github.io/dc.js/examples/paired-row.html

(note that [the row chart is completely faded out](http://ruhley.github.io/dc.js/examples/row.html), too) 

So it's kind of hard to do a fair comparison between #1053 and this one. I like the design and reuse of code better on this version, but maybe it's easier to get it right if it's just another chart with all the code duplicated. Anyway, I'm creating a special branch just to compare and merge the two, and will continue working on it.

ruhley:
Yea, I always felt it was a little hacky just making two sub row charts. Duplicating the code or creating a row chart mixin would definitely be a more robust solution.

gordonwoodhull:
Huh, some kind of mixin, so that the code is shared. That's an interesting idea.

gordonwoodhull:
The design of #1053 is kinda hacky in different ways. :grinning:

gordonwoodhull:
Here's a branch with both versions included, for comparison and eventual combination into the best-of-both:
https://github.com/dc-js/dc.js/tree/pyramid-paired

I've run out of time for today!

ruhley:
I will look and try to get something done if I have some time.

mtraynham:
Does this only support pairing?  Would there ever need to be a case that we support "stacking" or rather more than 2 domain values on for each row?

gordonwoodhull:
Sure, people want stacked row charts all the time. @RowanHart reported some progress on his fork in #397 but it never made it into a PR, due to the difficulty of writing tests IIUC. :(

I don't think that really helps us here since d3.stack doesn't do bidirectional as far as I can tell. (Maybe this is an oversight since it does support streamgraphs?) So there is still the matter of somehow tracking two row charts and combining them, and I somewhat prefer the approach here of combining the charts, over the approach in #1033 of creating a whole new chart, although that one is a little more stable and there are good arguments for each. As @ruhley says, some kind of row mixin might be the best of both worlds.

I know it's not helpful to bring it up, but just to stir the pot :smiling_imp:, bar charts might also want bidirectionality, i.e. negative stacks. And both row charts and bar charts already support negative bars. (Evil laugh.)

gordonwoodhull:
Here's the "mixed positive and negative stack" issue on d3. Guess it's an open problem: https://github.com/mbostock/d3/issues/2265

mtraynham:
I have a stacked version that works pretty well :smile: ...  It's a bit different from the dc.js implementation though (uses my own stack mixin which allows missing data points, and supports scrolling).  I can post the code if you're interested.

gordonwoodhull:
dc.js is diverging... I can't keep up with all the pull requests and a lot of people (like @ruhley) have live forks which would be pretty difficult to merge.

I think this is normal and healthy, actually, but it can be rather confusing, especially for me. I will merge a lot of these but it's also great to have so many "wild" forks.

So yes, @mtraynham, please at least push to your fork, or if you can handle the overhead, make a feature branch and submit a PR. 

Making sense of this area of stacks and groups and multiples is of particular interest to me and it's helpful to have many implementations to compare and combine.

mtraynham:
Well in the best interest of everyone else using the library, we should set some shared goals on the design of these charts.  I've seen quite a few variations between the PR's already and even my own version seems somewhat silo-ed in purpose.

At the same time, I would prefer we keep complexity of supporting 3rd and 4th dimensional charts to a minimum.  The less maintenance required going forward would really help make additional features easy to implement.

My post from a while back on the Google Group I thought was a good starting point, but it never really took off in terms of feedback from others. https://groups.google.com/forum/#!topic/dc-js-user-group/Rj_u-D5ygEk

gordonwoodhull:
Wow, really great post and I have arrived at the same conclusion after reviewing all these submissions. I see that I didn't read it closely enough because I was concerned about holding back contributors while working out a perfect design - but what's happened is that a lack of a good design has held *me* back from merging these PRs!  

I'll resurrect that thread soon, once I've read your code and thought about it some more. My gut feeling is that both data formats should be supported, depending on the desired filtering behavior. I.e. if one wants to allow filtering on sub-dimensions then one must use multidimensional keys, but if not then multidimensional values should be preferred because they are more efficient (less bins for crossfilter to handle).

Either way the data will be transformed by a mixin so that the individual chart doesn't need to be concerned with the original format. Bravo.

ruhley:
I have started to implement a row mixin for the row chart and paired row chart. I think it is the best solution. I will see if the stack mixin can easily be added on top of the row mixin.

@mtraynham The layer mixin looks like it could be a really great addition for dc.js

ruhley:
@gordonwoodhull I have been very busy lately and my version of dc.js fell behind a bit. Today I managed to rebase my develop branch up to yours. I then reimplemented everything that I had before (except for the responsive stuff as I see others have different solutions and the proper choice needs to be made). I have an initial version of the row chart and paired row chart based on a row mixin.

As you can see the https://github.com/ruhley/dc.js/blob/develop/src/row-chart.js and https://github.com/ruhley/dc.js/blob/develop/src/paired-row-chart.js files are fairly simple with most of the logic in https://github.com/ruhley/dc.js/blob/develop/src/row-mixin.js

I feel like it is a much better solution with no workarounds or hacks. There may still be some bugs and options that I haven't considered yet, but the standard display and crossfiltering works great.

jennakwon06:
Do you have a demo for your implementation @ruhley ?

gordonwoodhull:
Hi @jennakwon06, I think I linked it above: http://ruhley.github.io/dc.js/examples/paired-row.html

gordonwoodhull:
Screenshot of the version in this PR and the `pyramid-paired` branch.
image

PR #1080, opened 2016-01-05
Append choropleth titles by default

ferblape:
This PR tries to make the creation of the titles in choropleth graphs more resilient.

I have just copied how titles are built in other graphs, such as rowCharts.

### DIFF
--- a/src/geo-choropleth-chart.js
+++ b/src/geo-choropleth-chart.js
@@ -155,7 +155,8 @@ dc.geoChoroplethChart = function (parent, chartGroup) {
 
     function renderTitle (regionG, layerIndex, data) {
         if (_chart.renderTitle()) {
-            regionG.selectAll('title').text(function (d) {
+            regionG.selectAll('title').remove();
+            regionG.append('title').text(function (d) {
                 var key = getKey(layerIndex, d);
                 var value = data[key];
                 return _chart.title()({key: key, value: value});


gordonwoodhull:
Thanks @ferblape. By "resilient", you mean the titles are currently disappearing, or not getting updated?

ferblape:
I meant not getting updated.

mtraynham:
I've encountered this before, the code should be `.select` instead of a `.selectAll`.

> Nesting selections has another subtle yet critical side-effect: it sets the parent node for each group. The parent node is a hidden property on selections that determines where to append entering elements. … There is an important difference between select and selectAll: select preserves the existing grouping, whereas selectAll creates a new grouping. Calling select thus preserves the data, index and even the parent node of the original selection!

https://bost.ocks.org/mike/nest/

Using `selectAll` here is unbinding the parent datum from updates.

gordonwoodhull:
I looked into this as part of #1239 and added tests for the case where new data objects are supplied for each redraw in aa2785880d. 

I don't get it. It's not using data binding for the title! Here is the current source:

```js
            regionG.selectAll('title').text(function (d) {
                var key = getKey(layerIndex, d);
                var value = data[key];
                return _chart.title()({key: key, value: value});
            });
```
https://github.com/dc-js/dc.js/blob/31682740c4b5cd8f26087db5515150cf64f948c9/src/geo-choropleth-chart.js#L158-L162

`generateLayeredData` puts the current data into an object `data` that only exists for the lifetime of the render or redraw. It uses that data, *not the bound data*, to fill the regions and update the titles. It binds the map region data to the elements during render, and the crossfilter(ish) data is not bound to elements. The use of `selectAll` here is apparently intentional to avoid rebinding.

So if anyone can please tell me how the above code does not replace the titles with the current data, I am quite frustrated by this and would desire an explanation.

The only thing I can see might not be replaced is the key from the geoJson data, certainly not the value from the crossfilter data. But this chart does not support replacing the geoJson data with a redraw, you must render again. So I would think there would be a lot more broken than just the titles in that case.

@ferblape, is there any chance you can provide an example (e.g. a block or a fiddle) where titles fail to update? I really need a test for this, to make sure we are doing the right thing.

ferblape:
Of course @gordonwoodhull, I'll check the old code and create a block.

gordonwoodhull:
Thanks for your response, @ferblape! That's a big relief because I was afraid there was still another level of complication in these data joins that I didn't know about.

The way the choropleth binds data is odd, and IMO somewhat backwards, but in this case it should avoid the problem that @mtraynham has pointed out in many other charts.

Other folks have run into trouble adding d3-tip to choropleths, and there we determined that the problem was this strange data binding, which makes the map data available but not the reduced crossfilter(ish) data.

PR #1089, opened 2016-01-16
Update License year to 2016

pra85:


gordonwoodhull:
It's definitely valid. I've been considering generating these automatically, actually, which I do on another project.

gordonwoodhull:
Thanks @pra85!

pra85:
Apologies @gordonwoodhull , I closed the PR as I thought it had been abandoned. 

In case you don't want to update license year on a yearly basis, you can switch to an evergreen style `2012-present` (like one used by projects like React - https://github.com/facebook/react/blob/master/LICENSE )

gordonwoodhull:
No, dc.js is not abandoned, although I can never find as much time for it as I'd like.

Huh, I didn't know you could do that, wonder why the evergreen style is not more common. Like it.

gordonwoodhull:
Merging for beta 26. Thanks @pra85!

PR #1090, opened 2016-01-19
Give an example of filtering a chart by range

koefoed:
This may help users who try the regular crossfilter syntax of ```dimension.filter([15,20])``` and see nothing changing in the charts.


gordonwoodhull:
Good idea - that's a good point that the charts' `filter` function looks the same as the crossfilter one but really is quite different!

gordonwoodhull:
Thanks @koefoed! If you'd like [author credit](https://github.com/dc-js/dc.js/blob/develop/AUTHORS) besides your handle, please email me your name and email.

PR #1092, opened 2016-01-25
Add treemap - Could you please share sample jsonrequest

GunjanSh:
Hi,

I have been trying to use Tree map to show topic modeling.
But Treemap that i found in D3 js does not allow for Crossfilter.
I came across this code, but i am struggling to figure out the Json sample that would make this Treemap work. It would be great if you can share the same json request/format that this treemap accepts. Please respond back as i am stuck with this and unable to proceed.

Thanks!

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -409,5 +409,6 @@ module.exports.jsFiles = [
     'src/heatmap.js',
     'src/d3.box.js',
     'src/box-plot.js',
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
This appears to be a question on #1034 - closing.

GunjanSh:
Is it possible for you to share a sample json that can be used to create this nested treemap.
Or an example please.

Issue #1099, opened 2016-02-12
html docs: source links use local path


gordonwoodhull:
The generated HTML documentation uses paths on the build system where it was generated. As noted by @mtraynham, this is probably krampstudio/grunt-jsdoc#135:

> I did find that if you pre-concatenated everything it worked. But that made the source lookup extremely slow because it was all attached to one file instead of multiple...

but I'm seeing this even with a concatenated dc.js

Follow up to #1086

gordonwoodhull:
fixed by #1101

Issue #1097, opened 2016-02-12
html docs: classes in Namespaces menu


gordonwoodhull:
Our pseudoclasses are in a Namespaces menu, while the Classes menu just has the `dc.filters`

Follow up to #1086

gordonwoodhull:
fixed by #1101

PR #1101, opened 2016-02-13
Ref #1096, #1097, #1099, #1100: Fix some outstanding issues with HTML JSDoc

### LINKED ISSUE #1096: html docs: no pseudoclass parameters
The parameters to our "constructors" are not shown in the jsdoc html documentation.

Follow up to #1086

### LINKED ISSUE #1100: html docs: don't concatenate source
It's great that these docs refer to the source and format it, but they are referring to the huge concatenated dc.js not the individual source, which is just too big and slow.

Follow up to #1086

mtraynham:
Should resolve #1096, #1097, #1099, #1100.

For #1099, I tried splitting up the files and it seemingly fixed the paths, which also resolves #1100...

Minor caveat on #1097 (the classes).  Both jsdoc2md and grunt-jsdoc (with docstrap) add the "new" keyword to these classes.  Most of the documentation doesn't use it, but dc.js will work with/without the keyword...

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,7 +2,7 @@
 ## 2.0.0 beta 26
 * Apply pie chart labels before transition, so they are easier to manipulate with the pretransition hook. Added example of showing percentages in pie chart labels. (Workaround for [#703](https://github.com/dc-js/dc.js/issues/703))
 * Documentation of chart registry, by Jasmine Hegman ([#676](https://github.com/dc-js/dc.js/issues/676) / [#1082](https://github.com/dc-js/dc.js/pull/1082))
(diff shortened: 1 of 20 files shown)


mtraynham:
Aww fudge, targeted for develop.  Might want to pull back into master.

gordonwoodhull:
NP, I am pretty used to cherry-picking commits rather than doing straight merges.

Working on the docs raises a weird workflow issue: develop has more (pseudo)classes than master does. Currently just the one, `selectMenu`, needed its own separate update on develop. I'm not sure what the best approach is here. I guess it only affects these PRs that change just about every file, so it's not worth coming up with a procedure. But it's interesting that these kinds of changes can't be made in a single PR.

gordonwoodhull:
I merged this to master and develop. Since it doesn't affect the released code/markdown, there's no need for a tag/release, but the changes will be part of 2.0 beta 27 when that comes out.

PR #1114, opened 2016-03-08
Add ability to set max legend item count

renoth:
I needed this for a project so i implemented it and wanted to share with you guys.

### DIFF
--- a/src/legend.js
+++ b/src/legend.js
@@ -24,7 +24,8 @@ dc.legend = function () {
         _legendWidth = 560,
         _itemWidth = 70,
         _autoItemWidth = false,
-        _legendText = dc.pluck('name');
+        _legendText = dc.pluck('name'),
+        _maxItems;
 
     var _g;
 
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Thanks @renoth and sorry for the slow response. This is a good idea and I only wonder if we need to add an ellipsis or something to indicate that items were removed.

I take it you use this for a chart where the items are ordered from greatest to least, and you don't want to use [.cap()](https://github.com/dc-js/dc.js/blob/develop/web/docs/api-latest.md#dc.capMixin+cap), or it's not available for the chart you're using?

gordonwoodhull:
Thanks @renoth! This is going into dc.js 2.0.

I tried adding an ellipsis item and it looked dumb, so I'm not doing that. (Also I figured it would probably have to be a link to show the rest, and that's a bigger effort.) 

Added a couple of tests, checked for invalid numbers as you suggested, and also allowed zero (because why not).

PR #1116, opened 2016-03-11
transition delay update

mbustosorg:
Adding the ability to specify the transition delay in addition to transition duration.  This allows you to create some more interesting transition flows by adjusting when the transitions would start.  

```
  .transitionDelay(function (d,i) {
  return Math.floor((Math.random() * 1000) + 1);
   })
```

This creates a random 0-1 second delay which can give the transition another dimension.

### DIFF
--- a/spec/core-spec.js
+++ b/spec/core-spec.js
@@ -108,21 +108,26 @@ describe('dc.core', function () {
                 duration: function () {
                     return this;
                 }
+                delay: function () {
+                    return this;
+                }
             };
             spyOn(selections, 'transition').and.callThrough();
             spyOn(selections, 'duration').and.callThrough();
+            spyOn(selections, 'delay').and.callThrough();
         });
 
         describe('normal', function () {
             it('transition should be activated with duration', function () {
-                dc.transition(selections, 100);
+                dc.transition(selections, 100, 100);
                 expect(selections.transition).toHaveBeenCalled();
                 expect(selections.duration).toHaveBeenCalled();
+                expect(selections.delay).toHaveBeenCalled();
             });
 
             it('transition callback should be triggered', function () {
                 var triggered = false;
-                dc.transition(selections, 100, function () {
+                dc.transition(selections, 100, 100, function () {
                     triggered = true;
                 });
                 expect(triggered).toBeTruthy();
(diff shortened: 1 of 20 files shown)


mbustosorg:
Just realized this is failing.  Fixing now.

gordonwoodhull:
@mbustosorg wrote :
> I'm having trouble with the tests.  I'm getting the specs to run fine in
the browser but am seeing:
> > Error caught from PhantomJS. More info can be found by opening the Spec
Runner in a browser.

src/core.js
```diff
@@ -265,14 +265,12 @@ dc.redrawAll = function (group) {
  */
 dc.disableTransitions = false;
 
-dc.transition = function (selections, duration, callback, name) {
-    if (duration <= 0 || duration === undefined || dc.disableTransitions) {
-        return selections;
+dc.transition = function (selections, duration, delay, callback, name) {
+    if (dc.disableTransitions || duration <= 0 || duration === undefined || delay <= 0 || delay === undefined) {
+	return selections;
```
gordonwoodhull: Doesn't this enforce a delay? I would think a delay of zero is valid and should still invoke the transition.

gordonwoodhull:
Hrrmmm, the build system can be finicky, especially on Travis... and we haven't even gotten to the Sauce Labs cross platform tests. :-S

Do you mean it's also failing on your local machine? I'm not at the computer today but I can give it a shot tomorrow.

src/core.js
mbustosorg: Good point.  I'll update it.

mbustosorg:
It's behaving the same way on my machine.  Using --force shows all the tests succeeding and running it in the browser shows the same (all success).  I'm unable to figure out what the Error from Phantom JS is.

gordonwoodhull:
In that case, it's likely that one of our development dependencies has changed. This is all built on shifting sands. I'll take a look tomorrow.

spec/core-spec.js
```diff
@@ -108,21 +108,26 @@ describe('dc.core', function () {
                 duration: function () {
                     return this;
                 }
+                delay: function () {
+                    return this;
+                }
```
gordonwoodhull: it's simply a syntax error in the source here

gordonwoodhull:
You can run `grunt lint` to do a syntax check.

It will also point out that you're using tabs when the style here is to use spaces. (There's a commit hook to check this stuff but I'm not sure if it always gets installed automatically.)

Finally, before I merge this I'd like to see if it's possible to make `dc.transition` backward-compatible - it's not a documented function but it may be in use by folks out in the world (in derivative libraries for example). I think it may be possible to do some clever type-checking to see if `delay` has been omitted. I'll look at this before merging.

Otherwise, this seems like a nice general-purpose addition and a pretty clear way to open up another of d3's parameters. Thanks!

mbustosorg:
Ok!  Thanks for the hint.  Tests passing now and I've fixed the formatting issues.

gordonwoodhull:
Thanks @mbustosorg! Merged for 2.0 beta 33.

PR #1123, opened 2016-04-06
specified first two elements of d.key to pass to isFiltered()

KatiRG:
Needed when length of `d.key` > 2 (for example, it can contain a colour index) because `dc.filters.RangedTwoDimensionalFilter` returns false if `value.length !== 2` (see line 954).

### DIFF
--- a/dc.js
+++ b/dc.js
@@ -9046,7 +9046,8 @@ dc.scatterPlot = function (parent, chartGroup) {
             .attr('transform', _locator);
 
         symbols.each(function (d, i) {
-            _filtered[i] = !_chart.filter() || _chart.filter().isFiltered(d.key);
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Thanks @KatiRG! Applied your change for 2.0 beta 28. We presently do not have any tests for extra values in the keys, and I'm not adding any now, but it would probably be smart in the future.

In general this won't work for charts that use the entire key, but the scatter plot explicitly looks only at `key[0]` and `key[1]`, so this looks safe to me.

gordonwoodhull:
Rebased and merged in 2.0 beta 28

KatiRG:
Great, glad to help!

PR #1133, opened 2016-05-01
Pfafman

pfafman:
Fixed crossfilter2 in footer and update to latest cross filter version.


gordonwoodhull:
Thanks @pfafman! FWIW I'm not upgrading dc.js to crossfilter2 1.4 until it stabilizes.

Fixed in 2.0 beta 29.

gordonwoodhull:
Will push to npm once the tests cycle.

Issue #822, opened 2014-12-31
weird enter & exit transitions for bar chart


gordonwoodhull:
![bar enter-exit transitions](https://cloud.githubusercontent.com/assets/1366709/5590371/9b58c448-9100-11e4-8d64-edc5d5e72467.gif)

Once again, the enter transition should start where the bar would be in the before-scale, and the exit transition should end where the bar would end in the after-scale.  

In addition, the bars are collapsing from the bottom up when they should collapse from the top down.  (Although that might not matter, and might be a bad idea, if they are leaving to the left or right - maybe more accurate to just let them leave the screen.)

Can be seen in multi-focus example:

http://dc-js.github.io/dc.js/examples/multi-focus.html

gordonwoodhull:
The exit transitions are improved in #1146 / 2.0 beta 32. The enter transitions still need to be fixed, but that means keeping around the previous scale, which is a bigger change.

Issue #1144, opened 2016-05-19
knots in barchart transition


Fil:
In this example I'm adding data coming from a papaparse stream and the transition makes these funky knots.

![knots](https://cloud.githubusercontent.com/assets/7001/15410425/ee6c6bec-1e19-11e6-8ec2-4625466cb9df.gif)

Luiz-N:
It looks like the data being added is out of order (in the past) then later being sorted.. that's what causes the line to go backwards.

gordonwoodhull:
I'm having trouble viewing your GIF (it only runs once, and the effect is subtle). Does the problem look like one of the problems you can see in this testing page?

http://dc-js.github.io/dc.js/transitions/bar-transitions.html

If so, (or if you can repro similar problems in the area chart [on the main example page](http://dc-js.github.io/dc.js/)) this is an instance of #507. And this is the one issue blocking the release of 2.0. (You could also try 1.7.5 to verify it works there, if you aren't using later features.)

Fil:
Indeed the moment it makes knots seems to be when there's new data classes inserted in the line, resulting in a general mix-up in the line.

In my example the new data are the years 1954-1957 or so (beginning of the set), and so the visual problem disappears if I reverse-sort the data in `lineDraw()`:
`return safeD(line(d.values.sort(function(a,b) { return d3.descending(a.x, b.x)})));`

Hardly a generic solution, but it's really difficult to interpolate lines that don't span the same number of vertices (they are interpolated as strings not as sets of dots).

As I understand it, this is similar to #507, but a bit different. In #507 the reason is addition of new elements in the SVG; here it's addition of new segments in the same `path`.

Hope this makes sense. (I'm sorry I wasn't make it work with dc 1.7.3 to confirm.)

gordonwoodhull:
I think I get it. Yes I think this is another manifestation of the same problem, or at least should be solved by the same solution. However I'll keep this open to verify once that solution is in place.

What I think will work is:
- join the data using the X coordinates as keys (currently it's joined by index, the default)
- always make sure the full X domain including the starting and ending domain, is part of what's dream

Incidentally, I glean from your description of the problem that this issue would probably also be present in 1.7, since it also joined by index.

Issue #1142, opened 2016-05-16
Cannot read property 'charAt' of undefined


JasonBrannon:
npm --version: 3.8.6
node --version: 6.1.0

When running **dc.js$ grunt server** receive the following error output:

dc.js$ grunt server
Running "concat:js" (concat) task
Source map dc.js.map created.
File dc.js created.

Running "uglify:jsmin" (uglify) task
>> 1 sourcemap created.
>> 1 file created.

Running "cssmin:main" (cssmin) task
>> 1 file created. 4.87 kB → 3.39 kB

Running "copy:dc-to-gh" (copy) task
Copied 9 files

Running "jsdoc:dist" (jsdoc) task
Documentation generated to /Users/jason/Desktop/WhitelistUX/dc.js-develop/web/docs/html
Running "jsdoc2md:dist" (jsdoc2md) task
>> writing web/docs/api-latest.md
**Fatal error: Cannot read property 'charAt' of undefined**

gordonwoodhull:
Thanks for the report - looks like one of our zillion dev dependencies has been updated and broke things. 

I was not able to reproduce until I deleted node_modules and reinstalled and now I see it. Hope to take a look later today.

JasonBrannon:
Thanks for update.  For my purposes to run the examples, I changed the following in Gruntfile.js

grunt.registerTask('server', ['docs', 'fileindex', 'jasmine:specs:build', 'connect:server', 'watch:jasmine-docs']);
**to**
grunt.registerTask('server', ['fileindex', 'jasmine:specs:build', 'connect:server', 'watch:jasmine-docs']);

gordonwoodhull:
Great, glad you found a workaround for now!

mtraynham:
Seems to be related to some latest changes to jsdoc2md and the `@memberof` clause.  It breaks when you have a nested property dot-notation in the `@memberof` clause.
```
    /**
     * @memberof dc.barChart
     */
```
Simply doing `* @memberof barChart` won't work though.  I was able to get this to work after downgrading grunt-jsdoc2md to 0.4.5.
Maybe @75lb can comment, I believe he's been around here before.

gordonwoodhull:
Thanks @mtraynham, I filed an issue on grunt-jsdoc-to-markdown and @75lb tracked it down to a bug in a second-level dependency of grunt-gh-pages. I didn't know that was possible.

[...]

Issue #507, opened 2014-01-24
line/area chart transitions are weird when there are both updates and adds


gordonwoodhull:
looks like adds/updates are out of sequence.

example: change the time range on the stock example.  just moving forward forward/backward in time looks mostly okay, but changing the duration draws some lines first, some lines later, and the fill seems to follow the lines.

ldalegandar:
I'm very new to dc.js, but this is the one of the first things I noticed - the transitions on line and stacked area charts don't look right, particularly when using a rangeChart to filter the date range (I've been using them primarily for timeseries charts).

(If you've already seen Mike Bostock's path transition page then you probably know all this, but I’m writing it for my peace of mind!)

I think what's happening is that the chart is going through the enter/exit process, then updating the 'd' attribute for the remaining points to match the new filtered date range. 
Using an example- if the size of the filtered range doesn't change (i.e. the rangeChart brush stays at the same zoom, but is shifted left or right), then from d3's perspective there are no new points, and you’re just changing the position of the existing points. This means that rather than shifting the points left or right to match the new range, the points just move up and down. 
If new data points are added (e.g. the zoom level of the rangeChart brush is changed), then the new points are added to the map before the transition is called. Because they’re new points, they don’t have an existing position to transition from, so they’re just placed at the post-transition position (I think). This causes the ugly transition where the new points are in the post-transition position, while the existing points shift from their pre-transition position to the post-positon, causing the lines and shaded area to go all over the place.

What the chart should do is enter the new data points outside the filtered range (e.g. hidden outside of the clip path/svg), then translate the chart to match the new range, then exit/remove the old data points outside of the range (which is now hidden off screen). This would give a clean transition from left to right (or in and out), and maintain object constancy.
Strangely enough it looks like the axis already uses this transition structure (though I could be wrong).

Hopefully that makes sense. I tried to mess around with the source code to get the transition working, but broke it spectacularly! Once I’m more familiar with the dc.js code I can give it another crack though.

gordonwoodhull:
Thanks @ldalegandar, I agree with this assessment and it is helpful that you described it so well.

The reason why it works for the axes is that those are part of d3 itself, so of course mbostock got it right. The transitions in dc.js were written by a variety of people, probably none as expert as he. 

It is probably a simple fix, once we figure out what to do... any help would be appreciated!

ldalegandar:
[...]

PR #1145, opened 2016-05-20
fix weird transition on barChart exit() #822

Fil:
the transition to `height(0)` was making the bar "fly away", now we keep its y and height, but transition it towards where it "should be". trimming a little bit of the width gives nicer results imho so that a small ap is preserved between bars. But it's something that can be removed if you prefer to keep it simple.

### DIFF
--- a/dc.js
+++ b/dc.js
@@ -5433,7 +5433,8 @@ dc.barChart = function (parent, chartGroup) {
             .select('title').text(dc.pluck('data', _chart.title(d.name)));
 
         dc.transition(bars.exit(), _chart.transitionDuration())
-            .attr('height', 0)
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Thanks @Fil! I take it this is not a fix for #1144 or #507?

Changes should be made to the files in src/ in order for them to be picked up by the tests (dc.js itself is just a build artifact). Due to the docs currently being broken, you may have to disable them as commented in #1142.

Fil:
Indeed it solves only the barChart issue. #1144 and #507 are quite another beast.

(Sorry for the poor PR, I'm quite new to this project. I'm closing this one, please contniue the discussion on #1146 instead)

PR #1128, opened 2016-04-13
Fix Pie Chart tweening breaking on object-valued groups #1085

XaserAcheron:
See https://github.com/dc-js/dc.js/issues/1085 -- Pie tweening could sometimes cause d3.interpolate to throw an error, particularly when using the 'cap' feature.

PR #1151, opened 2016-06-01
Don't try to interpolate random data in label paths 

stillesjo:
At the moment when there is no data available for the pie chart dc.js will provide the value 1 as default, along with the emptyTitle (which can be set). I've experienced a problem when I have objects assigned as "value" in dc.js, with a value accessor that digs more deeply inside that object to find a value. When I then change filter so that it returns an empty set of data it will try to interpolate with my object (e.g. { value: 1}) and the empty value (e.g. 1). Interpolating between an object and a number will return NaN which will cause the charts to become unusable until refresh of webpage.

This PR gives you the possibility to set the value that should be used when no data is available after filtering. Then you can set it to an object which can be correctly interpolated to and from.

Another problem while developing this was that the key and value accessors for charts inheriting from the capMixin (cappedKeyAccessor/cappedValueAccessor) bothers with the others field in the object. Before this PR we automatically set it to an array containing the emptyTitle-variable. In order to be able to have the value accessed with the value accessor I also added the possibility to set "emptyOthers" to be used in the others field.

Added TC for emptyValue with both a number and an object. I was unable to create a good test case for the others field since it basically never happens since we have no data, so all cap options that I tried wouldn't trigger it to show the others tags; it always resulted in emptyTitle.

Another solution to this could be to combine title, value and others in an object and make it possible to set this object via only one method (instead of three as it is in this PR)

### DIFF
--- a/src/pie-chart.js
+++ b/src/pie-chart.js
@@ -214,8 +214,9 @@ dc.pieChart = function (parent, chartGroup) {
         polyline.exit().remove();
         dc.transition(polyline, _chart.transitionDuration())
             .attrTween('points', function (d) {
-                this._current = this._current || d;
-                var interpolate = d3.interpolate(this._current, d);
+                var current = this._current || d;
+                current = {startAngle: current.startAngle, endAngle: current.endAngle};
+                var interpolate = d3.interpolate(current, d);
                 this._current = interpolate(0);
                 return function (t) {
                     var arc2 = d3.svg.arc()


stillesjo:
Similar to #1128. Care to review @XaserAcheron ?

stillesjo:
The problem is that the values we tween between isn't consistent with eachother. In my case the value we go from is an object and the "empty value" is a Number. This will result in NaN which will break the charts.

Do you have any suggestion for how it could be done otherwise to avoid NaNs?

gordonwoodhull:
The heart of #1128 is that we were tweening data which didn't need to be tweened. We don't actually need the data to be tweened, just the angles. 

I haven't looked into this in-depth but my feeling is that there is probably the same root problem here: tweening an object we don't eventually use.

gordonwoodhull:
My other thought is that perhaps we don't need to call `pie` in this case, since we're basically passing it constants. We could just produce the post-pie-like data instead.

XaserAcheron:
Is the NaN-tween issue still happening in beta 28 or later? #1128 ought to have this fixed unless I missed an edge case or it's an eerily-similar-but-different issue. As @gordonwoodhull mentioned, the fix made it so the pie tweening doesn't care about or look at the data object, so the nefarious attempted-tweening of `1` and `{}` should no longer occur.

The PR likely has value for the other reasons mentioned in the first post, though. I think I've got my own workarounds for similar situations in my own project, but I'd have to spelunk through the code to be sure since my memory is notoriously zany.

stillesjo:
Hi again,
Thanks for you comments. I've been trying to reproduce the fault and noticed that the fault doesn't occur on tweenPie but in fact happens if you have external labels enabled, in the `updateLabelPaths` function in pie-chart.js.

I'll try implementing a similar solution there as in #1128 (i.e. only interpollating on only the angles) and see if it works.

[You can check out the fault in this codepen](http://codepen.io/stillesjo/pen/GqgPRy) by filtering in the line chart to give no data and then removing the filter. The fault is displayed in the log.

stillesjo:
@gordonwoodhull @XaserAcheron 
I've updated the PR with a similar solution as #1128 for updateLabelPaths. I've also tested this towards our application and it solves the problem. Please let me know what you think :cake:

XaserAcheron:
Oh heh, that explains it -- I'm not using external labels in the project I identified the issue in, so I missed  it. Thanks for the patch-up!

stillesjo:
@gordonwoodhull 
Did I miss any comment/stuff on this PR or could this be merged into develop?

gordonwoodhull:
Yes, it looks like this can be merged into the 2.0 betas. I've just overlooked it. Thanks!

gordonwoodhull:
Merged for 2.0.0-beta.33.  Thanks @stillesjo!

Issue #879, opened 2015-03-11
elasticY does not work if all values are negative (ordinal x axis)


jtsherba:
I believe I have found a little bug. 

For an ordinal bar chart with only negative values, the bar for the highest value does not draw correctly. 

The issue is not present when there are one or more positive values included. 

See example here: https://jsfiddle.net/tjsd4yqs/6/

jtsherba:
Never mind, just needed to set the y axis domain to start from 0.

gordonwoodhull:
Thanks @jtsherba. I think there is a bug here, because `.elasticY(true)` does not work if all values are negative.

https://jsfiddle.net/tjsd4yqs/12/

Retitling.

arjain:
Are your aware of any workarounds to make elasticY work with negative values?

sebgrohn:
We recently ran into this issue in our product as well, both with bar charts and row charts. After some investigation, we think that the following changes would fix the issue:

**For bar charts and others using Stack mixin,** change [`dc.stackMixin.yAxisMax`](https://github.com/dc-js/dc.js/blob/develop/src/stack-mixin.js#L185) to behave like `yAxisMin`, capping each stack's value to at least 0 (`y0`):

```javascript
_chart.yAxisMax = function () {
    var max = d3.max(flattenStack(), function (p) {
        return (p.y + p.y0 > p.y0) ? (p.y + p.y0) : p.y0;
    });

    return dc.utils.add(max, _chart.yAxisPadding());
};
```

**For row chart,** add a condition in [`calculateAxisScale` in `dc.rowChart`](https://github.com/dc-js/dc.js/blob/develop/src/row-chart.js#L56) for the domain max, like the one for the minimum:

```javascript
function calculateAxisScale () {
    if (!_x || _elasticX) {
        var extent = d3.extent(_rowData, _chart.cappedValueAccessor);
        if (extent[0] > 0) {
            extent[0] = 0;
        }
        if (extent[1] < 0) {
            extent[1] = 0;
        }
        _x = d3.scale.linear().domain(extent)
            .range([0, _chart.effectiveWidth()]);
    }
    _xAxis.scale(_x);
}
```

Are there more chart type, apart from Row chart, that don't use Stack mixin and should have a corresponding fix?

I would be happy to prepare a proper PR in the beginning of next week, if you think this is the way to go. :)

Ping @stillesjo

sebgrohn:
[...]

PR #1156, opened 2016-06-07
Include 0 as max in elastic axis if all values are negative

sebgrohn:
For Row chart and charts using Stack mixin, use the corresponding method for determining the max value for the y axis as used for min value:

> `yAxisMin`, current behaviour: If all values are _positive,_ use 0 as y axis _min._
> `yAxisMax`, new behaviour: If all values are _negative,_ use 0 as y axis _max._

Should fix issue #879.

Two Jasmine tests are currently failing, both related to Composite chart and its `alignYAxes`. As far as I can tell, it seems to be related to the calculations in [`calculateYAxisRanges`](https://github.com/dc-js/dc.js/blob/develop/src/composite-chart.js#L109) that does something magic with the y axis ranges if there are two y axes and `alignYAxes` are set. I am at loss here what to do, all suggestions are welcome!

[jasmine-failing-tests.txt](https://github.com/dc-js/dc.js/files/302730/jasmine-failing-tests.txt)

### DIFF
--- a/spec/row-chart-spec.js
+++ b/spec/row-chart-spec.js
@@ -91,7 +91,7 @@ describe('dc.rowChart', function () {
         });
     });
 
-    function itShouldBehaveLikeARowChartWithGroup (groupHolder, N) {
+    function itShouldBehaveLikeARowChartWithGroup (groupHolder, N, xAxisTicks) {
         describe('for ' + groupHolder.groupType + ' data', function () {
             beforeEach(function () {
                 chart.group(groupHolder.group);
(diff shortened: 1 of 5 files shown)


gordonwoodhull:
@sebgrohn, this looks like an excellent approach. I like the symmetry of it.

Would you be willing to also write/improve a few tests for this? 

In particular, the row chart [already has tests for all-negative values](https://github.com/dc-js/dc.js/blob/develop/spec/row-chart-spec.js#L359),  but it looks like it doesn't test `elasticX`. 

The bar chart [apparently does test elasticY with negative values](https://github.com/dc-js/dc.js/blob/develop/spec/bar-chart-spec.js#L507), but it didn't detect any problems. [Ditto for the line chart.](https://github.com/dc-js/dc.js/blob/develop/spec/line-chart-spec.js#L510)

Don't worry about `alignYAxes` since that feature is still somewhat in flux and #1033 may help there.

sebgrohn:
Great! I will try to produce some tests for negative values with elasticY enabled. I can't give any time frame right now, though.

sebgrohn:
Added tests for Row, Bar, and Line chart that checks for the correct x/y axis ticks when `elasticX` / `elasticY` is enabled.

*Notes:*
* I added the missing Line chart tests for enabled `elasticX`, all three cases (all positive, mixed, all negative). I'm not sure, however, if they should be included in the template function `itShouldBehaveLikeARowChartWithGroup` or not...
* The Bar and Line chart tests only tested with mixed-sign values, so I renamed them and added new ones for all-negative values.

spec/row-chart-spec.js
```diff
@@ -352,11 +352,30 @@ describe('dc.rowChart', function () {
...
+                        chart.elasticX(true)
+                            .xAxis().ticks(3);
+
+                        chart.render();
+                    });
+
+                    it('should generate x axis domain dynamically', function () {
+                        var nthText = function (n) { return d3.select(chart.selectAll('g.axis .tick text')[0][n]); };
+
+                        for (let i = 0; i < xAxisTicks.length; i++) {
```
gordonwoodhull: Crazy... all your code works well, but for some reason the `let` here caused Jasmine to skip the row chart tests! Completely silent failure!

dc.js is still ES5 and will be for a while yet. @sebgrohn, did you have to do anything special to get the tests to run?

Of course, I noticed 58 commits later that my test count was way down. Easy enough to fix, but whoah that's a little scary if one ES6 keyword causes a file to get dropped entirely.

gordonwoodhull:
Thanks @sebgrohn, rebased and merged in 2.0 beta 32!

spec/row-chart-spec.js
sebgrohn: Wow, that's interesting (and scary)! Sloppy to let a `let` (sic.) slip in there; I am truly ES6 damaged.

I don't remember if I did something out of the ordinary, and I have the code and dev env set up on another computer I can't access right now... :o

sebgrohn:
Great!

PR #1165, opened 2016-06-16
apply slider input to reassign pie chart

wdingx:
Using slider input threshold to trigger data redistribution in pie chart


gordonwoodhull:
Sorry, this looks more like a PR from develop to master. I'd expect to see just a file or two added to web/examples. Try again?

Issue #1173, opened 2016-06-28
Support D3 v4


jfsiii:
D3 is being broken up into smaller modules for [v4](https://github.com/d3/d3/releases/tag/v4.0.0) (e.g. `d3-scale`, `d3-selection`, etc) and there are some API [changes](https://github.com/d3/d3/blob/master/CHANGES.md).

Discuss version support & approach. Should dc.js support d3 v3 and v4 or only one of them? One option to support both is a jQuery-style approach where 2.x supports v3 and 3.x supports v4.

ruhley:
I would love to move to d3 v4, but I know it would be a big task. I will have a quick go at upgrading on my branch to gauge how hard it would be. I would also love dc to take some of the best practices that d3 now uses:

> D3 4.0 is modular. Instead of one library, D3 is now many small libraries that are designed to work together. You can pick and choose which parts to use as you see fit. Each library is maintained in its own repository, allowing decentralized ownership and independent release cycles.


> You can also cat D3 microlibraries into a custom bundle, or use tools such as Webpack and Rollup to create optimized bundles. Custom bundles are great for applications that use a subset of D3’s features; for example, a React chart library might use D3 for scales and shapes, and React to manipulate the DOM. The D3 microlibraries are written as ES6 modules, and Rollup lets you pick at the symbol level to produce smaller bundles.


> The default UMD bundle is now anonymous. No d3 global is exported if AMD or CommonJS is detected. In a vanilla environment, the D3 microlibraries share the d3 global, even if you load them independently; thus, code you write is the same whether or not you use the default bundle. (See Let’s Make a (D3) Plugin for more.)

In short: modular, extendable and anonymous.

gordonwoodhull:
Nods. All good things.

Actually I don't think this is the most difficult task ahead for d3 - some of the design and debugging problems are really fiendish. 

And I think it's a pretty safe change. 

However, I do want to watch the adoption of v4 because it is likely there will be some gotchas and best practices to be learned as it rolls out.

(Does modularity require splitting dc.js into a dozen or more repos? Because that will definitely increase the cost of maintenance.)

XaserAcheron:
Presumably the modularity just translates to "simply include the parts of d3 that dc.js requires," without any sort of funkiness on dc's side.

I'm not sure if there's any part of dc.js itself that would really make sense to split off into a separate bit -- anonymous-ness, perhaps, though.

ruhley:
[...]

PR #1175, opened 2016-07-06
Convert project to ES6 with babel, webpack & eslint.

mtraynham:
This is a re-implementation of #937 (which I'm going to close).  So much of the code changed with the docs, that a rebase was almost too complicated to perform and I messed things up along the way...

I know @gordonwoodhull, you were a bit hesitant in the past, but I'll rebase this pretty often to keep it up to date in case we ever decide to migrate.

I should note that this is a starter for _what could be_.  Webpack has gotten incredibly expressive, and a modular/component layout is much more beneficial in the long term.  A few things I'd like to point out, we could potentially:
- Have a css (or less/scss) sheet for each chart type which is imported by the chart class itself (i.e. `import './bar-chart.scss`'.  This would potentially allow users to download only parts of the library and have it all still working.  [Angular-Material](https://github.com/angular/material/tree/v1.1.0-rc.5/src/components) is a good example of how this works.
- The above would also reduce the size of code for end users.  If only a bar chart is required, there should be no need to download 500 Kb of a library.  Shipping the library as modules would allow end users through npm to select only the modules they need and bundle them as necessary.
- ~~Supporting both d3 v3 & v4.  Webpack has ways to [rewrite access to global variables (such as d3)](https://webpack.github.io/docs/configuration.html#externals).  We could potentially ship a dc for v3 and v4 (related to #1173), by simply converting the `import * as d3 from 'd3'` to the v4 separate projects, and then have webpack rewrite this using an associative array mapping back to v3.  We would likely need separate branches/bower/package to support both releases, but the code would always be the same.~~

Other notes:
- I've updated travis to use Node 5 & 6, as 0.10 is EOL.  Node 4 could be supported, but it doesn't relax peer dependencies and grunt-eslint (depends on eslint 3) has some conflicts with the very well thought out [eslint-config-airbnb](https://www.npmjs.com/package/eslint-config-airbnb) and it's dependencies.

Known issues:
- A test that spies on dc.redrawAll, I disabled [here](https://github.com/dc-js/dc.js/compare/develop...mtraynham:es6?expand=1#diff-05b834af2af9e8efcfe98352d167b9e0R764), which relates to https://github.com/jasmine/jasmine/issues/943.
- ~~Crossfilter2 vs. Crossfilter resolution.  For npm, it should be `import * as crossfilter from 'crossfilter2'`, but for globals it would be `import * as crossfilter from 'crossfilter'`.  Webpack `externals` (I mentioned above) will allow us to do this.~~

Breaking changes:
dc.disableTransition, dc.enableDebugLog and dc.dateFormat are now getter/setters.


gordonwoodhull:
Thanks @mtraynham! 

Porting to d3.js v4 is not just a name substitution. There are breaking changes on most components (notably `d3.selection`). It doesn't sound like it will be so difficult but it will touch most of the code.

Are Webpack import statements compatible with Rollup import statements? I know they are both ES6 but are they the same ES6? Honest question. I ask because upgrading to d3.js v4 (#1173) is a higher priority for me, and we could switch to Rollup as part of that effort.

It looks like Rollup is somewhat more powerful than Webpack, in that it can include only the needed code from a module. I don't know how much this affects dc.js, given that it consists mostly of big class-like functions. 

Sorry I don't plan to upgrade to ES6 soon. The cost of transpilation on debugging concerns me too much. And fixing bugs is more important to me than using the latest tools.

But thanks for doing this. People can use your fork if they want an ES6 dc.js!

mtraynham:
Just going to address some of your points 😄 

> Porting to d3.js v4 is not just a name substitution

Can you point me to the breaking changes?  I saw the change log, but must have missed the breaking changes section.  I thought it was mostly package separation of concerns and improvements to the API.
To clarify, my intention here is:
```
import {select, selectAll} from 'd3-selection'; // v4
import {select, selectAll} from 'd3'; // v3
```
You could write the `v4` version and we could have Webpack translate `d3-selection` to `d3` if necessary for `v3`.

> in that it can include only the needed code from a module.

Webpack is the same as they both support the CommonJS require spec; it's just slightly different because Rollup has ES6 imports builtin, whereas Webpack uses babel as a preprocessor to get all the benefits of ES6.  You can `import extent from 'd3/src/extent'` just the same and it will only include that in your bundled build.  If you don't *import* a module, Webpack won't include it.  Webpack 2.0 will include [tree-shaking](https://github.com/webpack/webpack/pull/861#issuecomment-149997270) which will drop unused modules, further reducing the size of the bundle.  I would tend to believe Webpack is a bit more battle-tested, with community size, number of contributors and feature set, but the two are likely interchangeable.

Webpack/Rollup is a bit beside the point.  People should be able to import charts directly (e.g. `import barChart from 'dcjs/src/bar-chart'`) when installing from npm and use Webpack/Browserify/Rollup, or whichever module bundler of their own choosing.  Webpack here is simply to provide a build for bower that mimics what we currently have, a browser global library.   It looks as if d3 v4 doesn't even support bower directly.

> Sorry I don't plan to upgrade to ES6 soon.

In all fairness, Rollup is an ES6 bundler and d3 v4 is entirely dependent on the [ES6](https://github.com/d3/d3-array/blob/master/index.js#L1) spec; so code created by Rollup is going to be transpiled to a bundle in a similar fashion.  This PR only contains two other things from ES6 which Rollup can't handle, `const`/`let` and arrow functions, both of which are trivial to decypher once converted back to ES5.

For a further comparison on Webpack vs. Rollup, the author of Rollup [has a long comment here](https://github.com/rollup/rollup/issues/219#issuecomment-150842108).  To kind of refute one of his points, where he says Webpack is not a great library bundler, it's really no different.  You have an index.js file which just exports (or re-exports) everything, just like the d3-selection index file I posted above.

gordonwoodhull:
I haven't read everything about v4 but there are numerous mentions of breaking changes. That's what happens when you make improvements to the API, and that's the way it should be. 

@ruhley might be able to elaborate because he started trying to do the port on his fork.

I understand that you've mostly done a mechanical translation which only introduces changes which are easy to debug after transpilation. But once we open that door then stuff will enter which is more difficult. Like destructuring assignment and so on. So why not go the route where we really are just using ES5 with modules, like d3 did? I am pretty sure that is the extent of their ES6 usage, but please correct me if I'm wrong.

I do encourage you to continue your fork and everyone who wants to use it, should. I'll keep this PR open as a pointer to your repo, but I do not intent to merge this until ES6 reaches the browsers.

ruhley:
I had a quick attempt at converting to d3 v4 and you can see my results at https://github.com/dc-js/dc.js/issues/1173#issuecomment-229219587. You can see in my comment that some things are just a rename and **_seem_** to work without any other changes, but some things have completely changed. So I think support d3 v3 and v4 with the same library is a bad idea. It could start out simple, but would most likely end up in a big mess.

In saying that I love the rest of the stuff you have done and I am in full support of using ES6. Your changes don't seem to change any logic and if the tests still pass then I would be happy for the upgrade.

mtraynham:
@ruhley Good to know and thanks 😊 !  I wonder if it's mostly the layout stuff that's incompatible... I noticed that `d3.layout.pie` was also renamed to `d3.pie` and exists in the `d3-shape` package.  I can't seem to find breaking changes documentation anywhere though...

I made this table as a [v4 migration reference](https://github.com/dc-js/dc.js/issues/1173#issuecomment-231616687) though, so maybe you will find it useful.

kum-deepak:
Great idea!

ialarmedalien:
I've been using this PR as a base for some other code -- it's great to be able to work with dc.js in ES6; thanks for the work you've done on it! I noticed a tiny discrepancy in `utils.js`, looks like a weird copy/paste artefact around line 272:

```
let _iunter = 0;
utils.uniqueId = function () {
    return ++_iunter;
};
```

The variable `_iunter` is `_idCounter` in the master branch. Not important, but it just looks a bit odd!

mtraynham:
Thanks @ialarmedalien! Probably occurred during a regex removal of the `dc` global throughout the code.  I'll rebase with it fixed.

gordonwoodhull:
Sorry, I am closing this to be clear dc-js/dc.js is not going this way. We're following d3 and using what's supported by the browsers, and I hope to move to `require` and rollup and modularize the charts in the same way that d3 is a lot of smaller, independent modules now.

I know, everyone loves transpilation and using the most modern everything, but it really causes a lot of trouble when trying to debug code on any developer tools !== Chrome. Source maps are really tricky and apparently no one outside of Google cares about them.

mtraynham:
@gordonwoodhull Thanks for the feedback.  

I'm not sure I completely follow the reasoning.  Rollup transforms the source in a manner that is not 1 to 1 with the originally source code, instead providing a bundle.  I concede that d3 doesn't use any other features of ES6 and the bundle is likely more readable to a developer, but it [does use the ES6 import/export syntax and modules](https://github.com/d3/d3-array/blob/master/src/bisector.js#L1) to work with Rollup instead of just concatenating everything.

So my question is, are you just concerned about the usage of `const`/`let` & arrow functions which are transformed by babel and the usage of Webpack instead of Rollup?

If that's the case, it would still be nice to ship the source as ES5 with ES6 import/export syntax (which has been done as part of this PR), so users can benefit from module loaders that tree-shake unused code.  D3 does this, and if you `npm install d3-array`, you can see an example.

The install has both a `build` (the bundle) and `src` (the original source).


Within the `package.json` file it has references to `main`, `module` and `jsnext:main`, which tells module bundlers where they can pull source code from.

gordonwoodhull:
Yes, that is what I meant, `import`/`export` rather than `require`. (I've corrected above.)

What you're describing is exactly what I mean by modularization of the build. I want to follow what d3v4 did.

The big win is that when someone makes their own version of a chart, people can use it immediately instead of waiting for it to be merged into dc-js. This is especially important when people make radically different versions of charts which break tests and previous functionality. Potential users of the new features get very impatient with me for not merging stuff I can't safely merge without breaking 100s of other applications. They could use the fork instead. 

Also, there will be no real difference between charts contributed to the dc-js organization and charts hosted elsewhere. 

The downside is that there will be >24 tiny libraries with one class each. I sort of understand how import/export work but I haven't yet learned how Mike deals with versioning and dependencies. [He has said that it's pretty tedious to update patterns across repos.](https://github.com/d3/d3/issues/3284#issuecomment-375674266)

ialarmedalien:
I started some d3-esque modularisation of the base mixin of this fork of the codebase -- split it into 16 smaller modules that get gathered up to create `base-mixin.js`. It was not possible to completely eliminate crossover/dependency between different modules and there are some places where there are dependencies on other dc.js modules (e.g. `core.js`).

PR #1177, opened 2016-07-06
Fixed numberDisplay tweening getting stuck on Infinity values

### LINKED ISSUE #1176: numberDisplay tweening breaks on divide-by-zero
Welp! Another bug of sorts -- currently seeing this on beta-31.

The gist is that setting the valueAccessor of a numberDisplay to a function that can divide by zero will cause tweening to malfunction when filtering away from Infinity -- the display gets stuck at NaN until the numberDisplay is redrawn once more.

Good ol' JSFiddle time:
https://jsfiddle.net/fzkjch04/1/

To reproduce, select and deselect the "Not Rush" slice -- you'll notice the numberDisplay labeled "borks" will, true to its name, get stuck at NaN on the way out. Just calling dc.redrawAll (i.e. clicking the "Redraw Charts" button) is enough to get it un-stuck.

The second number display (labeled "works") has its valueAccessor modified to explicitly return NaN when the divisor is zero -- this makes it work just fine for whatever reason.

[aside: Although I technically may be able to work around this in my own app, I've gotten into the habit of converting NaNs to positive or negative Infinity for table sorting purposes, meaning there are cases where a value of Infinity makes sense in context. I don't want to lose that.]

XaserAcheron:
Closes #1176. Basically just added `-Infinity` and `Infinity` to the list of values to alias to zero (along with all "falsy" types) for the first parameter of the `d3.interpolateNumber` call. Works like a charm.

### DIFF
--- a/src/number-display.js
+++ b/src/number-display.js
@@ -99,7 +99,9 @@ dc.numberDisplay = function (parent, chartGroup) {
             .duration(_chart.transitionDuration())
             .ease('quad-out-in')
             .tween('text', function () {
-                var interp = d3.interpolateNumber(this.lastValue || 0, newValue);
+                // [XA] don't try and interpolate from Infinity, else this breaks.
+                var interpStart = isFinite(this.lastValue) ? this.lastValue : 0;
+                var interp = d3.interpolateNumber(interpStart || 0, newValue);
                 this.lastValue = newValue;
                 return function (t) {
                     var html = null, num = _chart.formatNumber()(interp(t));


gordonwoodhull:
@XaserAcheron, I'm adding your example under transitions/ so that we have a visual test (along with a Jasmine test).

XaserAcheron:
Coolness -- I ought to figure out how to do this myself in the future to save you the trouble.

On that note, I'm exceptionally glad the Rush example is now canon. ;P

gordonwoodhull:
Thanks @XaserAcheron, rebased and merged in 2.0 beta 32!

I'm also very happy to include the Rush example. [It's here.](http://dc-js.github.io/dc.js/transitions/number-transitions.html)

PR #1181, opened 2016-07-09
Transition line-chart dots along with the lines

paulmach:
When updating data for line-charts, the line updates using a d3.transition(). The dots (if enabled) do not use a transition. This PR adds transitions to dots. See the updated series example.

I'm not exactly sure what test to write. Maybe check location, d3.timer.flush(), check location is different?
Also, probably should not update the example, but leaving in for now as a demo for anyone reviewing this.

### DIFF
--- a/src/line-chart.js
+++ b/src/line-chart.js
@@ -288,6 +288,7 @@ dc.lineChart = function (parent, chartGroup) {
                     .attr('r', getDotRadius())
                     .style('fill-opacity', _dataPointFillOpacity)
                     .style('stroke-opacity', _dataPointStrokeOpacity)
+                    .attr('fill', _chart.getColor)
                     .on('mousemove', function () {
                         var dot = d3.select(this);
                         showDot(dot);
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Thanks @paulmach! I've started some work in this area in a branch, [feature/area-transitions](https://github.com/dc-js/dc.js/tree/feature/area-transitions), and this will be helpful to compare and merge. 

Also your changes are interface-compatible, whereas that branch starts to tackle the big problem of joining by key and transitioning stacks in and out. So it might be good to apply your changes first in 2.0.

As for tests, I don't think automated testing is feasible for transitions. So I'm starting to create as many "eyeball tests" as I can, [here](http://dc-js.github.io/dc.js/transitions/). I guess that's similar to what you've done with the series example.

paulmach:
Thanks for the response. I would like to apply this to 2.0. It would nice to make use of this update for a project at work.

The "stacks" [area transition](http://dc-js.github.io/dc.js/transitions/area-transitions.html) example illustrates the problem. The dots move, while the lines transition. So I think that's a good visual test. 

I just removed the change to the stacks example as that was just for me before you pointed me to the area transitions example/test.

src/line-chart.js
```diff
@@ -288,6 +288,7 @@ dc.lineChart = function (parent, chartGroup) {
                     .attr('r', getDotRadius())
                     .style('fill-opacity', _dataPointFillOpacity)
                     .style('stroke-opacity', _dataPointStrokeOpacity)
+                    .attr('fill', _chart.getColor)
```
paulmach: FYI, without this the dots will be black and transition to the color. I caught this visually but [this spec](https://github.com/dc-js/dc.js/blob/master/spec/line-chart-spec.js#L462) will also catch it.

gordonwoodhull:
Rebased and merged in 2.0 beta 32. Thanks @paulmach!

paulmach:
Thanks @gordonwoodhull

Issue #934, opened 2015-05-14
capped mixin shouldn't use .top(N)


alan-unravel:
When using ordering and cap on a row chart I would assume the top-n values would be included in the cap but instead the values are takes in initial order from the grouping and then orders the result.

alan-unravel:
This issue can be prevented by using a '.order' function on your group before passing it to your dc chart and using chart.cap as usual.

gordonwoodhull:
Hi @alan-unravel. This is something @blairn and I were discussing on gitter. 

I consider it somewhat of a design bug. 

What happened here is that originally the capped charts just used `group.order` and `group.top`, but then
 1. `.all()` still needed to be called in order to fill in the Others, so the efficiency is lost
 2. We introduced `chart.ordering` because `group.order` only has access to the value, not the key - so this is also confusing.

Since `chart.ordering` is strictly more powerful, and using `group.top()` is kind of a blemish on the API, (especially for folks who don't use crossfilter or who use fake groups), it would probably be better just to always use `.all()` and `chart.ordering`. I'd like to hear more opinions on this, though.

alan-unravel:
I think that would make alot more sense!

gordonwoodhull:
I've noticed another odd usage of `.top(N)` in the bubble mixin. This is presumably in order to sort the data so that smaller bubbles will display on top of larger bubbles, but it won't work in general, because the bubble radius can be based on anything. 

We have `sortBubbleSize` for this, but it should be moved up into the mixin, and this should revert to `.all()`.

Another silently breaking change, fun!

gordonwoodhull:
Finally, `group.top(1)` is also used in the number display, when a group instead of a groupAll is passed for the `.group()` parameter. It should be backward-compatible to default
```js
_chart.ordering(function (kv) { return kv.value; });
```

and take the maximum of `group.all()` according to the ordering, so I'm implementing this in 2.1.2

PR #1184, opened 2016-07-22
#934 - Removing top from cap mixin

macyabbey:
Before, the cap mixin used crossfilter.group.top(_cap) before applying the ordering function. This produced really strange results when you use a non-default ordering function.

The order of group.all by default is ascending value.  The order of dcjs's ordering function is ascending key by default.  So if you left ordering default, you would get fairly expected results. The largest value groups, sorted by key.  

But if you use your own ordering function, you get very strange results, because your ordering function does not take affect until the biggest set has already been determined by group.all().top(). 

Added a new test which fails with current capping logic because of this problem. Changed implementation of data function in cap which resolves this new test.  Changed existing tests which tested for incorrect results that used to be produced.

### DIFF
--- a/src/cap-mixin.js
+++ b/src/cap-mixin.js
@@ -46,12 +46,38 @@ dc.capMixin = function (_chart) {
         return _chart.valueAccessor()(d, i);
     };
 
+    /**
+     * Base function orders by crossfilter's group.all.
+     * This returns all groups, in ascending order, by value.
+     *
+     * This function is expected to return items in the same direction,
+     * ascending but the N biggest where N is cap.
+     *
+     * Since computeOrderedGroups is using crossfilter's quicksort, the
+     * sorted items will always be in ascending order.
+     *
+     * This means if we cap, we need to take from N to the end of the array.
+     *
+     *  Index
+     *  0-------N-------length
+     *
+     *  Value
+     *  small -> big
+     *
+     */
     _chart.data(function (group) {
         if (_cap === Infinity) {
             return _chart._computeOrderedGroups(group.all());
         } else {
-            var topRows = group.top(_cap); // ordered by crossfilter group order (default value)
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Thanks @macyabbey, this is great. It's been something we've been talking about for a while so I'm glad to see someone take the plunge.

I think it is a breaking change even though it's hard to argue that anything was good about the old behavior. So I think this will need to go into the 2.1 (develop) branch rather than 2.0.

I'd be glad to hear arguments either way.

macyabbey:
@gordonwoodhull I agree this is  a breaking change. What do you think about updating the cap documentation? I think we should make it a lot more clear about how the cap mixin interacts with the ordering function, and the underlying crossfilter group order function.

gordonwoodhull:
Sounds like the right way forward - update the documentation for 2.0 with a note saying basically "We know this stinks but here is the way it currently works: group.order then top(K) then baseMixin.ordering. This behavior is deprecated and will change in 2.1." Would you be willing to take a stab at that in a separate PR?

Both branches are active, so we could commit both at once.

src/cap-mixin.js
```diff
@@ -46,12 +46,38 @@ dc.capMixin = function (_chart) {
...
 
+    /**
+     * Base function orders by crossfilter's group.all.
+     * This returns all groups, in ascending order, by value.
+     *
+     * This function is expected to return items in the same direction,
+     * ascending but the N biggest where N is cap.
+     *
+     * Since computeOrderedGroups is using crossfilter's quicksort, the
+     * sorted items will always be in ascending order.
```
gordonwoodhull: I guess not always, see #1269

gordonwoodhull:
Merged in 2.1.3 and the option to take from either the front or the back of the sorted array added in 2.1.4.

PR #1194, opened 2016-08-19
Ref #1193 - Move gh-pages to /docs directory

### LINKED ISSUE #1193: Remove gh-pages branch in favor of master/docs directory.
https://github.com/blog/2228-simpler-github-pages-publishing

I think this would remove some of the weird Grunt bloat around maintaining the web directory on master and pushing it to the `gh-pages` branch.  The only caveat I can think of is if the build is bad, the docs might be bad.

mtraynham:
I did the following to resolve #1193:
- Removed the `docs/coverage.js` file.  I could not find any references to this file throughout the repo and it was last modified in 2014.
- Move the entire `web/` directory to `docs/`
- Selective find and replace for 'web/' to 'docs/'.
- Updated Gruntfile to replace references of `web` with `docs`.
- Removed the grunt-gh-pages package and it's Grunt tasks.

@gordonwoodhull, following this PR, you'll need to go to the Settings for the project and change it to use the `master/docs` directory.  Once that is completed and verified, you could safely delete the `gh-pages` branch.


gordonwoodhull:
Okay, I can copy the markdown to the old location, to not break the thousands of links to https://github.com/dc-js/dc.js/blob/develop/web/docs/api-latest.md - that's easy.

But the drawback to this approach is that we now have to check in **all** the HTML documentation. That is going to cause a lot of dangerous noise on every merge.

mtraynham:
That is true, I hadn't considered the number of git objects that would affect the master/develop merges.

gordonwoodhull:
Me neither. ☹️ 

I guess I will close this. grunt-gh-pages was updated and it's no longer such a drag. (We have other dependency problems now, like grunt itself.)

Issue #961, opened 2015-07-14
Unable to mouseover on scatter plot and display the title rendering


louishenri13:
The title does not render when I use scatter plot without brush. It seems to be working with all the other graphs but scatter plot. This person has had the same question for almost 2 years and it does not seem to be resolved: "Tooltip not rendered on Scatterplots #463" (amergin, 2013). I was wandering if someone could potentially fix this issue.

Thank you,
Louis-Henri

ericgross:
:+1:

maxwang7:
Any updates on this issue? If not, I'll work on a PR to fix it.

gordonwoodhull:
Yes, please contribute a fix! This is a community-driven project.

maxwang7:
Cool. Thoughts on why this could be happening?

gordonwoodhull:
Not much of the actual rendering code is shared between charts. Looks like the title was never implemented here - take a look at how it's appended to the corresponding elements in the other charts.

daniel-gall:
Submitted pull request for this.

https://github.com/dc-js/dc.js/pull/1198

daniel-gall:
Closed original pull request and opened:

https://github.com/dc-js/dc.js/pull/1200

PR #1198, opened 2016-08-31
Scatter plot titles

daniel-gall:
Candidate to address issue 554:

https://github.com/dc-js/dc.js/issues/554

### DIFF
--- a/src/scatter-plot.js
+++ b/src/scatter-plot.js
@@ -22,7 +22,6 @@
  * @return {dc.scatterPlot}
  */
 dc.scatterPlot = function (parent, chartGroup) {
-    var TOOLTIP_G_CLASS = 'dc-tooltip';
     var _chart = dc.coordinateGridMixin({});
     var _symbol = d3.svg.symbol();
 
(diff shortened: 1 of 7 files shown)


daniel-gall:
Correction: wrong issue.  This is the issue addressed:

https://github.com/dc-js/dc.js/issues/961

gordonwoodhull:
I'll review this more thoroughly when I return from vacation next week. A few notes at first glance:

 - It doesn't look like `tooltips` is used - `g` is created from it and then unused. I don't think the title element needs a drawing layer.

 - There aren't really stacks or layers for the scatter plot - these are used for stacked charts like the line chart and bar chart. So I think that's another unnecessary complication. Really `renderTitle` is doing all the work here.

 - Please add a couple of tests to [the Jasmine test](https://github.com/dc-js/dc.js/blob/develop/spec/scatter-plot-spec.js) - one for the default behavior, and one for a custom title.

Overall this is a helpful contribution but I think it could be a lot simpler. Thanks!

daniel-gall:
Thanks for the onceover and comments.

Will look at tooltips.

I tried getting rid of the layers, but it didn't work without them.  I'll have to look into why that might be. Plus I was borrowing from lineChart's code.

I didn't add any tests because the lineChart only had one test - on a meta-feature, not core functionality.   if a bool is set, don't render titles.  I didn't know if that was handled by the base mixin or something.

Sorry for being a first time contributor.  At least I read the guidelines? ;)

daniel-gall:
Will make another PR with the updated scatterPlot.js.  No Jasmine tests incoming yet.  I want to get this in then take the time to understand Jasmin so I don't submit something terribad.

PR #1200, opened 2016-09-01
Scatter plot titles

daniel-gall:
Updated to remove cruft from title code borrowed from lineChart.

https://github.com/dc-js/dc.js/issues/961


gordonwoodhull:
Merged for 2.0. Thanks @daniel-gall!

PR #1043, opened 2015-11-20
Grouped bar chart

gazal-k:
Modification of https://github.com/dc-js/dc.js/pull/679
- Hopefully I have resolved most of the conflicts with dc.js mainline.
- Resolved a few issue in rendering multi/grouped bar chart.
- I've added some basic tests for multi/grouped bar chart.

Pending:
- not sure what functionalities expected of bar charts in general is missing in the multi/grouped version.
- multi/grouped row chart.

Please feel free to suggest changes or fork this and fix further.

PR #1212, opened 2016-10-10
Gazal grouped bar

cdoe0108:


gordonwoodhull:
Thanks but most of this branch already exists as PR #1043. (Basically this branch just adds an example to that work.) I'm guessing this was accidental so I'll close this.

(Sometimes I think GitHub makes it too easy to create pull requests, we get a lot of spurious PRs over here.)

PR #1213, opened 2016-10-11
Use crossfilter2 in footer, on the AMD path.

vprus:
I've tried to use current DC inside react, using https://github.com/facebookincubator/create-react-app, and DC could not find crossfilter. Looks like 'crossfilter' was changed to 'crossfilter2' on one path, but not another. I have no idea why react build process uses AMD modules, but this patch fixes the error.

### DIFF
--- a/src/footer.js
+++ b/src/footer.js
@@ -15,7 +15,7 @@ dc.crossfilter = crossfilter;
 
 return dc;}
     if(typeof define === "function" && define.amd) {
-        define(["d3", "crossfilter"], _dc);
+        define(["d3", "crossfilter2"], _dc);
     } else if(typeof module === "object" && module.exports) {
         var _d3 = require('d3');
         var _crossfilter = require('crossfilter2');


gordonwoodhull:
Thanks @vprus, over in npm-land we don't have a problem with the module name being different from the filename, but I guess the create-react-app uses webpack so we run into #1214. 

We may have to fix this on the crossfilter side.

gordonwoodhull:
@vprus, does the `resolve` config described in #1214 also work for you? 

This solution breaks vanilla AMD, and I'd rather keep the `crossfilter.js` filename for [crossfilter/crossfilter](https://github.com/crossfilter/crossfilter). Although it's unfortunate that having a different module name and filename causes so much confusion.

gordonwoodhull:
I'm going to write this off as a configuration issue, unless I hear otherwise.

It's unfortunate that the module name is different from the filename (and the function object name), but that's where we are at. Maybe someday the community fork of crossfilter will change enough to merit its own name, but for now it is a drop-in replacement

gordonwoodhull:
As of dc.js 2.0.3 / 2.1.6, we'll break requireJS compatibility in favor of webpack compatibility. 

See #1304 if this creates problems for you.

PR #1229, opened 2016-12-01
Fix legend item wrapping for _horizontal

alexnb:
Fixes bug, which lead to not wrapping the legend if the _legendWidth was exceeded.

### DIFF
--- a/src/legend.js
+++ b/src/legend.js
@@ -93,7 +93,7 @@ dc.legend = function () {
         itemEnter.attr('transform', function (d, i) {
             if (_horizontal) {
                 var itemWidth   = _autoItemWidth === true ? this.getBBox().width + _gap : _itemWidth;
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Thanks @alexnb and sorry for the slow response. 

I take it the bug exhibited as the row wrapping too late, i.e. an extra item gets displayed on each row and pushes out of bounds?

I'm going to try adding a failing test that will get fixed by this.

gordonwoodhull:
I was able to confirm the bug in a test, and merged your fix for v2.0. Thanks @alexnb!

It was a sophisticated kind of off-by-one error, and in reviewing this I noticed that we deliberately didn't test this feature well enough when merging it. Partly I've just gotten better at writing tests (or more patient). Partly it's messy to test anything involving text because the sizes can come out wildly different (especially on PhantomJS).

PR #1233, opened 2016-12-02
Update node dependencies to latest.  Fix styling issues

mtraynham:
Update the node dependencies to latest.  Fix any style issues with JSCS/JSHint bump.


gordonwoodhull:
Thanks @mtraynham! This is very helpful.

I hope the churn from jscs eventually settles down. It creates a lot of noise in the commit history. On the plus side, I'm getting really good at `git blame sha^ -- foo.js` 😉 )

mtraynham:
Ha.  Well, JSCS has now been deprecated in favor of ESLint, so I don't think there will be many more iterations of this :stuck_out_tongue_winking_eye:

gordonwoodhull:
Uh oh, I bet ESLint will complain about a subtly very different set of style quirks. 😱

gordonwoodhull:
Thanks @mtraynham, merged for 2.0!

ialarmedalien:
FYI: I converted the current JSCS and JSHint configs into ESLint using [polyjuice](https://github.com/brenolf/polyjuice), and then linted the current master branch. Some of the results can be fixed automatically, and some could probably be ignored. I [dc-ified the results](https://bl.ocks.org/ialarmedalien/0a4bf25ffc0fb96ae569a20f91957bc1)  (because code linting is much more fun when there are graphs to look at!) and included the eslint config that polyjuice generated.

mtraynham:
@ialarmedalien this is really impressive 👍  That row chart scrolling solution is pretty clever, haha.  Very cool.

gordonwoodhull:
That is very cool. Yeah, really neat way to do a scrolling row chart. A lot of people ask how to do that.

In terms of the content of the errors, I'd mostly be concerned about the variable shadowing and the jsdoc errors - the rest are stylistic. I guess unused variables could indicate a bug too. We've historically had lighter restrictions on the specs, and it's probably not worth worrying about variable shadowing there, because specs are meant to be copied and pasted.

I'd definitely accept a PR with an upgrade to eslint. I would want to coordinate, though, since this is likely to touch every line in the source, and I'm planning the d3v4 upgrade for EOY when things get quiet at work (which will also touch every line, making a merge painful).

gordonwoodhull:
finally moved to eslint - on develop for 3.1

PR #1234, opened 2016-12-05
Add call to xAxisPaddingUnit for stack-mixin when calculating max/min

stillesjo:
Hi!

We need to get xAxisPaddingUnit when calculating max/min for stack-mixin in order to get xAxisPadding with other units than day to work. 

Extends: ea511e06b2acd3fa62723de3ec7337b2b449a0bc
Reference: #892

### DIFF
--- a/spec/bar-chart-spec.js
+++ b/spec/bar-chart-spec.js
@@ -949,7 +949,43 @@ describe('dc.barChart', function () {
             });
         });
     });
+    describe('with elasticX and x-axis padding', function() {
+        var date = makeDate(2012, 5, 1);
+        beforeEach(function () {
 
+            var rows = [
+              {x: date, y: 4},
+            ];
+            data = crossfilter(rows);
+            dimension = data.dimension(function (d) {
+              return d.x;
+            });
+            group = dimension.group().reduceSum(function (d) {
+              return d.y;
+            });
+            chart = dc.barChart('#' + id);
+            chart.width(500)
+              .transitionDuration(0)
+              .x(d3.time.scale())
+              .elasticY(true).elasticX(true)
+              .dimension(dimension)
+              .group(group)
+            chart.render();
+        });
+        it ('should render the right xAxisMax/Min when no padding', function () {
+            expect(chart.xAxisMin()).toEqual(date);
+            expect(chart.xAxisMax()).toEqual(date);
+        })
+        it ('should render the right xAxisMax/Min when 10 day padding', function () {
+          chart.xAxisPadding(10)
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Whoops! Looks like that's the only other use of `xAxisPadding`.

I didn't realize we weren't testing `xAxisPaddingUnit` - the tests for the add and subtract utilities are great could you add some tests for the charts too, please? At least one stacked and one unstacked test, to hit both code paths. 

Hopefully there are already some tests for the bubble chart, box plot, or scatter plot that use a date x axis - I know there are some for the line chart and bar chart.

spec/bar-chart-spec.js
```diff
@@ -949,7 +949,43 @@ describe('dc.barChart', function () {
...
+        })
+        it ('should render the right xAxisMax/Min when 10 day padding', function () {
+          chart.xAxisPadding(10)
+          let expectedStartDate = d3.time.day.offset(date, -10);
+          let expectedEndDate = d3.time.day.offset(date, 10);
+          expect(chart.xAxisMin()).toEqual(expectedStartDate);
+          expect(chart.xAxisMax()).toEqual(expectedEndDate);
+        });
+
+        it ('should render the right ')
```
gordonwoodhull: Thanks @stillesjo. 

Wow, Jasmine just silently hangs when there's an incomplete test like this one - we should fail the build if there are any pending tests.

gordonwoodhull: No problem, I understood that. I'll wait for you to comment before I merge. 

Travis went ahead and tested it, and I commented because I was alarmed that our current tools will silently drop tests on the floor. There was another time this happened due to use of an ES6 feature but `let` doesn't appear to be a problem this time. 

I wonder if we need to upgrade our build tools yet again. :-/

stillesjo: Ah, that might explain it. Added `expect(false).toEqual(true);` while testing earlier and all tests passed successfully. Changing to var makes tests fails as they should. I'll try to avoid ES6 syntax in the future 😄

gordonwoodhull: Heheh. Oh I thought it was this dangling `it()` with no callback and *not* the `let` - or maybe it depends on the environment, or the moon. 

Anyway, there should be more than 1000 tests, if that number drops there's probably something wrong. It sucks that we have to second guess the test runner, it's supposed to be the other way around!

gordonwoodhull:
Thanks @stillesjo, I wrapped up the tests and it's in dc.js 2.0!

PR #1237, opened 2016-12-07
Bubble Chart: selectAll should be select when manipulating a enter/update selection

mtraynham:
`selectAll` in these cases is going to screw up the current enter/update selection, which results in some nodes not getting updated.  These should really be `select`.  I feel like this has come up before, but I can't seem to find the particular PR or conversation about it.

### DIFF
--- a/src/bubble-chart.js
+++ b/src/bubble-chart.js
@@ -117,7 +117,7 @@ dc.bubbleChart = function (parent, chartGroup) {
             .attr('fill', _chart.getColor)
             .attr('r', 0);
         dc.transition(bubbleG, _chart.transitionDuration(), _chart.transitionDelay())
-            .selectAll('circle.' + _chart.BUBBLE_CLASS)
+            .select('circle.' + _chart.BUBBLE_CLASS)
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Yes, #1032 is the previous issue. And you found similar problems in four other charts. So I guess we are not testing updates of at least these six charts. 😱

gordonwoodhull:
It's not that we don't test update. It's that the code currently works as long as the data is updated in-place, as crossfilter always does. 

But we definitely shouldn't assume that. 

On a hunch, I created a (rather excessive) bl.ock demonstrating the problem:
http://bl.ocks.org/gordonwoodhull/raw/14c623b95993808d69620563508edba6/

Everything is fine with the top two charts, which use straight crossfilter. But in the second two charts we clone the keys and values each time `group.all()` is called. Wham, no updates get through to those charts, because they still have the original objects and don't see the changed reductions.

gordonwoodhull:
added test and merging for 2.0 - thanks @mtraynham!

PR #1279, opened 2017-02-11
Pie-chart radius as per resizing of window.

fatangare:
During resizing of pie-chart, chart is redrawn for given width and height but it's radius is not changed making it to be cropped. Simple solution is to check if diameter is bigger than container...if yes, adjust radius accordingly.

### DIFF
--- a/src/pie-chart.js
+++ b/src/pie-chart.js
@@ -85,8 +85,9 @@ dc.pieChart = function (parent, chartGroup) {
     };
 
     function drawChart () {
-        // set radius on basis of chart dimension if missing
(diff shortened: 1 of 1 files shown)


tttp:
Hi,

@fatangare have you tried with the responsive width(0) way? Does it behave properly in that case?

gordonwoodhull:
I doubt that would help, however another workaround is just to not set the radius and let it be derived from the width/height.

I'm in favor of this PR, sorry I have not yet had time to merge it. Since it doesn't change the API it should be okay for the 2.0 branch.

PR #1282, opened 2017-02-21
make bracket selectors work with quoted values

IwfY:
This change allows coordinate grid chart's root element to be something like '.chart[foo="5bar"]'.

Before only selectors like '.chart[foo=bar]' worked but if the value starts with a number it must be quoted which was not possible to use with dc.js due to the way the clip path id was generated.


gordonwoodhull:
Thanks @IwfY! Merged to 2.0.2 (and will be merged to later versions).

PR #1284, opened 2017-02-27
fix url for images

tttp:
Ask discussed and requested by @gordonwoodhull 
I'm not too sure what is the branch I should pull into, but given it's a onliner change on a file no one else touched, probably not a big issue ;)

### DIFF
--- a/web/ep/index.html
+++ b/web/ep/index.html
@@ -18,7 +18,7 @@
 
 <script>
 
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Thanks @tttp! Fixed in 2.0.2 (and will be merged to later versions).

PR #1291, opened 2017-03-26
update to the correct path of d3-queue

idanb11:


gordonwoodhull:
I'm confused by this - the example appears to be working without the change?

Nevertheless, we should pull our development dependencies from npm, so I'm fixing this that way in 2.0.2.

Issue #1294, opened 2017-04-07
Tests are broken on develop


mtraynham:
https://travis-ci.org/dc-js/dc.js/builds/216810258

Looks like the cap mixin change affected the ordering of results.  Is that intentional?  Was going to rebase a branch or two, but noticed it was broken.

gordonwoodhull:
Yes, the change is intentional, sorry I have not fixed the tests yet. This change is not part of any release yet.

When we eliminated `group.top()`, this changed the ordering of many of the charts. In a few cases, the ordering was broken already (because the group order and `chart.ordering()` didn't agree), so it didn't matter. 

But in most cases, we can emulate the old `group.top()` ordering by sorting reverse by value. That's usually what people want for row charts, and what this change enables.

I'd suggest rebasing against the stable branch (`support/2.0`, currently 2.0.2). Or if you feel like fixing the tests, that's welcome too!

mtraynham:
I'll provide a fix as long as the changes were intentional 😄

gordonwoodhull:
That would be awesome, a second pair of eyes on the tests is always helpful.

mtraynham:
Closing per #1295.

gordonwoodhull:
Thanks!

PR #1295, opened 2017-04-07
Fix tests after cap mixin changes

mtraynham:
Ref #1294


gordonwoodhull:
Thanks @mtraynham!

I've cleaned up and clarified the tests even more, to make it clear that the default behavior should now be descending order by value, but you can still do alphabetical order if you want.

It will be tough to reproduce the old behavior where the capping was by descending value, but sorting was alphabetical, but that was ridiculous and I can't imagine anyone liked it that way.

Merged in 2.1.6

PR #1323, opened 2017-07-24
Faster dc.filterAll()

### LINKED ISSUE #1322: Suggestion for faster dc.filterAll()
I have built a dashboard with ~10 graphs and lots of data (which, unfortunately, I cannot share). Calling `dc.filterAll()` on it is pretty slow, taking 10-15 seconds. The default implementation of `dc.filterAll()` calls `.filterAll()` on each chart. Is there a reason why this couldn't be limited only to charts that actually have filters set? In my use cases, this typically applies only to a few charts. The following version works much quicker for me:

```
    function myFilterAll() {
        var charts = dc.chartRegistry.list();
        for (var i = 0; i < charts.length; ++i) {
            if (charts[i].hasFilter()) {
                charts[i].filterAll();
            }
        }
    }
```

I'm on `dc.js` version 2.1.4 (but the code for `dc.filterAll()` seems to be identical in the current 2.1.8).

ehansis:
Fixes #1322

### DIFF
--- a/spec/core-spec.js
+++ b/spec/core-spec.js
@@ -60,7 +60,7 @@ describe('dc.core', function () {
             chart = dc.pieChart('#id')
                     .dimension(valueDimension)
                     .group(valueGroup)
-		    .filter(0);
+                    .filter(0);
             spyOn(chart, 'filterAll');
             return chart;
         });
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
This looks good to me, but why is it necessary to call `.filter(0)` on all of those charts in the tests?

ehansis:
In `charts w/ grouping`, the tests check that `chart.filterAll()` is called or not called on a chart depending on its chart group. Since the modified `dc.filterAll()` only calls `chart.filterAll()` on charts that have a filter set, we need to set a filter on each chart to do a meaningful test.

PR #1335, opened 2017-09-06
Row chart and Pie chart incorrectly called valueAccessor instead of cappedValueAccessor.

alexcampana:
Row chart and Pie chart incorrectly called valueAccessor instead of cappedValueAccessor. This caused issues when using a custom valueAccessor and a cap.

### DIFF
--- a/spec/pie-chart-spec.js
+++ b/spec/pie-chart-spec.js
@@ -45,7 +45,7 @@ describe('dc.pieChart', function () {
             },
             //init
             function () {
-                return {count: 0, total: 0};
+                return {count: 0, total: 0, getTotal: function () { return this.total; }};
             }
         );
     });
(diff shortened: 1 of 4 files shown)


gordonwoodhull:
Thanks for the contribution @alexcampana! I really appreciate the extra tests too.

This looks correct to me. However, could you describe more in-depth the problem you were seeing? Was it crashing when trying to apply your custom `valueAccessor` to the others group?

I'm surprised no one got bit by this before, and I'll have to go back to the commits where `cappedValueAccessor` was introduced, and see if it wasn't applied everywhere, or if later commits brought back the vanilla `valueAccessor`, or what.

Any additional info you can provide would be helpful.

alexcampana:
Yes, the chart was crashing in my case.

I use reductio alias mapping, so all of my custom valueAccessors reference those aliased functions. Since the othersGroup value doesn't contain any of my aliased functions and only the final calculated value, the valueAccessor blows up when it tries to call the function.

Let me know if you need anything else.

gordonwoodhull:
Okay, thanks a lot. Like I said, I'll want to look into the root cause of the issue and the history, and then I'll merge this into the 2.0.* branch.

gordonwoodhull:
Thanks @alexcampana! Sorry for the delay; I am finally merging this for dc.js 3.0.7.

I looked into why the code was this way, and it looks like
* in the pie chart I added the "no data" check without thinking about which accessor to use
* the row chart was updated to `cappedValueAccessor` in bb20d4581 but this one line was missed

PR #1338, opened 2017-09-13
fixing IE 11 crash on multiple selectMenu

vparpoil:
overriding redraw by render function + applying filters manually

### DIFF
--- a/src/select-menu.js
+++ b/src/select-menu.js
@@ -56,7 +56,13 @@ dc.selectMenu = function (parent, chartGroup) {
         return _chart;
     };
     // Fixing IE 11 crash when redrawing the chart
-    _chart.redraw = _chart.render;
+    // see here for list of IE user Agents :
+    // http://www.useragentstring.com/pages/useragentstring.php?name=Internet+Explorer
+    var ua = window.navigator.userAgent;
+    // test for IE 11 but not a lower version (which contains MSIE in UA)
+    if (ua.indexOf('Trident/') > 0 && ua.indexOf('MSIE') === -1) {
+        _chart.redraw = _chart.render;
+    }
 
     _chart._doRedraw = function () {
         setAttributes();
--- a/src/select-menu.js
+++ b/src/select-menu.js
@@ -55,7 +55,24 @@ dc.selectMenu = function (parent, chartGroup) {
         _chart._doRedraw();
         return _chart;
     };
-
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
It's weird that this is necessary, because render calls redraw under the hood. Could you try

```js
_chart._doRedraw = _chart._doRender;
```

instead? I'd hesitate to merge something that duplicates so much code, and generally we don't implement `render` and `redraw` in the concrete charts, just in the bases.

However, I'm really glad we have this workaround for anyone who is struggling with the IE bug.

vparpoil:
@gordonwoodhull Thanks for your remarks, I updated the code for a cleaner fix. Do you think this can be merged ?

gordonwoodhull:
That looks really good, thank you!

The only thing I'd still like to see is a check for IE on the redraw = render line. Because this will be slightly less efficient. 

Unfortunately (or fortunately ;) I don't think we have any browser checks in dc.js currently, so I don't know the best technique.

In another project we used some checks based on the comprehensive detection documented here:

https://codepen.io/gapcode/pen/vEJNZN

Has anyone tested this on IE<11? I'd be inclined to just detect Trident in the user agent and call it a day.

vparpoil:
Thanks for your feedback
IE 10 and IE 9 don't have the crash issue. Lower versions don't display anything on the dc.js example for select.
I tested the proposed fix on IE 10 and 9, it behaves as expected (but the fix is not needed since these version doesn't have the issue)

The codepen you gave detects IE 11 and IE 10. Lower versions are not supported by codepen but the code in it is still valid (see [here](http://www.useragentstring.com/pages/useragentstring.php?name=Internet+Explorer) for a list of IE user agents) 

I updated the PR to add the test on Trident, I hope this fits your expectations :)

gordonwoodhull:
This looks excellent, thank you for iterating on this! I'll just confirm if automated testing on Win8 helps any, and then merge this for 2.0.*

gordonwoodhull:
Apologies for the delay. I had hoped to repro the problem in automated testing but I think this would require much more extensive tests than we currently have (like duplicating the example into a test). This is doable but I don't have the time for it, so I've simply cut the release 2.1.9.

Thanks @vparpoil!

rrameshkumar76:
Has 2.1.9 been published to npm? I could not find it in https://www.npmjs.com/package/dc

PR #1342, opened 2017-09-22
create computedMargins and use it instead of internal margins modific…

emiljas:
…ations


gordonwoodhull:
This is definitely a step in the right direction.

Thanks @emiljas!

kum-deepak:
This did not get merged for so long. It is being targeted for `dc-v5`. In similar lines, using a similar approach, `width` and `height` and no longer modified by `dc`. In `dc-v5` all chart configuration is getting moved to `conf`, so, the margins will move to `conf` and will not get modified by the charts,

PR #1348, opened 2017-10-06
Checkbox/radio button filtering

ialarmedalien:
I'm a fan of checkbox and radio button-styled filters, so I've added `dc.cboxMenu`, which functions similarly to `dc.selectMenu` but uses either radio buttons (for single selection) or checkboxes (multiple selection) instead of the `select` element.

PR #1337, opened 2017-09-12
Added dc.sunburstChart (fixes #781)

tehsenaus:
This is an update of @blairn's original PR at: https://github.com/dc-js/dc.js/pull/907

I've merged with ```develop```, made a few fixes and performance improvements, fixed the unit tests, and code style.

Documentation is missing though, hopefully someone can give me guidance on that.

PR #1329, opened 2017-08-06
Add Html legend support

rrameshkumar76:
Adds a basic HTML legend supporting both vertical and horizontal legends as discussed in  [issue](https://github.com/dc-js/dc.js/issues/1325)  and based on previous [PR](https://github.com/dc-js/dc.js/pull/577) by @davejlong

PR #1363, opened 2018-02-18
PoC - d3v4 support

kum-deepak:
Reference: 
https://github.com/dc-js/dc.js/issues/1173


gordonwoodhull:
Thanks @kum-deepak! This sounds like a good approach. 

The selection behavior has also changed a lot, since d3v3 used to automatically merge the enter selection into the modify selection. I guess maybe your compatibility layer can handle this, but when eventually removing the compatibility layer, it is still going to be tricky.

Please continue to update with your progress and I'll be happy to review or help out as you go.

kum-deepak:
Thanks for your inputs. I have been meditating for past two days on best way to handle changes to selection behavior.

.enter() is currently used in 36 places. Writing a wrapper is possible but quite clumsy, the effort to test the wrapper is going to be more than direct changes (which anyway will need to be done).

I understand that it is not pragmatic to maintain support for both versions in the same branch. At the same time due to nature of changes it is likely that support for D3v3 and D3v4 will need to co-exist for significant time.

So, I was thinking - as far as possible - to modify in the code in a way that modified code works with D3v3 as well. So far there are following known changes 

- accessing nodes from a selection
- behavior of selectAll in regard to .enter()

I think both of these can be created a functions that can be implemented for D3v3 and D3v4.

I will commit again when I reach someplace.

gordonwoodhull:
I think it's perfectly acceptable to simply move to d3v4. This was how I planned to make the port when I find time.

It's true, there will be some overlap, and fixes will have to be applied separately to two branches for a while. But once we get d3v4 fully working, it should be fine to drop d3v3, since the rest of the world has already moved on.

My two cents. Thank you for taking this challenge, and feel free to pursue it however you think is best.

kum-deepak:
Thanks! I will proceed with making direct changes. That will actually make the changes simpler :)

I was going through the code - the more complex changes brush() and dispatch() are used only once each.

kum-deepak:
Your message was really timely :)

gordonwoodhull:
Great! I think the scatter plot modifies the brush a little bit too, but otherwise yes that should be it.

kum-deepak:
Have been able to get more chart types to work. One setback - the stack API has changed beyond recognition. Spent some time on that but was not able to move it. Good thing, it is a relatively isolated code in D3v3, extracted relevant part and placed in the d3-compat file. That has allowed me to go ahead. I will come back to it when other issues are under control.

Next target for me is brushes.

gordonwoodhull:
Awesome! Yes, having read CHANGES.md, it looks like every interface has changed somewhat, and many in fundamental ways. 

So grateful you are taking the time to work on this.

kum-deepak:
Brushing itself works :) New brush APIs are simpler and is likely to yield a simpler code.

I am facing a very stupid doubt, brush needs to be told the region brushing is allowed. For this I need the width and height of the chart area. How do I get that?

gordonwoodhull:
That's great!

You can use `.effectiveWidth()` and `.effectiveHeight()` to get the chart extent. They are defined in the `marginMixin`.

kum-deepak:
Started seeing end of the tunnel. Many examples have started working. Major dev work:

- Zoom
- Brush - need to implement 2D brushing for scatter plots, refactor general brush
- Area - lines are working, area is not

Examples that are not working as of now:

- http://localhost:8888/web/examples/area.html (lines ok, area missing)
- http://localhost:8888/web/examples/box-plot-time.html
- http://localhost:8888/web/examples/line.html (lines ok, area missing)
- http://localhost:8888/web/examples/multi-focus.html
- http://localhost:8888/web/examples/range-series.html (brush/zoom not working)
- http://localhost:8888/web/examples/scatter-brushing.html (brush not working)
- http://localhost:8888/web/examples/series.html
- http://localhost:8888/web/examples/splom.html (brush not working)
- http://localhost:8888/web/examples/time-intervals.html 
  (bar width, some warnings, brushing not working)
- http://localhost:8888/web/examples/time-intervals.html (bar width incorrect)

gordonwoodhull:
Wow! You're really getting there!

Amazing work.

kum-deepak:
Zoom and focus works now. :)

kum-deepak:
Today reached an important milestone - brushing now works for all cases. This makes example on the homepage completely usable (only one visual difference - line chart does not draw the area).

List of known issues have reduced to quite low number.

gordonwoodhull:
Wow, you are really getting there! Assuming the rest goes as well, the changes can form the core of dc.js 3.0.

It will be at least a week, probably longer, until I'm able to review, as I have a broken collarbone and can't spend much time upright at the computer.

So when you finish up, please try to document anything you had to do that was non - obvious.

Thanks again for the substantial effort that went into this.

kum-deepak:
Great news to share - reached functional completeness today. Floating on clouds now :)

I have checked all samples (homepage, vc, crime, all examples). Visually they have reached close.

I will move to fix test cases - as of now around 10% are failing.

I am expecting subtle errors due to change in API behavior. Interesting cases might arise because of changes in SelectAll(...).enter() behavior. We will get more cases when it is opened to wider user base.

At this point, I will be grateful if you could glance over the code changes - let me know if you find something wrong/improper.

I will do following before it can be opened to wider user base as a beta:

- Update documentation
- Create a migration guide

gordonwoodhull:
As noted in my last comment, I'm physically unable to review right now. Sorry about this! But I'm very excited about your work.

kum-deepak:
Oh, really sorry to hear that (did not read your previous comment - this window was open for long when I posted mine)! Get well soon!

After having done it, I realize that most of the changes that D3 team has done are actually logical. If anything - it will allow the code to be simpler and more predictable.

In this version I have been trying to keep the code diff to be as small as I could. This is helping me in matching the behavior easily.

This year I have allocated my significant time on visualization - I am hoping to spend most of it in this library. Please let me know if you have any items in your wish list. :)

NickQiZhu:
Big ++1!

gordonwoodhull:
Thanks @kum-deepak! There is definitely lots to  work on.

My priority is getting pull requests merged. But they often need a lot of work to get there - new tests or broken tests, edge cases not thought of, or the feature breaks old functionality which wasn't tested. 

But there are a couple which I believe are ready to go, and now just need to be updated to d3v4. Assume that this branch will become 3.0 and create sub branches for any you work on.

#1337 sunburst
#1348 checkbox menu
#1329 html legend 
#936 input filter 

Or any others you think can be improved.

Or, if you want to take this in a systems-y library maintenance direction, I want to follow the lead of d3 and break up dc.js into a lot of smaller libraries, using rollup, so that it's easier to take just what you need, and easier to write extensions. E.g. we'd have a dc-barChart library and a dc-compositeMixin library.

But this is difficult and I haven't thought everything through.

kum-deepak:
Many thanks! Will look onto all the 4 pull request. I think these can be merged for 3.0. 

I made a quick glance to the code, some of these do need work before it can be included. In some cases we may need to copy and paste needed code :)

After that I definitely think splitting this library into smaller files will be great idea. We should consider ES6 or TypeScript at that point. A lot of code will need to be refactored. It initially looks like a lot of work but my guess is that current code structure should lend easily to splitting.

#1175 has attempted to switch to ES6 however more as replica of the current code. Anyway once D3v4 work nears completion I will create an issue for this. :)

gordonwoodhull:
Unfortunately I am not all that excited about switching to ES6 or Typescript. It can cause a lot of trouble with debugging, since source maps are not perfect and there are multiple browsers to consider, not just Chrome.

Sorry about that. I want to follow precisely the lead of d3 and stick with ES5 plus modules.

kum-deepak:
Finally all test cases are passing. All examples works.

I will merge in D3 compat files now.

kum-deepak:
Merged d3-compat into main code base. So other than 2 functions `functor` and `d3.layout.stack` all compatibility layer is removed.

Next focus on the following:

- Update documentation
- Create a migration guide

gordonwoodhull:
What a thorough job you are doing. This is a joy to behold.

I'm still not able to do much testing, but I understand the major remaining difficulties are

1. Stack layout works differently so we'll have to rewrite. I'm sure the old functionality is still supported. This may be an opportunity to support more features because I think negative stacks and streamgraphs are easier now.

2. Programmatic control over the brush, as used by range/focus and the restrict panning example. As for restricting panning, IIIRC we knew there would be forward compatibility issues with the hack we use, and the feature is supported directly by d3v4. The issue is #1026 and the commit that fixed it is  41aca80, if you want to dig into that. Hopefully it's a similar situation for range/focus.

I should be able to help with these when I return, but that still may not be for a couple weeks. Anything else?

Great work!

kum-deepak:
Thanks for your detailed reply.

1. I have copied the stack code from D3v3 into dc namespace, this allows the current code to work same as previous version. However, the D3v4 implementation seems way closer to what the Line and Area functions need. So, it will make the code simpler over all (in terms of reducing pre/post processing around the call to the stack generator).
1. I am currently looking at brush - one bug in some specific sequence of operations. I will look at the issue #1026 and the associated commit. Brush code has been cleaned quite a lot already. I will look at zoom related code as well.

I have some additional items in mind (none are priority):

1. The D3v4 selectAll leads to better code structure if we did remove, enter, and update - in this sequence. Currently most places sequence is enter, update, and remove. Might be a great time to change it. (Not mandatory, it works without issues, only need extra variables).
1. I have a suspicion that Bubble Overlay is not working fully in /crime/index.html. Will look at it. (Low priority)

I will now be preparing the Upgrade guide (I am planning to use links into the D3 changes page) and updating documentation. Once this is done we can do a baseline. At that stage all other work can be carried out as separate issues/pull requests into the baseline.

kum-deepak:
I just now realized, it is about 3 weeks now. Feeling happy with the progress so far :) All the analysis in #1173 and your suggestions have helped quite a lot.

Many thanks for your wonderful work!

kum-deepak:
Realized a strange issue in dc.event.trigger - if there are 2 different set of events firing simultaneously - one of those are likely to be lost. I think it is happening in new code when brushing in range chart is causing focus (or vice versa).

I will rewrite that part after the baseline. I am planning to create multiple queues, one for each event.

Justus-Maier:
This is amazing work! Im coming here from the D3 Cookbook v2 (Zhu), which features D3 Example libraries like NVD3, Rickshaw and DC. D3v4 migrations are hard and I won't go to v3 just to import some charts. Very much interested in your updated DC library though, could you publish as a beta?

gordonwoodhull:
Hi @Justus-Maier. There are still some outstanding issues on this PR and a couple things which need to be documented. I'd be willing to publish a 3.0 alpha but it won't be until next week.

@kum-deepak, I'm surprised by the trigger bug you're reporting. I know that trigger is really more of a debounce (#630) but I wasn't aware that separate triggers could interfere with each other...

kum-deepak:
Hi @Justus-Maier, code changes are nearly done, currently work in on for documentation updates and upgrade guides. Stay tuned, will let you know when you can test the branch.

@gordonwoodhull - I was actually surprised for it to have survived so long. When I go for implementing/fixing that I will try to move to as close to debounce as possible.

kum-deepak:
@gordonwoodhull I am making documentation changes, getting along well so far. I need your expert advice. :)

Please go through https://github.com/d3/d3-scale/blob/master/README.md - I think dc will be able to reliably support only continuous and ordinal scale types. Quite few places we are using .invert, .rangeRound etc. Some places it might be ok to use other scales like in the y axis, bubble size.

Based on your inputs, I will add a caution note.

gordonwoodhull:
Yes it will be good to document this, although it might be a bit of a rabbit hole.

I think we have only ever consistently supported linear and ordinal scales. There have been some attempts to support log scales, but they still break in some places, and many of the other continuous scales may work "by accident", but I don't think they have any tests.

The attitude had sort of been that this is an open system (based on Nick's cookbook) and if something doesn't work, there is often a workaround. It's a "leaky abstraction by design".

So I guess I'm saying that I agree with your analysis, but I wouldn't word it too strongly.

I'm still not able to spend a lot of time at the computer but I look forward to doing a thorough review once I'm a bit better!

gordonwoodhull:
And yes, as for the triggers, I agree, I don't think `dc.events.current` should be a global. Ideally I guess it should be per-event and per-chart although maybe that is overkill since triggers only happen in response to the mouse. 

Do you think this old bug is exacerbated by the changes you made in this PR? I.e. are you seeing worse bugs with d3v4 than with d3v3, or is it just bugs that have always been there?

kum-deepak:
@gordonwoodhull regrading scales, I will park it as of now with keeping a similar language as current documentation. I think we can create a table in our FAQ or a separate page with one Axis being the scale type and another usage type (x, y, elasticX, elasticY, color, radius, brush, zoom ...) - then we can fill cells with Yes, No, and Maybe. Values for most of the cells will be well known (which are commonly used) with remaining to be initially Maybe.

gordonwoodhull:
:+1: great idea! 

Yes, d3v4 doesn't really change the situation here, except that there are a bunch of new scale types.

kum-deepak:
@gordonwoodhull trigger issue was discovered accidentally. The new implementation for zoom in the focus was causing setting of filter in the range chart (reverse is also true, i.e., brushing in one chart is causing zoom in another). In this case the zoom event in the focus chart and setting of filter in the range chart were both using trigger.

It is not noticeably easily. Occasionally the the brush may stop earlier than how much I dragged. But again if I try it will move properly. So, users might be confused but probably will be able to pinpoint if there is an issue. Using a debugger I could not find out. Once I switched to age old console.log it started showing the behavior.

Yes I agree, that we should make it per event per chart. Currently it does impact day-to-day use. So, I will handle it after the fist base line.

kum-deepak:
Once I am able to create the base line, I will create few new pull requests one each for variety of stuff :)

kum-deepak:
I am not worrying too much now :) Let us tackle one issue at a time.

gordonwoodhull:
Completely agree. Looks like we're almost there - congratulations!

gordonwoodhull:
I was able to do some testing today, so I ran through all the examples.

Here are the issues I found, mostly to do with initial values for elements that are entering.

We don't need to fix these before putting out the alpha, but we'll definitely need to fix them, as well as using the d3v4 stack implementation, before releasing 3.0.0. I'll open issues for any that you don't fix before we merge this PR.

### bar charts
 * [ ] introductory transition on bars is from lower left instead of just from bottom
 * [ ] bar labels transition from upper left
 * [ ] multi focus example: bars turn grey before being filtered/focused (?) out instead of just disappearing. Not sure which behavior is better here, but we should figure out  and decided one way or the other.

### line charts
 * [ ] dots transition from upper left instead of just being there
 * [ ] range series example: brush handles not removed when clicking to remove brush. probably because composite chart?

### heat map
* [ ] squares transition from upper right instead of just fading in

kum-deepak:
Thanks for your review.

I have committed now draft version of upgrade guide. I now realize that other than changes to the user code for d3 changes, there are only two changes.

It is possible to make those two backward compatible, will wait for your feedback.

kum-deepak:
I think we have reached a stage where we can baseline it as a single commit, and split for work on localized/specific issues.

My current (non exhaustive list):

- Cleanup of zoom behavior. Testing of restrict panning etc.
- dc.event.trigger
- exit/enter/update sequence - this is an important one
- Additional test cases for dc.util.add and dc.util.subtract
- One for each new chart type - original PRs from community
- Consider removing transitions for zooming and brushing
- d3 stack to D3v4
- Changing .tension call in lineChart
- http://localhost:8888/web/examples/filter-stacks.html 
 (fading is peculiar)
- http://localhost:8888/web/zoom/restrict-panning.html
  Need to increase height of range chart - currently height of
  brushable area is zero. New implementation needs it be more than zero.
- Issues with transitions (d3 behavior might have changed, or I clobbered some functionality as I did not understand that)

I will describe each of these on the individual issue/PR.

I will keep working one/few at a time. Will love to hear your prioritization/grouping of the above.

kum-deepak:
It was my first experience with working with D3 - completing a month now. :)

I now realize that behavior of `.selectAll` in D3v4 implicitly favors code that first handles exiting elements, following by entering and updating. The current dc code does it in sequence enter, update and exit.

Making the change will make the code cleaner and more manageable. However I am not sure if will make any adverse noticeable change for the user. In any case I wanted first a working version before attempting this one.

In the current code sometimes, `.enter` is called in a function other than the function that called `.selectAll`. I was needed to return the entering collection from the function as it it is needed for the call to `.merge`. A more maintainable code will be to call `.enter` in the same function as `.selectAll` and adjust the parameters accordingly.

I think that when we make the above change - some subtle bugs would get fixed - which may be related to updating elements not behaving properly.

Will wait for your suggestion before proceeding on this one.

kum-deepak:
@gordonwoodhull can you please merge this pull request as a new branch (say d3v4) as a single commit. I will use that as base to continue further work.

Thanks!

gordonwoodhull:
Yes I'll create a 3.0 branch today. Sorry for the delay - yesterday I got bogged down thinking about whether we should be backward compatible. (It's easy enough so I think we should.)

Please start opening issues for the remaining work. I'll create 3.0 milestone now.

kum-deepak:
There may still remain some incompatibility around tension in interpolate/curve function. Which may impact very few users.

gordonwoodhull:
Sorry for the delay - I'm still not able to work much. I've created the 3.0 branch now!

BTW you're not following the jshint/jscs style guides. I guess maybe you never ran plain `grunt` with no parameters, which installs some useful commit hooks. If you don't want the hooks just run `grunt lint` to check everything. It doesn't matter to me if those don't pass every time, but you might want to start out with a single commit to fix these since it will touch a lot of files.

From now on, please select `3.0` as the base for your future PRs. I'll close this PR now, although there has not been an official release yet.

src/bubble-chart.js
```diff
@@ -123,9 +128,9 @@ dc.bubbleChart = function (parent, chartGroup) {
         // override default x axis brush from parent chart
     };
 
-    _chart.redrawBrush = function () {
+    _chart.redrawBrush = function (g, selection, doTransition) {
```
gordonwoodhull: Whoops - looks like you added some arguments here but never pass them anywhere. I just ran into this because I wanted to add a `dontSetZoom` argument. Do you remember what you were trying to do here?

gordonwoodhull: Specifically, it looks like you changed the signature to take the current filter/selection, but this also takes `g` and `doTransition` - so I guess it's a simple fix since those are unused.

src/coordinate-grid-mixin.js
```diff
@@ -1034,28 +1087,34 @@ dc.coordinateGridMixin = function (_chart) {
         }
     };
 
-    _chart.redrawBrush = function (g, doTransition) {
-        if (_brushOn) {
-            if (_chart.filter() && _chart.brush().empty()) {
-                _chart.brush().extent(_chart.filter());
-            }
+    _chart.redrawBrush = function (selection) {
```
gordonwoodhull: Why did you remove `doTransition` here? [_brush.move](https://github.com/d3/d3-brush/blob/master/README.md#brush_move) can take a transition or a selection, just like the old brush could be applied to either a selection or transition.

It's not a big deal, just a tiny feature that makes chart resizing work better, but it makes me worry when functionality is removed with no explanation. In the future, could you please try to document the things that you disable or remove? 

I guess I need to go over this PR with a fine-toothed comb before we release 3.0. We don't have full test coverage, especially where transitions are concerned - it's unclear how to test them.

kum-deepak: Thanks for letting me know, I will mark in future if I remove or change any functionality.

This one, in my initial implementation, was aggravating the infinite recursion. However the current code, may not suffer from that.

I will be more comfortable if you could go through the entire change set considering it to be my first encounter with both dc and d3.

gordonwoodhull: I see, you probably simplified it in order to figure out why it was recursive, and then didn't restore the feature when you were done.

No harm done. Yes, I'll read over all the changes before we release.

Justus-Maier:
@gordonwoodhull how far are you through code review? Would you publish an alpha asap?

kum-deepak:
@Justus-Maier there are lot of changes, so it will take a while. Meanwhile you can install from 3.0 branch. Code can be considered alpha.

gordonwoodhull:
Yes, as @kum-deepak says, review is taking a while. Although I've started giving them alpha numbers (we're at 2 going on 3) I am not sure when I will publish to npm & cdnjs. Hopefully later this week.

(The trouble is that a lot of people will take the latest version even if it says alpha, and then complain about it.)

Meanwhile, if you're using npm, you can change your dependency in package.json to:

```js
    "dc": "git://github.com/dc-js/dc.js#3.0",
```

to install directly from the github branch.

src/composite-chart.js
```diff
@@ -508,15 +520,21 @@ dc.compositeChart = function (parent, chartGroup) {
...
-     * object](https://github.com/d3/d3-3.x-api-reference/blob/master/SVG-Axes.md#axis) therefore it supports any valid
+     * object](https://github.com/d3/d3-axis/blob/master/README.md) therefore it supports any valid
      * d3 axis manipulation.
      *
      * **Caution**: The y axis is usually generated internally by dc; resetting it may cause
-     * unexpected results.
+     * unexpected results.  Note also that when used as a getter, this function is not chainable: it
+     * returns the axis, not the chart,
+     * {@link https://github.com/dc-js/dc.js/wiki/FAQ#why-does-everything-break-after-a-call-to-xaxis-or-yaxis
+     * so attempting to call chart functions after calling `.yAxis()` will fail}.
```
gordonwoodhull: Nice.

PR #1365, opened 2018-02-25
Composite Pass-Through Properties + Rescale

### LINKED ISSUE #554: .title not working on composite of line charts

### LINKED ISSUE #611: Composite Chart Not Rescaling Bar Width
I have been using the following to rescale bar widths when resizing a full-width chart, where width is assigned by the anchor set at 100% in CSS:

```
chart.on('preRender', function(){
    chart.rescale();
});

$("#some-chart").on("resize", function(){
    chart.render();
});
```

Which works perfectly if `chart` is a bar chart. However, if the bar chart is part of a composition, where `chart` is the composite chart, bar widths no longer rescale.

It seems that `_chart.xAxisLength()` in `calculateBarWidth` is not reflecting the new width.

### LINKED ISSUE #1003: need to call margins before compose for child charts to inherit
I have simple example of using composite chart with currently latest dc js version (18). Anyway if I try to add any margin to chart definition I get chart lines shifted.

``` js
myChart
    .width(500)
    .height(200)
    //.margins({top: 10, right: 50, bottom: 30, left: 40})
    .dimension(timeDimension)
    .compose([
        dc.lineChart(myChart)
            .group(group1, "G1"),
        dc.lineChart(myChart)
            .group(group2, "G2")
            .useRightYAxis(true)
            .ordinalColors(["#ff7f0e"])
    ])
    .x(d3.time.scale().domain([minDate, maxDate]))
    .xUnits(d3.time.weeks)
    .elasticX(true)
    .elasticY(true)
    //.margins({top: 10, right: 50, bottom: 30, left: 40})
    .legend(dc.legend().x(60).y(10).itemHeight(13).gap(5))
    .on("renderlet", rotateTimeAxisLabels);
```

**UPDATE** Depends on position of `.margins(...)` call. If it is at the beginning then everything is ok. If it is at the end of config - I get a bug with shifting.

You can play a little bit with method call position, it depends.

### LINKED ISSUE #1314: 2.0.x: label() not working for composite charts.
The `label()` function doesn't seem to do anything for composite charts, even with `renderLabels(true)` explicitly set.

Apologies for the lack of code example; this is a quick drive-by since I'm on the busy side, but if it helps, `label()` works as intended for me for the individual bar/line charts, but when I composite them, it doesn't work even if I set up label/renderLabels the exact same way on the composite itself. Suffice to say, it's either unimplemented or there's a significant API difference I can't figure out.

dahlbyk:
As proposed in https://github.com/dc-js/dc.js/issues/611#issuecomment-367912017:

* [x] `height` and `width` updates should cascade to children (fix #611)
* [x] `margins` updates should cascade to children (fix #1003)
* [x] `title` updates should cascade to children if/when `shareTitle(true)` (fix #554)
* [x] `rescale()` should cascade to children (no specific issue that I can find, but it seems important?)
   * There might be a larger bug here - my test is failing because `resizing()` on the children never goes to `false`
* [ ] <del>`label` and `renderLabel` updates should cascade to children</del> (defer fix for #1314)

<del>Still in progress, but opening PR to check CI and get feedback.</del> cc @gordonwoodhull

### DIFF
--- a/spec/composite-chart-spec.js
+++ b/spec/composite-chart-spec.js
@@ -873,4 +873,45 @@ describe('dc.compositeChart', function () {
             });
         });
     });
+
+    describe('title and shareTitle', function () {
+        function f (value) { return function () { return value; } }
+
+        it('should propagate existing title', function () {
+            chart.children().forEach(function (child, i) {
+                child.title(f('Title ' + i));
+            });
+
+            chart.shareTitle(false);
+            chart.title(f('Share Me'));
+
+            chart.children().forEach(function (child, i) {
+                console.log(child.title());
+                expect(child.title()()).toBe('Title ' + i);
+            });
+
+            chart.shareTitle(true);
+
+            expect(chart.title()()).toBe('Share Me');
+
+            chart.children().forEach(function (child) {
+                expect(child.title()()).toBe('Share Me');
+            });
+        });
+
+        it('should propagate new title', function () {
+            chart.shareTitle(true);
+            chart.children().forEach(function (child, i) {
+                child.title(f('Title ' + i));
+            });
+
+            chart.title(f('Share Me'));
+
+            expect(chart.title()()).toBe('Share Me');
+
+            chart.children().forEach(function (child) {
+                expect(child.title()()).toBe('Share Me');
+            });
+        });
+    });
 });
--- a/spec/composite-chart-spec.js
+++ b/spec/composite-chart-spec.js
@@ -828,4 +828,49 @@ describe('dc.compositeChart', function () {
             });
         });
     });
+
+    describe('rescale', function () {
+        beforeEach(function () {
+            expect(chart.resizing()).toBe(true);
+            chart.children().forEach(function (child) {
+                expect(child.resizing()).toBe(true);
+            });
+
+            chart.render();
+
+            expect(chart.resizing()).toBe(false);
+            chart.children().forEach(function (child) {
+                expect(child.resizing()).toBe(false);
+            });
+
+            chart.rescale();
+        });
+
+        it('should rescale child charts', function () {
+            expect(chart.resizing()).toBe(true);
+
+            chart.children().forEach(function (child) {
+                expect(child.resizing()).toBe(true);
+            });
+        });
+    });
+
+    describe('re-compose rendered chart', function() {
+        beforeEach(function () {
+            chart.render();
+            expect(chart.resizing()).toBe(false);
+
+            chart.compose([
+                dc.lineChart(chart).group(dateGroup)
+            ]);
+        });
+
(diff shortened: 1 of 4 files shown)


dahlbyk:
> There might be a larger bug here - my test is failing because `resizing()` on the children never goes to `false`

To fix this I've made `coordinateGridMixin.resizing()` a setter, too, so it can be overridden in `compositeChart`. Technically this is a change to the public API, but it's not currently documented.

dahlbyk:
> `label` and `renderLabel` updates should cascade to children

I don't have a specific need here, so I'm going to leave this change out of this PR. If someone else needs #1314 fixed, you'd likely follow a pattern similar to what's found here.

@gordonwoodhull any feedback?

kum-deepak:
Many thanks for your work @dahlbyk, currently heavy work is going on branch 3.0. Will it be possible for you to rebase and target your PR onto 3.0?

dahlbyk:
> Will it be possible for you to rebase and target your PR onto 3.0?

Done. We'll see if tests still pass. 😀

I'll leave the 2.x version at https://github.com/dc-js/dc.js/compare/da3c437f1161380c4d2357e9b7116f824ea7ab0c...NewBoCo:composite-props-2.x in case anyone needs it.

dahlbyk:
Revisiting this (a year later!) as I'm finally upgrading from my patched v2 to v3. I've refactored a bit and included a fix for `title`/`shareTitle` as well.

gordonwoodhull:
Apologies for the long delay on this. I was reminded of this PR by [a question on SO](https://stackoverflow.com/questions/58215086/margins-on-composite-chart).

Nice, clean implementation and special thanks for including tests!

Released in 3.1.5 and will port to the es6 branch shortly.

Thanks @dahlbyk!!

dahlbyk:
Rebased v3.1.5 commits here: https://github.com/dc-js/dc.js/compare/1438f80...cdc1cda

Issue #949, opened 2015-06-11
Line chart doesn't render between points if one falls outside the x-range/focus


onorinbejasus:
I couldn't find a previous issue that detailed this:

If a point lies beyond the X-axis' current focus/domain, the line between the points will not render. This seems to have been a problem since 2.0.0 alpha 1. 

Here is the expected behavior using 1.7.3:
https://jsfiddle.net/7qre8p51/2/
![1 7 3](https://cloud.githubusercontent.com/assets/452256/8113395/cf3f8480-103b-11e5-90fa-21b95d7c27d1.png)

And the current behavior with 2.0.0b12

https://jsfiddle.net/7qre8p51/8/

![2 0 0b12](https://cloud.githubusercontent.com/assets/452256/8113404/e12b5fd4-103b-11e5-9c9a-1dcbc64df889.png)

onorinbejasus:
The problem stems from the function that domainFilter() uses to compare values to be plotted. As the it stands currently, the comparator is determined by the chart's domain: _chart.x().domain. Thus, data outside of the chart's range will never be rendered -- or drawn to in this case.

The (temporary) solution:

Extend xDomain in domainFilter() to include the previous and next values if they exist:

                var xDomain = _chart.x().domain(),
                    prev = _chart.group().top(Infinity).filter(function(d) {return d.key < xDomain[0] } ),
                    prev = (prev.length > 0) ? prev[0] : xDomain[0],
                    next = _chart.group().top(Infinity).filter(function(d) {return d.key > xDomain[1] } ),
                    next = (next.length > 0) ? next[next.length-1] : xDomain[1];

                // set the new XDomain
                xDomain = [prev.key, next.key];

gordonwoodhull:
Thanks @onorinbejasus, that sounds about right.

onorinbejasus:
I actually like this solution more, but both work. The benefit of this one is that we don't have to worry about filters that may exist on the crossfilter object:

```javascript
        dc.stackMixin = function (_chart) {
            function prepareValues (layer, layerIdx) {
                var valAccessor = layer.accessor || _chart.valueAccessor();
                layer.name = String(layer.name || layerIdx);
                layer.values = layer.group.all().map(function (d, i) {
                    return {
                        x: _chart.keyAccessor()(d, i),
                        y: layer.hidden ? null : valAccessor(d, i),
                        data: d,
                        layer: layer.name,
                        hidden: layer.hidden
                    };
                });

[...]

PR #1367, opened 2018-02-28
Fix elasticY with evadeDomainFilter

dahlbyk:
Fixes that `evadeDomainFilter`, introduced by 77d22ba221f697e40cf3a63fb26a1098422ba0aa for #949, doesn't work with `elasticY(true)`. The fix requires access to the domain-filtered values separate from all values needed to render the chart.

Bonus fix: `x().domain()` transitions from `[0, _]` to `[x, _]`, as in the [bar transitions demo](http://dc-js.github.io/dc.js/transitions/bar-transitions.html) with "right" selected, will now `rescale()` correctly.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto
--- a/src/coordinate-grid-mixin.js
+++ b/src/coordinate-grid-mixin.js
@@ -458,7 +458,7 @@ dc.coordinateGridMixin = function (_chart) {
 
     function compareDomains (d1, d2) {
         return !d1 || !d2 || d1.length !== d2.length ||
-            d1.some(function (elem, i) { return (elem && d2[i]) ? elem.toString() !== d2[i].toString() : elem === d2[i]; });
+            d1.some(function (elem, i) { return (elem && d2[i]) ? elem.toString() !== d2[i].toString() : elem !== d2[i]; });
     }
 
     function prepareXAxis (g, render) {
--- a/src/stack-mixin.js
+++ b/src/stack-mixin.js
@@ -11,7 +11,7 @@ dc.stackMixin = function (_chart) {
     function prepareValues (layer, layerIdx) {
         var valAccessor = layer.accessor || _chart.valueAccessor();
         layer.name = String(layer.name || layerIdx);
-        layer.values = layer.group.all().map(function (d, i) {
+        var allValues = layer.group.all().map(function (d, i) {
             return {
                 x: _chart.keyAccessor()(d, i),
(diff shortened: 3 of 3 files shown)


gordonwoodhull:
Thank you for all these excellent contributions, and sorry for my slow response.

I was in a car accident last week and broke my collarbone, so I haven't been able to spend much time at the computer.

I am very excited to review your three PRs when I'm able, but it won't be for at least another week.

dahlbyk:
> I was in a car accident last week and broke my collarbone, so I haven't been able to spend much time at the computer.

Oh my, so sorry to hear that. Hope you have a quick recovery!

dahlbyk:
Rebased onto `3.0`.

I'll leave the 2.x version at https://github.com/dc-js/dc.js/compare/da3c437f1161380c4d2357e9b7116f824ea7ab0c...NewBoCo:elasticY-evadeDomainFilter-2.x in case anyone needs it.

gordonwoodhull:
Ok, I was about to look at `d3.stack` #1375 so I figured I should merge this first.

Thanks @dahlbyk, this makes a lot of sense. I added a simple test, too - turns out we already covered this situation but didn't test it.

Let's fix those line endings too.

I'm dropping the fix to `compareDomains` because @kum-deepak has a more comprehensive fix in the pipeline in #1402.

This will go into 3.0 alpha 7.

dahlbyk:
>https://github.com/dc-js/dc.js/issues/1391#issuecomment-379763031: I found it a bit confusing but maybe I didn't stare at it long enough.

To clarify what I'm trying to fix, with `elasticY(true)` added to the bar-transition demo:

### With `evadeDomainFilter(false)`
Note bars outside the filtered domain transition up/down.
elasticy-evadedomainfilter-false

### With `evadeDomainFilter(true)`, broken
Bars outside the domain filter now slide in/out (the point of `evadeDomainFilter`), but the y-axis is not elastic because `flattenStack` always looks at values for the full domain.
elasticy-broken

### With `evadeDomainFilter(true)`, fixed
Now y-axis is elastic because `flattenStack` considers values for the current domain (saved in `layer.domainValues`) while `layer.values` keeps all values to evades the domain filter.
elasticy-fixed

gordonwoodhull:
Thanks for the demo! That's very helpful for documenting the change going forward. 

Yesterday when I looked at this PR again, I understood it, and this fix is already on the 3.0 branch along with a test (linked above).

I haven't tagged it yet. Hoping to include it in alpha 7 with part of #1402 today.

PR #1372, opened 2018-03-22
jslint and jshint clean

kum-deepak:
I have excluded `src/d3v3-compat.js` from the checks - did not want to change external code. Also handled little differently unused functions warning in ` spec/coordinate-grid-chart-spec.js`

Thanks for your suggestion - now lint in run before commit.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -75,7 +75,7 @@ module.exports = function (grunt) {
             source: {
                 src: [
                     '<%= conf.src %>/**/*.js',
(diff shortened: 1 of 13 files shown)


gordonwoodhull:
Great! I also pushed artifacts so now there is a built dc.js 3.0.0-alpha.1 and d3.js 4.13 in `web/js`

(technically we don't need these until release, but it's nice to have a baseline and people can try it out if they want.)

PR #1381, opened 2018-03-24
Adding support for dc.lineChart.curve. Deprecating interpolate/tension.

### LINKED ISSUE #1376: Upgrade lineChart.interpolate -> lineChart.curve in a backward-compatible way
Since [d3 has replaced `d3.svg.line.interpolate` with `d3.line.curve`](https://github.com/d3/d3/blob/master/CHANGES.md#shapes-d3-shape), we should too. But we have the opportunity to do it in a mostly backward-compatible way.

First, we can create the new `lineChart.curve()` which takes a curve object, but leave `lineChart.interpolate()`, now taking a string and instantiating the right curve object based on the old names of interpolations. We should deprecate `interpolate` using [dc.logger.deprecate](https://github.com/dc-js/dc.js/blob/ad5be91e0ffc034e29e4ce32f45eb8872ec404bb/src/logger.js#L29), and remove it in 3.1 or later.

@kum-deepak, I think we can leave `tension` the way you have implemented it in [_interpolateWithTension](https://github.com/dc-js/dc.js/blob/acff05b9b72ab6a7c7d85893d698d990c2007d16/src/line-chart.js#L205), but instead of silently dropping tension when it's not applicable, we should warn the user in the browser console, "Tension specified for a curve that doesn't support it, ignored"

kum-deepak:
Resolves #1376 

Updated documentation and code examples. Looking forward to your feedback.

### DIFF
--- a/src/line-chart.js
+++ b/src/line-chart.js
@@ -275,8 +275,12 @@ dc.lineChart = function (parent, chartGroup) {
             curve = d3.curveLinear;
         }
 
-        if (_tension && typeof curve.tension === 'function') {
-            curve = curve.tension(_tension);
+        if (_tension !== null) {
+            if (typeof curve.tension !== 'function') {
+                dc.logger.warn('tension was specified but the curve/interpolate does not support it.');
+            } else {
+                curve = curve.tension(_tension);
+            }
         }
         return curve;
     };
--- a/src/line-chart.js
+++ b/src/line-chart.js
@@ -39,8 +39,9 @@ dc.lineChart = function (parent, chartGroup) {
     var _dataPointRadius = null;
     var _dataPointFillOpacity = DEFAULT_DOT_OPACITY;
     var _dataPointStrokeOpacity = DEFAULT_DOT_OPACITY;
-    var _interpolate = d3.curveLinear;
-    var _tension = 0;
+    var _curve = null;
+    var _interpolate = null; // d3.curveLinear;  // deprecated in 3.0
+    var _tension = null;  // deprecated in 3.0
     var _defined;
     var _dashStyle;
     var _xyTipsOn = true;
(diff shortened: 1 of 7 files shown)


src/line-chart.js
```diff
@@ -201,10 +236,49 @@ dc.lineChart = function (parent, chartGroup) {
...
+            curve = mapping[_interpolate];
+        }
+
+        // Default value
+        if (!curve) {
+            curve = d3.curveLinear;
+        }
+
+        if (_tension && typeof curve.tension === 'function') {
+            curve = curve.tension(_tension);
```
gordonwoodhull: Please add a warning here if `_tension` was specified but the curve doesn't support it.

gordonwoodhull:
This is really solid, careful work. Just one more backward-compatibility warning would help, I think.

Thank you for all your efforts.

kum-deepak:
Thanks @gordonwoodhull, will add the warning soon. Currently deep into the zoom behavior.

src/line-chart.js
kum-deepak: While fixing also realized that 0 is a valid value for _tension, which would have gotten discarded.

gordonwoodhull:
Thanks @kum-deepak! Made some tweaks and merged.

kum-deepak:
Thanks @gordonwoodhull, all your changes make sense!

Any suggestions what I should pick up next?

gordonwoodhull:
Thanks @kum-deepak, please review your comments and Upgrade Notes from #1363 and see if there is anything else you want to address. 

After that, please start looking at the 4 PRs I listed there, which I believe are ready to go, but need to be ported to d3v4. I have not thoroughly reviewed them, so if you see anything that needs improvement, go for it. I think they are ready because they have passing automated tests and most of them have also been tested by people beside the author.

I will look at the transition issues and `d3.stack`. Apart from an issue or two in the examples, I think these are the only blockers for releasing a beta.

Congratulations, and thank you!

Issue #1379, opened 2018-03-23
Choropleth keep d3.geoAlbers the default projection but deprecate it


gordonwoodhull:
In the future we want geoChoroplethChart to default to the null (identity) projection since it is built on `d3.geoPath` which so defaults.

However we have the opportunity to get there in a backward-compatible way, with a deprecation warning.

Here's one way to get there:

* add a member `_projection`, which defaults to undefined
* at the start of `doRender` and `doRedraw` (maybe in a helper function)
```js
if(_projection === undefined) { 
    console.warn('choropleth projection default of geoAlbers is deprecated');
    _geoPath.projection(d3.geoAlbers());
} else {
     _geoPath.projection(_projection);
}
```
* `geoChoroplethChart.projection` becomes a regular getter/setter of `_projection` (note getter is currently broken)

gordonwoodhull:
In the future, we just default `_projection` to null and get rid of the `if(_projection` stuff

kum-deepak:
Good idea!

kum-deepak:
Going after this one next :)

From https://github.com/d3/d3/blob/master/CHANGES.md

> The default projection for d3.geoPath is now null rather than d3.geoAlbersUsa; a null projection is used with pre-projected geometry and is typically faster to render.

Also going through https://bl.ocks.org/mbostock/5557726

It seems that `null` is a valid value for `projection` now. So, will program accordingly.

gordonwoodhull:
Great! Yes let's plan for null being the default in the future. If people want it now they can set it explicitly. But undefined for now means "no value set", which we'll default to Albers.

PR #1382, opened 2018-03-25
Choropleth keep d3.geoAlbers the default projection but deprecate it

kum-deepak:
Reference #1379

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: node_js
 node_js:
   - '8'
 script:
-  - 'if [ "${TRAVIS_PULL_REQUEST}" = "false" ]; then grunt ci; else grunt ci; fi'
+  - 'if [ "${TRAVIS_PULL_REQUEST}" = "false" ]; then grunt ci; else grunt ci-pull; fi'
 env:
   global:
   - secure: Kt+5IJDJRVwr28xRnmR5YDsJceXDcDR21/JUBfk6DYFixPbIq7LCPnZUmiSZQs8akU95ucXwB5hsirUAdEhXdYKilec6go70lticVlZBLy8IdJ+Di1uPwMOeMHvalC2P0woIRJSMzP8u5E+e+5ASggTjsXID7/p1rE0jXtoOueQ=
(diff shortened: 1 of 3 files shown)


kum-deepak:
@gordonwoodhull The code seems to keep a flag to apply the projection again if it changes. I have a hunch that `_projectionFlag` can be removed.

If consider the general convention of dc, in case of any change other than data, render must be called. I think that it can be cleaned up.

I will create an independent commit so that you can reject that part without affecting the core work.

kum-deepak:
I am updating the VC example where in 6 seconds the projection is changed and the chart re-rendered. I am pushing that as a separate commit, the last commit should not be merged, it is only for testing.

kum-deepak:
I have disabled cross browser tests, those were always failing for me :)

Maybe we switch back to old scheme of invoking cross browser testing.

gordonwoodhull:
Yeah, I tried to find earlier discussion, but certainly disabling saucelabs for pull requests was intentional (16e686300158c19). So I'm reverting eff5bf963902f816d33a43bf8a608e6f90bb1b3e.

I wouldn't say that there is a convention that redraw is only for data changes. Rather I would say that redraw works for whatever people have debugged it for. So I think I'll keep `_projectionFlag`. No one ever said simplicity was one of dc.js's virtues. 😉

gordonwoodhull:
I took the first commit and tested using your VC example test, only with redraw not render, and with a more dramatic projection change (Mercator). Works great!

gordonwoodhull:
Thanks @kum-deepak!

kum-deepak:
Thanks @gordonwoodhull!

I am intending to devote time on this project for reasonable time. We are relying on this library heavily and will like it to do even more. To that effect, I will participate in cleaning up as much as possible.

I have realized that the code has been written over many years and has varying code quality across files.

Should we open a separate task for discussion - to cover what someone can expect out of redraw vs render. This will simplify the library and make life easier for users.

gordonwoodhull:
Sure, it's worth discussing. 

In general the functional difference is that redraw will transition from the old state to the new state, but  render starts from a blank SVG, so there is usually a blink. 

Zooming is one example of something besides the data changing, and although the choropleth doesn't have transitions, perhaps it could morph from one projection to another.

kum-deepak:
It makes sense. :) There is so much I do not know about this library.

PR #936, opened 2015-05-18
Adding an input filter

tttp:
Thanks to the suggestion on github, added throttling. It turns out that the result is smooth enough by simply waiting until the 200ms, rather than complicating and offering the option to draw separately slow and faster graphs.

The spec needs better test, but I'd need some help, I'm a bit lost

I'd love if someone could comment on the feature itself, I'll figure out what makes sense to test soon I think

PR #1387, opened 2018-03-29
Input filter dc v3

kum-deepak:
Started work. Will work on making in similar in code convention as other D3 modules. I will inform when I am happy with the work.

Based on #936


kum-deepak:
Marking it complete from my side for now. :)

My suggestion will be to choose a more appropriate name for this. Something to do with text/search/filter/widget. Will look forward for suggestions.

src/filter-input.js
```diff
@@ -0,0 +1,145 @@
...
+dc.inputFilter = function (parent, chartGroup) {
+    var INPUT_CSS_CLASS = 'dc-filter-input';
+
+    var _chart = dc.baseMixin({});
+
+    var _normalize = function (s) {
+        return s.toLowerCase();
+    };
+
+    var _filterFunctionFactory = function (query) {
```
gordonwoodhull: The coding convention in dc.js is to use a function declaration rather than a variable, for local functions that won't be replaced. (Using a variable for `_normalize` makes sense because it will change.)

kum-deepak: Did you mean for `filterFunctionFactory`?

`filterFunctionFactory` can change in some specific cases. There is get/set for this.

gordonwoodhull: Whoops, you're so right. Sorry for the noise.

gordonwoodhull:
I agree, 'inputFilter` isn't specific enough. Maybe `textFilterWidget` or something like that? 

@tttp, any opinions?

kum-deepak:
Renamed inputFilter to textFilterWidget. Ready for your review.

kum-deepak:
Rebased and changed base to dc.js/develop.

ntbrock:
@kum-deepak this PR #1387 patched on top of dc.js/develop and worked wonderfully in my application layer.

dc.textFilterWidget is a wonderful feature addition!  I would love to see this become part of 3.0 release.

PR #1394, opened 2018-04-01
D3v5 compatibility

kum-deepak:
Very few changes and all test cases pass. Good thing is that change for us can be done in backward compatible way.

Examples will need changes because they now use Promises. Once changed it works. As an example I have changed the stocks example.

We can finish it after dc v3 release.


gordonwoodhull:
As for the colors, there are good reasons why `schemeCategory20c` was removed. So we should switch probably to schemeCategory10c or a colorbrewer scheme e.g. as suggested in #831.

This means fixing a lot of tests, documenting the change pointing to the rationale in d3v5, and offering the workaround used here in the documentation. IMO we should not provide `schemeCategory20c` because it is misleading.

kum-deepak:
Fully agree!

Once other PRs have been merged, I will go ahead and make the changes.

Alternatively we can introduce a global switch to keep current behavior with a warning. In 3.1 we can make full switch.

BTW there are not many (less than 10) test cases that we would need to fix.

gordonwoodhull:
The color brewer colors are dramatically different but much more clear, so I agree we should probably do backward compatibility with a warning.

We don't currently have any global flags, but that's an interesting idea. Maybe it would be worth introducing a special facility under `dc` just for this, say `dc.compatFlags`, since we have so many things we want to fix while allowing backward compatibility.

In this case, there could be a flag, say `brewerCategoricalColors`, which defaults to false. The first time we need defaulted categorical colors, because the chart doesn't have any colors set, we check this flag and if it's still false, we use the copied colors from above and issue a warning that says we're moving to brewer and mentions `schemeCategory10c`, which is close to the old default but obviously has less colors.

If it's true, we use the color brewer colors with no warning.

kum-deepak:
I will do a commit implementing it schematically, we can then progress from there.

kum-deepak:
I have reached a milestone on this. Key changes:

- compat code
- d3.csv/d3.json changes
- `play-ground.html` got missed during D3v4 round, works similar to current version
- I realized two of resizing examples were failing, fixed now
- The current version resizing examples don't work (I had done some fixes during D3v4 work)
- Some places I have changed == to === and != to !==

These are mechanical changes, at this stage we can merge. The other work for flag and warning message etc., can be done a separate PR.

In addition, I will work onto to run tests on both D3v4 and D3v5 (as a separate PR)

gordonwoodhull:
Any reason I shouldn't merge this now?

kum-deepak:
I think we can merge this. Currently I am not aware of any issues in this regard.

gordonwoodhull:
Great, thanks @kum-deepak! I filed #1403 for removing schemeCategory20c.

Released as 3.0.0-alpha.5 (but still not to npm).

Issue #1071, opened 2015-12-21
silly filter outro on bar clickaway


gordonwoodhull:
When you click an empty spot to get rid of the bar chart filter, there's a silly animation.

![silly filter outro](https://cloud.githubusercontent.com/assets/1366709/11940212/20fc45e6-a7f5-11e5-95d7-8c37088a7862.gif)

It's cute, but no.

gordonwoodhull:
Also on scatterplot.

Frozenlock:
FWIW, I discovered this animation when grabbing the last minified version from here.

My old version (2.0.0-beta19) didn't have this.

gordonwoodhull:
Thanks @Frozenlock, that's interesting. If anything I would have guessed it was introduced by the chart resizing stuff around beta 15.

Looks like it's transitioning to a zero-width brush instead of an empty brush. I don't think there should be any transition at all.

Frozenlock:
I don't know if I should open a separate issue for this, but it's another weird animation issue that wasn't present in 2.0.0-beta19 but is in the most recent one:

Bars are now changing width (?!) and don't transition correctly when they need to disappear.


**Old**
![old-dc-js](https://user-images.githubusercontent.com/1152041/27748901-3ba65308-5d9f-11e7-8718-bf622349c3c8.gif)

**New**
![new-dc-js](https://user-images.githubusercontent.com/1152041/27748905-40687f74-5d9f-11e7-82ea-8c8cd020e6bb.gif)

gordonwoodhull:
Hi @Frozenlock, sorry I missed this. If you're still seeing this and can file a separate issue, that will help a lot. Especially if you can create a demo fiddle.

I don't think I've seen seen this, and I don't see it in any of the standard examples. So it must mean that you've run into some combination of features that we don't test. (Keep in mind that the interaction of features is combinatorial!)

Frozenlock:
@gordonwoodhull I haven't seen this since I upgraded to a newer version... but I also made a major overhaul in pretty much everything.

I'll pop in here if I ever encounter it again.

Issue #1216, opened 2016-10-14
Brush behaves erratically if scatterPlot is placed before seriesChart


cramatt:
I'm experiencing strange brush behavior in the following scenario: 

- verison 2-beta.32 
- scatterPlot with brushing on
- seriesChart with brushing off
- both charts placed on same page
- scatterPlot placed BEFORE seriesChart

If I reverse the order the brush works as expected.

cramatt:
Note this is likely what I was experiencing here: https://github.com/dc-js/dc.js/issues/878#issuecomment-253790527

gordonwoodhull:
Hi @cramatt, sorry for the slow response. I'm confused how this could be #878, since that's about filtering the composite chart, and here it sounds like you're brushing on a solitary scatterPlot.

Could you describe the strange brush behavior?

Issue #878, opened 2015-03-11
Brush not filtering when using composite chart


gramster:
Is this supposed to work?

If I create a line chart with a data table, selecting an area of the chart with the brush filters the data in the table. If I take that same line chart and wrap it in a composite chart (even if it is the only chart in the composite), the data table no longer gets filtered.

My code is the same in each case except for this:

```
        dcl = dc.lineChart("#dc-chart");
        dcl.dimension(dateDim)
            .group(open, "Open");
```

versus:

```
        dcl = dc.compositeChart("#dc-chart");
        dcl.compose([
                dc.lineChart(dcl)
                    .dimension(dateDim)
                    .group(open, "Open")
      ]);
```

gordonwoodhull:
This is a duplicate of, well, a lot of things. 
https://github.com/dc-js/dc.js/search?q=filter+composite&type=Issues&utf8=%E2%9C%93

This branch is a start on a solution:
https://github.com/dc-js/dc.js/tree/sync-filter

Please try it out and see if it helps.  Will merge to develop soon.

jbliesener:
Any news on this one?

Fil:
I'm stump too

Fil:
This fixed it for me:

https://groups.google.com/forum/#!topic/dc-js-user-group/yI6_cbvgfbU
http://jsfiddle.net/cBgkT/1/

joysword:
As of 2.0.0-beta32, CompositeChart's _chart._brushing() still does not trigger any redraw events. The google group link posted by Fil did not work for me.

A solution that worked for me is to simply comment out CompositeChart's _chart._brushing() function. This makes CompositeCharts use CoordinateGridMixin's _chart._brushing() for brushing, and works just fine.

gordonwoodhull:
Thanks @joysword, that sounds like a good workaround, as long as the dimension is the same for the composite and its children. I think we could make it possible to restore the overridden method without having to change the source.

phsstory:
@Fil  updated for 404'd external resources http://jsfiddle.net/cBgkT/42/

The redefined _chart._brushing function works if the composite graph has the defined dimension but fails if the sub-charts have defined dimensions.

cramatt:
When I override the `_chart._brushing()` method I get unexpected behavior - the filtered set and brush jump around sporadically.  Using 2.0.0-beta32.

gordonwoodhull:
Fixed by #1408 in 3.0.2

Issue #479, opened 2013-12-13
series chart does not work as range chart


hudannis:
Hi, I am using seriesChart to draw multiple lines in one chart with a selecting chart. But it seems seriesChart doesn't work with rangeChart. I select a range by brush but the main chart doesn't change as expected.
expected image:
![expected](https://f.cloud.github.com/assets/5782042/1740836/8a93793c-63d4-11e3-8682-0a1c00dedf0f.png)

get the following image:
![notchange](https://f.cloud.github.com/assets/5782042/1740839/ae6081c0-63d4-11e3-9182-c24e29204470.png)

Moreover, if the mainchart can show the selected range of the indicator chart, can the data be selected correctly? Because the dimension for the seriesChart has two keys, while the filter getting by the selected range has only one key.

peter-houck:
I tried hacking series.html example to add a rangeChart.  The brush still does not filter the main chart.  Any ideas? Here is the hacked up code:

`<!DOCTYPE html>
<html lang="en">
<head>
    <title>dc.js - Series Example</title>
    <meta charset="UTF-8">
    <link rel="stylesheet" type="text/css" href="../css/bootstrap.min.css">
    <link rel="stylesheet" type="text/css" href="../css/dc.css"/>
</head>
<body>

<div class="container">
<script type="text/javascript" src="header.js"></script>
<div id="test"></div>
<div id="test-overview"></div>
<button onclick="button1()">1</button>
<button onclick="button2()">2</button>
<button onclick="button3()">3</button>

<script type="text/javascript" src="../js/d3.js"></script>
<script type="text/javascript" src="../js/crossfilter.js"></script>
<script type="text/javascript" src="../js/dc.js"></script>
<script type="text/javascript">

var chart = dc.seriesChart("#test");
var overviewChart = dc.seriesChart("#test-overview");
var ndx, runDimension, runGroup, overviewRunDimension, overviewRunGroup;

d3.csv("morley.csv", function(error, experiments) {

  ndx = crossfilter(experiments);
  runDimension = ndx.dimension(function(d) {return [+d.Expt, +d.Run]; });
  overviewRunDimension = ndx.dimension(function(d) {return [+d.Expt, +d.Run]; });
  runGroup = runDimension.group().reduceSum(function(d) { return +d.Speed; });
  overviewRunGroup = overviewRunDimension.group().reduceSum(function(d) { return +d.Speed; });

  chart
    .width(768)
    .height(480)
[...]

PR #1402, opened 2018-04-10
Brush animation

kum-deepak:
Looks like I have some understanding of transitions now.

Only changes in coordinate-grid-mixin.js so far. Test cases fail for derived classes with brush. For your initial feedback.


kum-deepak:
Completed for now. I have currently enabled transition only for resizes. During brushing we definitely should not enable transitions.

Let me know if it makes sense to enable it for any other case as well.

gordonwoodhull:
The only other case I can think of where an animated brush might be nice is when setting the brush programmatically. But I don't know if we can detect that, and I agree it's definitely annoying to animate the brush when dragging it. (E.g. you've fixed #1071)

Thanks for `dc.utils.arraysEqual`, good idea!

kum-deepak:
I will rebase it again before working further.

web/examples/range-series.html
```diff
@@ -84,7 +84,9 @@
                 });
             } else if (!rangesEqual(chart.filter(), overviewChart.focusChart().filter())) {
                 dc.events.trigger(function () {
-                    overviewChart.focusChart().focus(chart.filter());
+                    // The second parameter is quite important. It asks the focus operation not to propagate events
+                    // In absence of that it will cause infinite mutual recursion
+                    overviewChart.focusChart().focus(chart.filter(), true);
```
gordonwoodhull: Have you considered defaulting this the other way? I'm not clear which circumstances lead to infinite recursion, or which way is more backward-compatible

kum-deepak: I will actually love to name the 2nd parameter in `_chart.focus` to `raiseEvents`  instead of `noRaiseEvents`. To use default value `false`, I had to settle for that name.

kum-deepak:
I have rebased to 3.0 again.

Brush transition in Y is now working, line chart (composite chart) brushing and transition works.

Need to add resizing examples for Bubble and Scatter, once added I can check the functionality.

Signing off for today.

kum-deepak:
Cleaned up brushing in composite chart. The strategy I am taking is to enable and apply brushing only in the container. While applying a filter, it is applied to all children.

Visual examples work. My next course of action:

- Fix #1404 before I can move on this one.
- Add example for bubble chart brushing and resizing.
- Reverse the default for second argument for `chart.focus`.

gordonwoodhull:
Cleaning up composite brushing is a noble effort. I can guarantee that you will fix some issues, but also break some other things. 

I agree it's the right approach, but just be prepared for blowback. At the very least, everyone who has worked around this one way or another is likely to break now.

Some of the earlier issues: #878 #1216 #390 #479

kum-deepak:
Wow! Ignorance can be bliss sometimes. Now I realize that I build upon earlier work of many in this regard. The current work is likely to fix most, if not all, of the issues that you have referred. 

The approach I have taken is similar to the one earlier discussed.

I am already quite way ahead in this. I will fix the title issue and then actually try to use series chart as a range chart. I have mixed feelings that it might work now :)

Also, I have now started realizing that X axis filter and 2D filter (only for scatter plots) - currently it is determined by the chart type. Interesting case happens when scatter plot is used inside a composite chart - where it is supposed to work with X axis brush. I was interesting to see that brushing test cases do put a scatter chart inside composite chart.

So, it appears that brush type should not depend on chart type, but coordinate mixin should offer it as a setting. As of now I am not sure about best API approach. I will create a marker issue for it, so that I can come back to it.

kum-deepak:
I tracked the reason for #1404, There is following code in `line-chart.js`:

```javascript
    function drawDots (chartBody, layers) {
        if (_chart.xyTipsOn() === 'always' || (!_chart.brushOn() && _chart.xyTipsOn())) {
```

Since I am now setting `brushOn(false)` for integrated charts, the condition is now true which is making the test case to fail.

I need to know here if this chart is part of a composite chart and if the parent has brush switched on. I see following approaches:

- Add a field `parentBrushOn`
- If there is a way to find out this chart has a parent, rewiring some functions to call out to parent. This definitely has more work. But it may be worth to try it out (or some other generic approach) post 3.0.

Currently going ahead with `parentBrushOn`

gordonwoodhull:
I like PRs to have one distinct purpose if possible. 

I'm glad that you are tackling all of these other issues related to composite brushing, and I mostly agree with your approach. However, these fixes have a lot more risk than the original intent of this PR, which was to fix brush resizing.

I'm going to merge the commits for brush resizing now. Please create separate PRs for the other features. 

In particular, I'd like to merge the composite brushing stuff after we go beta, so that when it breaks something (and it will, and it's nothing to fear), people can say "gee, everything was working fine with 3.0 beta 1 but 3.0 beta 2 causes the following problems..."

kum-deepak:
Yes, I agree. Once it has been merged I will create a new PR with remaining changes.

Thanks!

gordonwoodhull:
Great, thanks for understanding. It looks like the two purposes got entwined in c001c0eb - so I'll check if the line chart brush still goes away after resizing, and open a separate issue for that if necessary.

gordonwoodhull:
Awesome, I took just the related commits and brush resizing works for all the examples.

Merging for 3.0 alpha 7. When you rebase your branch I think you should find the commits I didn't merge - lmk if you have any questions.

PR #1407, opened 2018-04-14
Remove brush handles for scatter-plot

### LINKED ISSUE #1406: horizontal brush is showing in scatter plot
screen shot 2018-04-14 at 12 49 57 pm

It looks like somehow we have two brushes in the scatter plot. The XY brush works fine, but the horizontal brush crashes if you try to drag it.

I haven't looked closely at the implementation to see why this is.

kum-deepak:
Fixes #1406

### DIFF
--- a/spec/scatter-plot-spec.js
+++ b/spec/scatter-plot-spec.js
@@ -234,6 +234,11 @@ describe('dc.scatterPlot', function () {
                 chart.redraw();
             });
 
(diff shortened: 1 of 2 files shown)


kum-deepak:
Sorry #1402 went too ambitious. I separated out just this bit. Other changes I will submit as a new PR.

PR #1409, opened 2018-04-15
Remove schemeCategory20c

kum-deepak:
This PR contains following:

- Concept of `dc.config`, which will be a single instance object. It currently holds one setting and one variable. Please see if you are ok with the concept.
- For backward compatibility, following needs to be added before creating charts:
```javascript
    dc.config.useV2CompatColorScheme(true);
```
- Concept of standard colors in specs. This will help us in minimizing breakage when making global changes in color palette.
- A special matcher for color which covers more cases than a reg ex.
- Enabled back `index.html` generation in `specs` folder, this allows to run test cases interactively by going to http://localhost:8888/spec/ - useful for interactive debugging.

Charts work but they do look different :)

### DIFF
--- a/spec/color-spec.js
+++ b/spec/color-spec.js
@@ -1,4 +1,4 @@
-/* global loadDateFixture, standardColors */
+/* global loadDateFixture */
 describe('dc.colorMixin', function () {
     function colorTest (chart, domain, test) {
         chart.colorDomain(domain);
@@ -18,8 +18,7 @@ describe('dc.colorMixin', function () {
 
         it('default', function () {
             expect(colorTest(chart, domain))
-                .toMatchColors([standardColors[0], standardColors[1],
-                    standardColors[2], standardColors[3], standardColors[4]]);
+                .toMatchColors(dc.config.defaultColors().slice(0, 5));
         });
 
         it('custom', function () {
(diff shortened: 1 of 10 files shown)


kum-deepak:
If you like the concept of `dc.config`, it can be used for more things. One immediate usage may be to keep the default ordinal colors (which currently lie in color mixin).

spec/helpers/standard-colors.js
```diff
@@ -0,0 +1,8 @@
+/* exported standardColors */
+
+/* Use these in tests to avoid breaking tests if default colors are changed */
+var standardColors = [
+    '#e41a1c', '#377eb8', '#4daf4a',
+    '#984ea3', '#ff7f00', '#ffff33',
+    '#a65628', '#f781bf', '#999999'
+];
```
gordonwoodhull: Why copy `d3.schemeSet1` here? It's not going to change in d3 and if it did, it would still break all our tests.

src/config.js
```diff
@@ -0,0 +1,44 @@
...
+ * General configuration
+ *
+ * @class config
+ * @memberof dc
+ * @returns {dc.config}
+ */
+dc.config = (function () {
+    var _config = {};
+
+    var _useV2CompatColorScheme = false;
```
gordonwoodhull: I think we should default to true and issue the warning. This way people get backward compatibility by default, to ease their transition into the new world, but they are also informed that it's going away.

src/config.js
```diff
@@ -0,0 +1,44 @@
...
+    _config.compatColors = null;
+    var _schemeCategory20c = [
+        '#3182bd', '#6baed6', '#9ecae1', '#c6dbef', '#e6550d',
+        '#fd8d3c', '#fdae6b', '#fdd0a2', '#31a354', '#74c476',
+        '#a1d99b', '#c7e9c0', '#756bb1', '#9e9ac8', '#bcbddc',
+        '#dadaeb', '#636363', '#969696', '#bdbdbd', '#d9d9d9'];
+
+    /**
+     * Set this flag to use DCv2's color scheme (d3.schemeCategory20c).
+     * Please note this color scheme has been deprecated by D3
```
gordonwoodhull: It's been **removed** in d3v5.

src/config.js
```diff
@@ -0,0 +1,44 @@
...
+     * (https://github.com/d3/d3/blob/master/CHANGES.md#changes-in-d3-50) as well as DC.
+     * @method useV2CompatColorScheme
```
gordonwoodhull: Let's add an example of what you should do if you want to keep the old colors and not get the warning. Namely, you'd copy `_schemeCategory20c` and apply it to each chart.

Alternately, since we are adding a level of indirection here, we could rename `compatColors` to `defaultColors` and make it settable and permanent, so that users could automatically have the same color set for all charts. That might be a good idea if it turns out that `schemeSet1` was the wrong choice.

spec/color-spec.js
```diff
@@ -48,7 +50,8 @@ describe('dc.colorMixin', function () {
         });
 
         it('default', function () {
-            expect(colorTest(chart, domain, test)).toMatchColors(['#9ecae1','#3182bd','#c6dbef','#6baed6','#e6550d','#3182bd']);
+            expect(colorTest(chart, domain, test)).toMatchColors([standardColors[2], standardColors[0],
+                standardColors[3], standardColors[1], standardColors[4], standardColors[0]]);
```
gordonwoodhull: This test is really opaque - is there any way to make it clear what it proves? I can't figure it out, it's not numerical or alphabetical order is it?

(We have this problem in a lot of our tests, and it's unfortunate because it means people will "fix the tests" instead of understanding why they failed.)

spec/color-spec.js
```diff
@@ -17,7 +17,9 @@ describe('dc.colorMixin', function () {
         });
 
         it('default', function () {
-            expect(colorTest(chart, domain)).toMatchColors(['#3182bd','#6baed6','#9ecae1','#c6dbef','#e6550d']);
+            expect(colorTest(chart, domain))
+                .toMatchColors([standardColors[0], standardColors[1],
+                    standardColors[2], standardColors[3], standardColors[4]]);
```
gordonwoodhull: `standardColors.slice(0, 5)`

spec/helpers/standard-colors.js
kum-deepak: Actually I like the idea of `defaultColors` in config, once we move there, we can use that to match, this variable would not be needed at all.

src/config.js
kum-deepak: Will do!

src/config.js
kum-deepak: Thanks, will change!

src/config.js
kum-deepak: If we allow setting defaultColors, we need to add in the API docs. In addition we may create an example to show how to use their color scheme.

spec/color-spec.js
kum-deepak: I have realized what it does, I will refactor it is a way that it does not seem so opaque :)

spec/color-spec.js
kum-deepak: Will do!

kum-deepak:
I will make one more iteration and then we can have one more round of discussion. I want to get it right :)

kum-deepak:
Moved to  concept of `dc.config.defaultColors`. Refactored few test cases for readability and resilience. Added additional notes for opaque-ish test cases.

If we are happy with the approach, I will add the warning message.

For one time warning messages, I want to create one helper method in `dc.logger`, which will ensure one message is shown only once.

gordonwoodhull:
Huh, until now I thought we need the flag *as well as* `defaultColors`.

We do need to warn when `defaultColors` is not set, but maybe it's sufficient to test that it is *not* `_schemeCategory20c` - is that what you are thinking?

kum-deepak:
Yes, that indeed was my thinking - this led me to the need of warnOnce (in dc.logger). Simplifies the code :)

kum-deepak:
Took a stab at adding to Change log as well. :) English is my 3rd language, so you will need to review these text as well.

gordonwoodhull:
Sure, I don't mind doing copy editing. It's really helpful if you put something there so it doesn't get lost.

kum-deepak:
Completed my current round of work on this one.

gordonwoodhull:
Merged in 3.0 alpha 10

PR #1410, opened 2018-04-15
Coordinate mixin cleanup - xUnits

kum-deepak:
This function took one of my longest time while I was working on the D3 v4 to comprehend about how it is working. Then I realized the code makes big distinction on whether Ordinal or Continuous scale is in use. This prompted me to make the logic explicit that what this function does in each case.

In this PR I have rearranged the code to make it more explicit about what it does.

### DIFF
--- a/D3v4-Upgrade-Notes.md
+++ b/D3v4-Upgrade-Notes.md
@@ -33,3 +33,12 @@
 - Consider removing transitions for zooming and brushing
 - d3 stack to D3v4
 - Changing .tension call in lineChart
+
+
+### For end users
+
+- Changes to `xUnits` interface.
+- geoChart
+- Color scheme
+- tension in curves
+
--- a/src/coordinate-grid-mixin.js
+++ b/src/coordinate-grid-mixin.js
@@ -390,7 +390,8 @@ dc.coordinateGridMixin = function (_chart) {
 
     /**
      * Returns the number of units displayed on the x axis using the unit measure configured by
-     * {@link dc.coordinateGridMixin#xUnits xUnits}.
+     * {@link dc.coordinateGridMixin#xUnits xUnits}. If it is an Ordinal chart, it will return
+     * number of items in the domain.
      * @method xUnitCount
      * @memberof dc.coordinateGridMixin
      * @instance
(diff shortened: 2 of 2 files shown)


kum-deepak:
I also checked my notes from the days and I realized in case of non ordinal charts, having the 3rd parameter present causes test failure.

In D3v3 code `_chart.x().domain()` was always passed as the 3rd parameter - but the receiver (other than ordinal) was supposed to discard it. It seems in D3v4, some unit functions are impacted by the 3rd parameter. If we add the 3rd parameter at line 406, some tests fail.

gordonwoodhull:
I was always confused by `xUnits` sometimes returning a number and sometimes an array, so in principal I'm in favor of this.

However, if d3v4 is forcing us to change our API for one of the more confusing corners of dc.js, that's kind of a big deal and we need to document it. In particular it means that a user can't pass their own function that acts like `dc.units.ordinal` - I doubt too many people did, but it was possible and now it's not.

What tests fail using the third argument? I bet we were misusing it and d3 got stricter about it.

gordonwoodhull:
Actually I think this change means that `dc.units.ordinal` is never called at all - it's just a magic variable. That's not a great design.

gordonwoodhull:
Alright I'm off to do weekend things. Good work! :)

D3v4-Upgrade-Notes.md
```diff
@@ -33,3 +33,12 @@
...
 - Changing .tension call in lineChart
+
+
+### For end users
+
+- Changes to `xUnits` interface.
+- geoChart
+- Color scheme
+- tension in curves
+
```
gordonwoodhull: I am documenting all of these in Changelog.md, and also in the API documentation. We could do a D3-style CHANGES summary as well

gordonwoodhull:
Okay, I have now over-documented this, made the function throw, tested the throw, etc. Hopefully people will notice somehow.

gordonwoodhull:
Thanks @kum-deepak, merged for 3.0 alpha 9

kum-deepak:
I really think it can not be documented more :)

PR #1413, opened 2018-04-16
Number display transition

kum-deepak:
Your suggestion makes sense. 'quad-out-in' was not proper and may be the reason that D3v4 removed it. Changed it to `d3.easeQuad`.


gordonwoodhull:
Please put even the smallest behavior changes into `Changelog.md`

gordonwoodhull:
Thanks @kum-deepak! merged for 3.0 alpha 9

PR #1414, opened 2018-04-16
Prepare x axis - added warning

kum-deepak:
Warn before converting d3.scaleOrdinal() to d3.scaleBand(). Updated all examples accordingly.

There is one unrelated change in this PR - I have removed /web/ep/list.js. It seemed stale and was probably not getting used.


gordonwoodhull:
That sounds good! I guess you'd be able to test both that the warning happens and that the scale gets replaced.

kum-deepak:
Finished work for this round, ready for your review. 👍

gordonwoodhull:
Great work, thanks @kum-deepak!

Merged for 3.0 alpha 10.

Issue #1398, opened 2018-04-07
d3v4 cleanup


gordonwoodhull:
Hi @kum-deepak!

As promised, I'm doing a thorough line-by line review of #1363 d3v4 support.

Rather than comment inline on a PR that has already been merged, I decided to start a new issue.

I think I have gotten through about 1/3 of the changes, and I'm learning a lot along the way!

Here are my comments so far. I'm going to go for a Saturday walk now. :)

### "selection"
The argument name `selection` is confusing for `fadeDeselectedArea`/`brushIsEmpty`/`extendBrush` / `scatterPlot.brushIsEmpty` `scatterPlot._brushing` because it sounds like it should be a d3 selection. Personally I would use the term `filter` or `extent` here. 

I guess this arose because of `event.selection` from [Brush Events](https://github.com/d3/d3-brush#brush-events) so maybe `brushSelection` would be a good compromise. Kind of confusing that they reused such a common term.

## composite chart
`rightYAxis` documentation: "depending on whether you are going to use the axis on left or right" doesn't make sense, as it should always be `d3.axisRight`. I guess this was copied from `coordinateGridMixin.yAxis` but it's different.

### `xUnitCount` 
This function now distinguishes `dc.units.ordinal` but it doesn't need to because calling the function would have exactly the same effect. I'd rather not add the special case unless it's necessary for some reason? The design of `xUnits` functions is perhaps a bit strange but it is generic.

If this really needs to be then it should at least use `chart.isOrdinal()`.

### `prepareXAxis`
I agree that we need to use `d3.scaleBand` instead of `d3.scaleOrdinal` -- it's so much cleaner! -- but it will confuse users if their input `.x()` scale is changed. We should issue a warning every time we make the replacement so users can update their code accordingly.

### `prepareYAxis`
```js
       // Ideally we should update the API so that if someone uses Right Y Axis
       // they would need to pass _yAxis as well
       if (!_yAxis) {
           if (_useRightYAxis) {
                _yAxis = d3.axisRight();
            } else {
                _yAxis = d3.axisLeft();
            }
         }
```

I respectfully disagree with the comment - it's really nice how the axes are automatically initialized and I think we should be able to extend that to the right axis. 

This code will never get hit because `_yAxis` is initialized and the user would never set it to undefined or null. 

Could we start with `_yAxis` null and initialize it like we're doing here, when
* this function is hit, **or**
* when `.yAxis()` is called as a getter?

[...]

PR #1416, opened 2018-04-17
_prepareYAxis changes and Axis related documentation updates

kum-deepak:
Fixes part of #1398.

`_yAxis` is initially null and gets initialized when first needed based on whether it will left or right axis. Also added a warning in case axis orientation changes after y axis was initialized.

Several corrections in documentation of axes.

### DIFF
--- a/src/coordinate-grid-mixin.js
+++ b/src/coordinate-grid-mixin.js
@@ -40,7 +40,7 @@ dc.coordinateGridMixin = function (_chart) {
     var _lastXDomain;
 
     var _y;
-    var _yAxis = d3.axisLeft();
+    var _yAxis = null;
     var _yAxisPadding = 0;
     var _yElasticity = false;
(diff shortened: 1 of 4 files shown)


src/coordinate-grid-mixin.js
```diff
@@ -436,6 +436,14 @@ dc.coordinateGridMixin = function (_chart) {
...
             return _useRightYAxis;
         }
+
+        // We need to warn if value is changing after _yAxis was created
+        if (_useRightYAxis !== useRightYAxis && _yAxis) {
+            dc.logger.warn('Value of useRightYAxis has been altered, after yAxis was created. ' +
+                'You might get unexpected yAxis behavior. ' +
+                'Make calls to useRightYAxis sooner in your chart creation process.');
+        }
+
```
gordonwoodhull: Nice!

PR #1417, opened 2018-04-17
pie-chart code cleanup

kum-deepak:
Fixes part of #1398. See https://github.com/dc-js/dc.js/issues/1398#issuecomment-379861299

Cleaned up pie-chart code, removed unnecessary return values. Code now much closer to DCv2.1 version.

Ready for review.


gordonwoodhull:
Thanks @kum-deepak, much cleaner.

Merged for 3.0 alpha 10

PR #1418, opened 2018-04-18
Scatter plot brushing - more defensive

kum-deepak:
While checking for brush being empty use >= instead of ===. After brush rounding check again for brush being empty. Fixes part of #1398


gordonwoodhull:
Thanks @kum-deepak! Merged for 3.0 alpha 10

Issue #1401, opened 2018-04-08
exiting bars turn grey first


gordonwoodhull:
From https://github.com/dc-js/dc.js/pull/1363/#issuecomment-373955979

> In the multi focus example, bars turn grey before being filtered/focused (?) out instead of just disappearing. Not sure which behavior is better here, but we should figure out what is going on here and decide one way or the other.

Also, the bars stay deselected when the range chart is reset to no filter:

<img width="880" alt="image" src="https://user-images.githubusercontent.com/1366709/38468880-2dca867e-3b1a-11e8-93e1-f630992327d0.png">

Looks like the reset bug can be fixed by simplifying the example. Applying `fadeDeselected` to the focus charts is arguably correct but distracting & annoying.

gordonwoodhull:
Feel free to take it. I guess we are applying `fadeDeselected` earlier than we used to.

kum-deepak:
Tried to find a solution. However I am unable to find any simple code fix. Please see if you are able to find a solution.

kum-deepak:
Please check #1422

gordonwoodhull:
Your fix makes sense. Will check this morning.

gordonwoodhull:
Fixed in 3.0 beta 1

PR #1422, opened 2018-04-20
Fixing fade deselected. Possible fix of #1401

kum-deepak:
Bit of a coincidence, figured out why bar were turning gray in 3.0 in multi focus example. It also fixes strange fading in http://localhost:8888/web/examples/filter-stacks.html.

In 2.1, `fadeDeselectedArea` was effecting the fade only if the chart had brush on. I have added these as explicit check now.

See, if this should be accepted as a fix.


gordonwoodhull:
Merged for 3.0 beta 1

gordonwoodhull:
Rats, this broke ordinal bar selection, eg 

http://127.0.0.1:8888/web/examples/ordinal-bar.html

vs 

https://dc-js.github.io/dc.js/examples/ordinal-bar.html

@kum-deepak, do you remember any other special cases having to do with `isOrdinal`?

gordonwoodhull:
Thanks @kum-deepak, I tried moving it to the non-ordinal section of `barChart.fadeDeselectedArea` and that fixed it. I'll  push that fix to 3.0.

However, there are still infinite recursion bugs on range-series. I'll see if #1408 helps with that and file an issue.

PR #1388, opened 2018-03-29
Sunburst dc v3

kum-deepak:
This is an straight upgrade of #1337 with as little changes that I could.

Will look into qualitative aspects soon.

PR #1426, opened 2018-04-23
Colors - stable test case

kum-deepak:
This is low risk PR. This should fix spec failure in #1388


kum-deepak:
I also have a suspicion that some minor release in D3 v5 has change some minor behavior that breaks one of the test cases. I will investigate separately.

gordonwoodhull:
merged for 3.0.0. Thanks @kum-deepak!

PR #1430, opened 2018-04-30
Travis builds for d3v4 and d3v5

kum-deepak:
I have updated Travis so that on every push it starts two jobs in parallel, one for d3v4 and another for d3v5.

A test case in color-spec is patched so that it succeeds for d3v4 as well.


kum-deepak:
@gordonwoodhull when you do a commit, it should run Sauce lab tests for both versions. Let us see if that works as expected.

kum-deepak:
@gordonwoodhull ready for your review.

gordonwoodhull:
I agree, it's good to verify compatibility with both D3v4 and D3v5.

I'm surprised that we don't use d3-request or d3-fetch anywhere in the specs, which would make this more difficult.

My first thought looking at this was "can't we detect the version of d3 from JavaScript rather than modifying files?" It's easy to do in the Jasmine tests - just look at `d3.version.split('.')[0]`. But I don't know if there is an easy way to do this in the Gruntfile.

Any thoughts?

kum-deepak:
There are two places we needed change:

- The test case, I do like the idea of checking the d3 version and activate one of the version of the tests.
- In Gruntfile we can setup a variable based on d3 version.

At that stage we will not need this file, as we would only need to install d3 v4 which can be in .travis file directly.

Let me get cracking.

kum-deepak:
Take 2 ready

gordonwoodhull:
Nice use of a library for the version comparison. 

Thanks @kum-deepak, merging for 3.0.1!

kum-deepak:
You can consider merging it to `develop` branch. That way all new PRs targeting `deveop` would get tested in both versions of D3.

gordonwoodhull:
It's merged to develop already.

PR #1366, opened 2018-02-27
Subscribe to rangeChart's filtered event with a namespace

dahlbyk:
Fixes #842. It's trivial to change, but for now I've chosen a more specific namespace than `.dcjs`, considering the possibility of internal conflicts. I couldn't find any other examples of internal event subscriptions.

PR #1435, opened 2018-05-13
Propagate filters on composite chart to children fixes #677 and #706

### LINKED ISSUE #677: Composite ordinal chart doesn't redraw filtered then reseted bars properly

kum-deepak:
Final change is quite small.

All filter related calls eventually call `.filter`, so covering just one case proved to be enough. Got a simpler code by listening to 'filtered' event and replacing filters on children. Overriding was getting messy because coordinate mixin already overrides `filter`.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,6 @@
+## 3.0.4
+* Propagate filters on composite chart to children, by Deepak Kumar ([#1435](https://github.com/dc-js/dc.js/pull/1435)). Fixes remaining parts of ([#390](https://github.com/dc-js/dc.js/issues/390) / [#706](https://github.com/dc-js/dc.js/issues/706)).
+
 ## 3.0.3
 * Update versions and release new fiddles and blocks pegged to dc@3 and d3@5
 
--- a/spec/composite-chart-spec.js
+++ b/spec/composite-chart-spec.js
@@ -262,8 +262,12 @@ describe('dc.compositeChart', function () {
             });
 
             describe('when filtering the chart', function () {
+                var filter1, filter2;
+
                 beforeEach(function () {
-                    chart.filter([makeDate(2012, 5, 1), makeDate(2012, 5, 30)]).redraw();
+                    filter1 = [makeDate(2012, 5, 1), makeDate(2012, 5, 30)];
+                    filter2 = [makeDate(2012, 6, 10), makeDate(2012, 6, 20)];
+                    chart.filter(filter1).redraw();
                 });
 
                 it('should set the extent height to chart height', function () {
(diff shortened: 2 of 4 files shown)


gordonwoodhull:
Seems too simple somehow! 😄 

If this is correct, then `compositeChart.applyBrushSelection` can be eliminated too.

`.on('filtered', ...` should definitely be namespaced because users will listen to it too.

I'll try to devise more tests / examples.

kum-deepak:
Thanks @gordonwoodhull, I had not realized that `compositeChart.applyBrushSelection` has actually become redundant 😃 

Also, had not realized behavior of d3.dispatch with same event names - changed in composite chart, also changed another similar usage in coordinateGridMixin.

I think additional test can be added to one of the existing test groups for setting/removing filter(s) and checking in children. Would you like to take a shot?

kum-deepak:
Added few additional test cases. I think these should cover the updated functionality.

kum-deepak:
I noticed that #1366 also fixes the 'filtered' listener in `coordinateGridMixin`. I think the code in this PR is more consistent with rest of the dc. However it may be worth picking up the test case from #1366.

gordonwoodhull:
Ah, thanks for pointing that out! I'll merge that along with this.

gordonwoodhull:
Thanks @kum-deepak! Merged for 3.0.4

Amazing how simple this turned out to be, after all the other fixes anyway.

I went with namespacing the event with `dcjs-range-chart` and `dcjs-composite-chart`. Rationale is that we still want the user to be able to guess the namespace and remove/replace the listener if they need to. 

The unique ids we use for charts in a page, but there is no risk of the charts interfering with each other here.

I didn't manage to write any more composite chart examples today. Hope to return to that another day!

Issue #1370, opened 2018-03-13
Enhanced BoxPlot chart


cwolcott:
I plan to submit an enhanced BoxPlot chart that is fully backward compatible.  This enhancement will also fix issue #1120.  I have never submitted a pull request, but would like to learn so I can help maintain this chart and potentially help with others.  

Currently I am using a different name for this chart (BoxPlotJitter), but I will change it back to BoxPlot.  It looks like I need to submit changes for src/box-plot.js, src/d3.box.js and style/dc.scss.

We are currently running against dc.js 2.1.9.

I will build a jsfiddle example and post the link later today to show the enhancements.

cwolcott:
Here is a [jsfiddle](https://jsfiddle.net/yjxLr5gu/17/) example of the BoxPlot Enhancements.  The example looks best with bottom results. 

Row 1 - BoxPlot (Original BoxPlot from dc.js 2.1.9)
Row 2 - New BoxPlotJitter (No new attributes) *
Row 3 - New BoxPlotJitter (All new attributes enabled)

The browser warnings/errors are coming from the original BoxPlot.  You can disable the original BoxPlot charts by toggling the variable showOldBoxPlots in jsfiddle.

* Row 2 shows that the enhanced BoxPlot is backward compatible, except for the fix to the yAxis (#1120).

Row 3 has all the new attributes enabled:
- renderData(true) - Show the actual values as opaque circles;
- renderTitle(true) - Show tooltips for each value;
- dataBoxPercentage(0.50) - Use a percentage of the box to display the data (0.00 - 0.99)
- boldOutlier(true) - Display the outlier as a red size 3 circle

Enabling the renderData is more impressive when you have a lot of data to show.
![boxplot databoxpercentage](https://user-images.githubusercontent.com/3777526/37349652-4891188c-26ad-11e8-83c6-19d31bf1e5e6.png)

gordonwoodhull:
That looks really great. I can see how it's really useful to show the data points.

I think the original d3 boxplot plugin that the chart was based on has been abandoned

https://github.com/d3/d3-plugins/tree/master/box

So there should be no issue modifying that file. And of course it's never a problem to add more styles.

There is something funny in your demo where if the window is initially narrow, the chart gets squished, which doesn't seem to happen with the old chart. I guess this is partly because you're not specifying the width in this example, but it might be worth looking into.

gordonwoodhull:
Hi @Zaknarfen. 

[...]

Issue #1120, opened 2016-03-18
Box plot y axis extent is too large on small numbers


amergin:
I'm seeing an issue where my data is somewhere between [0, 0.5] yet the chart is drawn with a clearly larger extent:

``` javascript
d3.extent(config.reduced.top(Infinity)[0].value)
[0.003021, 0.832]
d3.extent(config.reduced.top(Infinity)[1].value)
[0.0002859, 0.6982]
d3.extent(config.reduced.top(Infinity)[2].value)
[0.000007057, 0.5115]
$scope.chart.yAxisMax()
12.832
$scope.chart.yAxisMin()
-11.999992943
```

![image](https://cloud.githubusercontent.com/assets/1969229/13892766/4804d772-ed62-11e5-9b73-cb3eeaf08058.png)

Any way around this without having to override the domain manually - that is, to have to recompute the extent whenever the chart filter is triggered?

``` javascript
.elasticY(false)
.y(d3.scale.linear().domain([currentExtent]))
```

?

amergin:
My bad, it seems that this can be controlled by tweaking the setting `yAxisPadding`, i.e. set

```javascript
.yAxisPadding('10%')
```
produces sane figures. Although it's weird the default value would be so high considering the docs mention the default value to be `0`.

gordonwoodhull:
Yeah, we've run into this before. I think this chart was tuned for [its example](http://dc-js.github.io/dc.js/examples/box-plot.html) and it contains the suspicious lines

```js
    // default padding to handle min/max whisker text
    _chart.yAxisPadding(12);
```
https://github.com/dc-js/dc.js/blob/develop/src/box-plot.js#L54-L55

... which may work great for that example but it's not great to adjust the domain in order to fit some text in range coordinates.

I don't think we currently have range padding (?) but that's what's needed here. At least there is access to the thing you need to override.

stevenmyhre:
Another issue, if you have very small values as @amergin does, the `yAxisMin` value is calculated based on the minimum value, not taking into account the range.  When you use a percentage for `yAxisPadding`, this doesn't scale well.  For example: if your minimum value is 0.5, and your max is 2000, with a `yAxisPadding` of `5%` then you will end up with only 5% of 0.5 (.025) when I would expect it to calculate based on the target range (a percentage of pixels not values).  I was able to just override the `yAxisMin` function to use a percentage of the `max - min`, but thought I would give my 2 cents.

cwolcott:
[...]

PR #1439, opened 2018-05-14
Boxplot enhancements

cwolcott:
This is my first contribution to a github library.  I have demonstrated the boxplot enhancements in Issue [1370](https://github.com/dc-js/dc.js/issues/1370).  This enhancement should also fix [1120](https://github.com/dc-js/dc.js/issues/1120).

I am currently using 2.1.9 in our code baseline, but have submitted this pull request after branching from 2.1.10.  Once I can get some feed back, I will be happy to add additional test to the box-plot-spec and make sure the documentation for the new attributes are correct.

d3.box.js now has a cyclomatic complexity score of 18.  I assume I need to fix this.

Once this is integrated into the 2.1 branch, I will be happy to look at integrating it with 3.x


cwolcott:
I think that is it.  Please look at the 2 new examples:

- boxplot enhancements

- boxplot render data

When running **grunt lint --force** I still have a cyclomatic complexity of 18 on d3.box.js.  I assume you would like to me to see if I can refactor the code.

gordonwoodhull:
I'm surprised that a couple of new if statements could push the cyclomatic complexity that high. 

I am somewhat dubious of this metric: that code doesn't look all that hard to read.

I'd be okay with just disabling that warning for the file or function, as long as the lint is otherwise okay. You can see an example where we disable a jshint error at the top of`src/core.js`

cwolcott:
Here are my responses:

- the "magic number" in yAxisMin & yAxisMax functions (Images show numbers of 0, 8, 16, 96)
Forces a minimum pixel padding on the Y Axis.  I am changing the magic number to 8.
So if the magic number is 8, minimum data value is 1 and the yAxisPadding is 0 then the chart yAxisMin is 0.0095.  Without the magic number the box plots would be drawn as in chart 1.
magicnumber 0magicnumber 08magicnumber 16magicnumber 96
I will investigate how other charts might pad the data display a little bit.

The chart yAxisPadding attribute is a unit padding. So if the magic number is 0, minimum data value is 1 and the yAxisPadding is 11 then the chart yAxisMin is -10.

- The yAxisRangeRatio uses the (max data value - min data value) / chart height.  This is then used to multiple by 8 to determine the padding to use on the yAxis.  Previously the developer defaulted the yAxisPadding to 12 which caused the issue for #1120. This fixes that.

- I will resubmit another pull request and let you know when I think it is ready for review again.

gordonwoodhull:
You can push more commits to the same branch and they will be reflected in this ticket automatically. GitHub is amazing!

Let's open up that magic number as a parameter. What you are doing makes a lot of sense but magic numbers are always a recipe for trouble (as `12` was!)

I bet there is a better way to handle that `TODO` like maybe the data is already bound to something closer. At the very least you can always use `d3.select(x).datum()` rather than `x.__data__` - more idiomatic, safer. Have not had a chance to look for myself.

gordonwoodhull:
Another more sophisticated way to implement padding is have a pixel padding parameter, then use [`scale.invert()`](https://github.com/d3/d3-scale/blob/master/README.md#continuous_invert) (somehow, but I haven't thought it through completely) to determine the amount to add to the domain to allow that many pixels of padding in the range.

I know we have the same problem with the bar chart labels above the bars, but I think people just use the [margins](http://dc-js.github.io/dc.js/docs/html/dc.marginMixin.html) to work around it. Ugly, but better than manipulating the domain to fit something measured in pixels.

Just giving some background here, don't feel obligated to pick it up. Ideally this would be something at the coordinate grid mixin level. We have domain padding but we also need range padding.

cwolcott:
I have changed the parameter for getColor to your suggestion: d3.select(this).datum()

I just played around with invert, but it was always returns the value I passed in.
_chart.y().invert(1) returned 1, 100 returns 100.  So I will keep the current code for now since the chart looks good for any number range and continue to research.  Just to give you an idea of the current code (with the Magic Number set to 8).

minDataValue = 1; maxDataValue = 27; yAxisRangeRatio = 0.1238; padding = 0.9904;
minDataValue = 1; maxDataValue = 100; yAxisRangeRatio = 0.4714; padding = 3.7714;
minDataValue = 1; maxDataValue = 1000; yAxisRangeRatio = 4.7571; padding = 38.0571;

minDataValue = 1001; maxDataValue = 1027; yAxisRangeRatio = 0.1238; padding = 0.9904;

cwolcott:
I am done unless you have more suggestions.

gordonwoodhull:
I just looked at the demos for the first time. Excellent work!

I am changing a few names for clarity:
 * `renderData` -> `renderDataPoints` (not exactly the same functionality as [lineChart.renderDataPoints](http://127.0.0.1:8888/web/docs/html/dc.lineChart.html#renderDataPoints__anchor), but it could be upgraded to allow setting individual SVG attributes, in a backward-compatible way, in the same way that function was)
 * `dataBoxPercentage` -> `dataWidthPortion` (awkward but more descriptive I think)

The datum is passed as the first parameter, so `d3.select(this).datum()` is just `d`!

You are right that `scale.invert()` won't work in `yAxisMin` and `yAxisMax`... because the scale has not yet been initialized; these calculations are used to calculate the domain for the scale. What you are doing should be equivalent: adding some padding in pixel coordinates, specifically a bit more than half of the font height.

I'd really like to see this promoted to `coordinateGridMixin` someday. For now I am renaming `_internalRangePadding` and exposing `yRangePadding` as a getter/setter (no longer "magic"). This is necessary because the user can change the font using CSS ` .dc-chart .box text`. We do not currently try to measure font sizes anywhere, so there are these little knobs everywhere.

It wasn't necessary to define `renderTitle` because it is already a member of `baseMixin`.

gordonwoodhull:
I also restored the necessary changes for D3v4. This will be released in 3.0.4

gordonwoodhull:
Thanks @cwolcott!

cwolcott:
@gordonwoodhull will this also be included in a 2.1.x release?

I have not worked with dc.js 3.0.x line yet, but do you want me to develop a pull request for this or have you already done it?

gordonwoodhull:
I wasn't planning to do a 2.* release for this but I could without too much difficulty. 

I should still have the right commit at home since I made my changes before rebasing onto develop. I'll check tonight.

gordonwoodhull:
Okay, I've published 2.2.0!

(And started a support branch `support/2.x` for future requests of this nature.)

cwolcott:
Thanks @gordonwoodhull.  Just integrated 2.2.0 into all of our examples and our application.  I notified several other folks that 2.2.0 with the enhanced BoxPlots is available.  Hopefully we will upgrade to the 3.x line over the next month.

PR #1441, opened 2018-05-20
Readme updates - references to v3.0

kum-deepak:


gordonwoodhull:
Aha! I knew there was a third place I had to update, along with the main page and the Wiki.

Thanks!

gordonwoodhull:
I ended up copying the simpler CDN instructions [from the Wiki](https://github.com/dc-js/dc.js/wiki). Although CDNJS was great for us for a long time, I think unpkg is a better option now.

Thanks @kum-deepak! Merged for 3.0.4

PR #1444, opened 2018-05-21
Sunburst value accessor issue

### LINKED ISSUE #1440: sunburst chart doesn't render when key reduced to multiple values
I am using dc.js  v3.0.3

Sunburst chart works as expected when a key is reduced to a single value.
However when a key is reduced to multiple values and valueAccessor is used to specify one of them I see the following error.

screen shot 2018-05-18 at 10 33 17 am

this shouldn't be the case as all the values are 'defined'. Am I missing something here?
Both the cases are illustrated in the following jsfiddle
https://jsfiddle.net/brL1x26g/

kum-deepak:
Attempt to fix #1440.


kum-deepak:
I have cleaned up the example. 

@gordonwoodhull, the current example uses Reductio, which may be good. However it is possible to demonstrate the concept using multiple group sums similar to Stocks example. Will wait for your advice.

If we continue with Reductio, currently redictio is loaded directly from CDN, it might be better to use npm to install and Grunt to copy it to our js folder, like other similar cases. Again, will wait for your advice.

mukherjeea:
@gordonwoodhull the data I've used in the fiddle is from production. I will need to mask couple of attributes before it can be used as an example. Give me some time. I will update you guys.

kum-deepak:
@gordonwoodhull, thanks for your responses, it all makes sense.

It does intuitively sound to me that keyAccessor would have same issue, however, I am unable to come up with an example. So, while I can make the change, do not want to do it without producing the problem first 😄.

There is one more fundamental thing I have breaking my head on. Does using Capping actually help (or even work for) Sunburst charts. I guess it is there because the original code was picked from Pie chart. Sunburst charts create a hierarchy. If someone uses Capping - the hierarchy will be incorrect (as may nodes would get collapsed into 'Others' bucket before the computation). I am unable to think about correct behavior of Capping in this context. At this stage none of the examples use Capping. What is your opinion - should we drop Capping from Sunburst?

gordonwoodhull:
@mukherjeea, any luck producing a data set we could use for the test? Unfortunately we are testing based on the example you gave in #1440, so I can't currently merge this fix without merging that data.

mukherjeea:
@gordonwoodhull really sorry about the delay guys. wasn't able to revisit this. [This fiddle](https://jsfiddle.net/brL1x26g/21/) has a usable dataset. I've also updated the code that follows. Again sorry about the delay.
 
https://jsfiddle.net/brL1x26g/21/

mukherjeea:
[dc_example_masked.csv.zip](https://github.com/dc-js/dc.js/files/2183846/dc_example_masked.csv.zip)

the dataset for completeness.

gordonwoodhull:
Thanks @kum-deepak, @mukherjeea!

I merged the new data into the example and squashed, so the unmasked data is gone from the main branch (it will disappear entirely once @kum-deepak  deletes this branch).

Thanks for providing clear field names on the new dataset - with that info I was able to call the dataset  `ordered_returned.csv` - I like that much better than naming it after the feature it's demonstrating.

We can fix the `keyAccessor` if it turns out to be a problem in the future.

Merged for 3.0.6

kum-deepak:
Deleted my branch. Thanks!

Issue #539, opened 2014-03-13
support ordinal y scale for bubble/scatter plots


gordonwoodhull:
While it has special cases for an ordinal X scale, currently coordinateGridMixin assumes that the Y scale is linear.  This doesn't always makes sense for bubble or scatter plots.

Here is a fiddle sort of halfway hacking around the problem in `prepareYAxis`:
http://jsfiddle.net/gordonwoodhull/xZFx4/10/

the3ver:
At least for a scatter plot I was able to get a simple example running with both x and y scales ordinal:
http://jsfiddle.net/the3ver/Lf72gmmy/3/

Looks linke the `prepareYAxis` workaround is no longer necessary.

gordonwoodhull:
I see.. probably because `d3.scaleBand` is its own thing now, instead of having to call`d3.scale.ordinal.rangeBands`

gordonwoodhull:
This is fixed by using d3v4 in dc.js 3.0, except for the ticks undefined problem, which is fixed by #1448.

PR #1448, opened 2018-05-25
fix "TypeError: scale.ticks is undefined"

the3ver:
Fix "TypeError: scale.ticks is undefined" when setting .renderHorizontalGridLines(true) on charts with ordinal y scale, see #539


gordonwoodhull:
Thanks @the3ver. LMK your name if you want more than your handle in the AUTHORS file.

I added tests for scatter plot ordinal scales, so this is officially supported now.

Issue #1447, opened 2018-05-23
Add stroke attribute to line-chart dots


rmanzano-TE:
For line-charts we are given the following options for renderDataPoints: 
`{fillOpacity, strokeOpacity, radius}`
but strokeOpacity is unused since the rendered dots don't have a stroke attribute defined. Would it be possible to add that attribute? Adding:
`.attr('stroke', _chart.getColor)`
to dc.line-chart.drawDots() on line 394 seems to have the desired effect. Or is there a reason why that attribute is excluded?

gordonwoodhull:
I agree, that's odd. I can only guess that the person who contributed `renderDataPoints` was setting the stroke another way.

Are you sure that setting stroke as an attribute helps? Deepening the mystery, we have 

```css
.dc-chart circle.dot {
  stroke: none
}
```
[in dc.css](https://github.com/dc-js/dc.js/blob/b3ee0bbbaf48aa83fd973410334c4e7b03de021f/style/dc.scss#L254)

And [CSS trumps attributes](https://stackoverflow.com/a/36755226/676195), so changing the attribute on the circle element doesn't do anything in my testing.

I'm definitely willing to fix this, but I'd like to leave the default with no stroke somehow. Hmmm, kind of a mess.

rmanzano-TE:
Ah, yea. I forgot I had removed that CSS rule in a local override. Looking through the code, I can't find any other place where the stroke is being set. Maybe removing that CSS rule and setting the default stroke-opacity to 0?

gordonwoodhull:
Yes, that sounds backward compatible since the stroke opacity didn't have any effect without changing the CSS.

Would you have any interest in filing a PR for this?

rmanzano-TE:
Yep, I'm assuming this would not count as an API change so I should branch of master right?

rmanzano-TE:
Would it be possible to include this fix with v2? If so how would I go about doing that? I've tested the changes with our local copy of v2 but we want to avoid making local changes.

gordonwoodhull:
Sure, I'm supporting 2.* for a while, while people migrate to dc v3 and D3v4. I can do another 2.2.* release. 

Technically, you could base your PR on the `support/2.0` branch instead of master, but it doesn't make a big difference as I'm used to applying commits on other branches.

gordonwoodhull:
Fixed in 2.2.1 & 3.0.6

PR #1449, opened 2018-05-30
Allow stroke to be used for line-chart data points

rmanzano-TE:
remove stroke: none css rule for circle.dot, add stroke attribute to line-chart dataPoints and set default value for dataPoints.strokeOpacity to 0.0, update line-chart-spec test for default options

see [#1447](https://github.com/dc-js/dc.js/issues/1447)


gordonwoodhull:
Thanks @rmanzano-TE, merged for 3.0.6.

If you're online today, could you tell me your first name for the AUTHORS file?

gordonwoodhull:
Per special request, I'm also releasing 2.2.1 with this change. Sorry it took a while!

PR #1459, opened 2018-07-03
Grouped and Stacked Bars2

jaklub:
Updated @OivalfMarques grouped-stack-bar-chart #1356 to work with version 3 of dc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "dc",
-  "version": "3.0.4",
+  "version": "3.0.4-groupstack",
   "license": "Apache-2.0",
   "copyright": "2017",
   "description": "A multi-dimensional charting library built to work natively with crossfilter and rendered using d3.js ",
--- a/src/bar-chart.js
+++ b/src/bar-chart.js
@@ -42,6 +42,7 @@ dc.barChart = function (parent, chartGroup) {
 
 
     _chart.type = _type;
+    
     dc.override(_chart, 'rescale', function () {
         _chart._rescale();
         _barWidth = undefined;
@@ -93,9 +94,10 @@ dc.barChart = function (parent, chartGroup) {
 
 
     function getCharts() {
-        if (parent instanceof Object) {
-            if (parent.children() instanceof Array) {
-                return parent.children().filter(function (chart) {
+        if (dc.instanceOfChart(parent) && typeof parent.children === 'function') {
+            var children = parent.children();
+            if (children instanceof Array) {
+                return children.filter(function (chart) {
                     return chart.type === _type;
                 });
             }
--- a/src/bar-chart.js
+++ b/src/bar-chart.js
@@ -27,16 +27,21 @@
 dc.barChart = function (parent, chartGroup) {
     var MIN_BAR_WIDTH = 1;
     var DEFAULT_GAP_BETWEEN_BARS = 2;
+    var DEFAULT_GAP_BETWEEN_BAR_SERIES = 5;
     var LABEL_PADDING = 3;
+    var _type = "dc.BAR_CHART";
 
     var _chart = dc.stackMixin(dc.coordinateGridMixin({}));
 
     var _gap = DEFAULT_GAP_BETWEEN_BARS;
+    var _serieGap = DEFAULT_GAP_BETWEEN_BAR_SERIES;
     var _centerBar = false;
     var _alwaysUseRounding = false;
 
     var _barWidth;
 
+
+    _chart.type = _type;
     dc.override(_chart, 'rescale', function () {
         _chart._rescale();
         _barWidth = undefined;
@@ -46,7 +51,7 @@ dc.barChart = function (parent, chartGroup) {
     dc.override(_chart, 'render', function () {
         if (_chart.round() && _centerBar && !_alwaysUseRounding) {
             dc.logger.warn('By default, brush rounding is disabled if bars are centered. ' +
-                         'See dc.js bar chart API documentation for details.');
+                'See dc.js bar chart API documentation for details.');
         }
 
         return _chart._render();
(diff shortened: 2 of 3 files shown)


gordonwoodhull:
Thanks @jaklub! I have some ideas about a cleaner, easier-to-maintain implementation, but you are right that porting this to v3 is the first step.

For now, could you add an example of grouped+stacked bars? The grouped-bar example is very helpful but it would be great to see the whole thing in action.

gordonwoodhull:
Also, if you're up for it, could you also enable brushing and put another chart into the example? Let's start a discussion about how interaction should work. 

I tried enabling brushes in the original plunkr in the previous PR, and it looks like brushing does not work on the grouped&stacked chart:
* bars are not greyed out when they are outside the brush (until it is redrawn, e.g. when the other chart is filtered)
* filtering doesn't happen when brushing the grouped&stacked chart

https://plnkr.co/edit/3q8pK4NGHYw6h7Rck7t0?p=preview

It looks like you may have done some additional work in this area?

I think this can be made to work, although it would have to round the brush so that a whole group is selected at once, not just individual bars.

jaklub:
Hey @gordonwoodhull! I was using one of the earlier grouped barchart PRs, but I kind of wanted to use the new shiny dc v3 :). So I tried to cook up something that worked (at least for me). 

But I must admit, I mostly considered my own use cases when creating this PR. Didn't for instance try to make it work with both grouped and stacked bars in the same chart as in the plunkr of the previous PR. Is that something that you would like to be working?

Tried quickly with brushing also, but don't think it worked. I can look into it bit more. 

Added one more example though, with two grouped and filtering barcharts. 

And yes, I did work some on the select/deselect portion of the code. It kind of works for my use cases, but it dosen't feel clean. I did add a few lines of code that overrides the click event for instance, feels a bit like a hack.

I am a bit of a beginner with this library (and d3), so please point out my newbie mistakes :). And by the way thanks for all work with dc, its a great lib.

Frozenlock:
Unfortunately it doesn't work.

Previous version : 
image

This branch : 
image

In addition, the new branch causes my CPU to run wild.

jaklub:
@Frozenlock I've made a jsfiddle https://jsfiddle.net/jaklub/8dcsmbuk/ could you try to fork and modify to reproduce your problem?

Frozenlock:
Hmm... I get an error about `.bandwidth` when I try to switch the jsfiddle to a non-ordinal chart.

``` js
  .x(d3.scaleLinear().domain([0, 9]))
  .xUnits(function () {return 10;})
```
Did I forget something?

https://jsfiddle.net/hnaumer2/

jaklub:
@Frozenlock I've updated the grouped barchart to be able to handle non ordinal charts. Also found small mistake in the fiddle when using the non ordinal bar chart.
```javascript
dim = ndx.dimension(function(d) {return "Run " + d.Run;}),
```
=>
```javascript
dim = ndx.dimension(function(d) {return d.Run;}),
```
https://jsfiddle.net/jaklub/hnaumer2/5/

@gordonwoodhull: Played around with the brush, it seems to be working somewhat now. Added a new example with three charts, one with the brush on.

gordonwoodhull:
Hi @jaklub - awesome, thanks for following through and providing these fixes. The examples looks very good.

It strikes me that we might want to change the ordinal-grouped hover and click behavior. Since you always get the whole group rather than an individual bar, it seems to me that 
1. the hover highlight should affect all of the bars of the group, not just the individual one the mouse is over
2. possibly we should allow clicking in the area behind the bars, i.e. the click target is the group itself, not the bars.

These are just ideas for future work. 

My main concern with merging this remains the way the chart reaches into its parent the composite chart and its siblings, in order to decide how to draw. I feel that this could be clearer and easier to maintain, but I don't yet have a concrete plan.

Until then, thanks for bringing this PR into the new world and making it available for everyone to try out!

Frozenlock:
@jaklub Now it works as expected.

image


Thank you!

jaklub:
@gordonwoodhull - I Agree that would be nice, to have the groups behave more like a group when hovering and selecting. 

In my opinion, this implementation makes the "physical" grouping of the bars a bit warped from how you might perceive them as a developer or user. I mean each "bar group" contains one bar from each child in the composite. This approach might also make it more difficult to implement something like you are suggesting above (and implementing #1428 which also sounds like a great idea). And I guess this also forces the charts to reach out to its siblings. 

I was looking through some of the earlier grouped bar chart versions. Think I was using the @gazal-k branch from #1043 earlier myself. Do you think that would be a better approach? To use stacking and an extra property like groupBars? Would of course make it more challenging to implement grouped and stacked bars in the same chart (don't know how requested that feature is though).

Or do you have any other thoughts on how to implement this feature in an more maintainable way?

Would love to see grouped bar charts, in one form or another, eventually make its way to the master branch :).

gordonwoodhull:
One of the arguments from the previous PR that made me prefer this approach, is that you can switch between a composite of line charts and stacked bars with mostly the same code.

What's missing, I think, is a class that sits between the composite chart and the bar chart and says "you are part of a group". This class would figure out the offsets of the bars within their groups, and could also mediate the mouse events we're talking about here. Then the offset is just another property of the bar chart and hopefully it doesn't have to look at its siblings.

I'm glad you mentioned #1428 - that's exactly what I was thinking for "clicking on the group".

gordonwoodhull:
Hi @grugknuckle, I wasn't able to repro that with npm@6.1.0, node@10. When I use the above line in package.json it installs a version of dc.js with the header

```js
/*!
 *  dc 2.1.10-groupstack
 *  http://dc-js.github.io/dc.js/
 *  Copyright 2012-2016 Nick Zhu & the dc.js Developers
 *  https://github.com/dc-js/dc.js/blob/master/AUTHORS
```

which is what I expect.

grugknuckle:
Ok ... Thank you @gordonwoodhull . Apologies for polluting this PR with my troubleshooting problems.

grugknuckle:
Excellent !

I was going to ask about v3.0.5 compatibility but now I don't have to. Thanks

Frozenlock:
Would it be possible to update the grouped branch with the newest release?

Frozenlock:
Thank you very much!

dadokkio:
Hi all, I'm playing with 3.0.8-groupstack and it works very well. I've just a question. 

I have a dashboard with a lot of other charts and one of this is a piechart with the values that the groupchart is using for his x axis.

[Example](https://jsfiddle.net/u3pq5mgw/)
The two pie filters each other but only the second one (that is not using exp as dimension) works also for the groupedchart and viceversa.

It's that correct?

gordonwoodhull:
Hi @dadokkio, that has nothing to do with this char.t It's because [a group does not observe its own dimension's filters](https://github.com/crossfilter/crossfilter/wiki/Crossfilter-Gotchas#a-group-does-not-observe-its-dimensions-filters). If you want two charts to filter each other, they need to be on different dimensions.

Frozenlock:
Hello @gordonwoodhull,

Have you made a decision as to which implementation should be merged into master?
(#1459 or #1470)

gordonwoodhull:
I dislike both implementations for different reasons. I am currently of the opinion that stacking needs to be expanded to multiple dimensions, but I haven't tried it.

Sorry for the disappointment. Maybe someone should fork and maintain their own version of this.

PR #1463, opened 2018-07-07
Moved .exit before .enter, should fix #1460

### LINKED ISSUE #1460: [v2 -> v3] ScatterPlot no longer removes symbols
### V2 behavior
The scatter plot translates or removes the various symbols when we regroup the dimension.

peek 2018-07-04 18-44

-------

### V3 behavior
The scatter translates but **does not remove** the various symbols when we regroup the dimension.

peek 2018-07-04 18-45

kum-deepak:
Possible fix of #1460.

### DIFF
--- a/src/scatter-plot.js
+++ b/src/scatter-plot.js
@@ -81,6 +81,9 @@ dc.scatterPlot = function (parent, chartGroup) {
         var symbols = _chart.chartBodyG().selectAll('path.symbol')
             .data(_chart.data());
 
+        dc.transition(symbols.exit(), _chart.transitionDuration(), _chart.transitionDelay())
+            .attr('opacity', 0).remove();
+
         symbols = symbols
             .enter()
                 .append('path')
@@ -117,9 +120,6 @@ dc.scatterPlot = function (parent, chartGroup) {
             })
             .attr('transform', _locator)
             .attr('d', _symbol);
-
-        dc.transition(symbols.exit(), _chart.transitionDuration(), _chart.transitionDelay())
-            .attr('opacity', 0).remove();
     };
 
     function renderTitles (symbol, d) {


kum-deepak:
@Frozenlock please test it out, see if it resolves the issue.

Frozenlock:
Yup!

kum-deepak:
@gordonwoodhull for your feedback.

gordonwoodhull:
I'm willing to merge this, but as I asked in https://github.com/dc-js/dc.js/issues/1460#issuecomment-403054912, I would still like to know *why* this works.

If `selection.exit()` has no side effects, and `transition.remove()` is delayed by the `transitionDuration`, how come this has any effect? 

Deepening the mystery, Mike's [General Update Pattern](https://bl.ocks.org/mbostock/3808218) series of blocks have `.remove()` last up until the last example, where transitions are introduced. So I think there may be something about the asynchronous nature of transitions which make this necessary.

kum-deepak:
@gordonwoodhull it indeed is a mystery to me that how it works. If we consider the original code and see http://localhost:8888/web/examples/scatter-brushing.html, symbols do get removed. So, in general there is no issue in `.exit` and `.remove`.

@Frozenlock, I could not find a link to a fiddle to check the problem. Please share the link (preferably the filter one), I will dig deeper into it.

Frozenlock:
@kum-deepak My apologies, but the gif is from a pretty complex setup in Clojurescript.
In fact I don't program in javascript. I can read, tweak a few things, but that's about it.

The simplest way to test this is probably to have a button with which you can remove an entry in the chart's group. @gordonwoodhull will probably know more about what is going on, as he correctly guessed I was using fake groups.

gordonwoodhull:
Okay I'll try to test this. 

The scatter-brushing escape doesn't exhibit the problem because no bins are actually removed - the values just go to zero and the chart reacts by setting the size to zero based on the `existenceAccessor`.

This is what you get with straight crossfilter, which is probably why we didn't have any Jasmine tests for it.

gordonwoodhull:
Ah, okay, the answer is really trivial. I was totally overthinking it. 😊

It's just that using this coding pattern we replace the `symbols` variable. The result selection of `.merge()`, of course, has no exiting elements.

Think I understood this before and just forgot.

Issue #657, opened 2014-07-30
filter individual segments of stacked chart


NikhilChandrasenanNair:
Hi,

I have created a stacked bar chart.I want to filter data of other charts by clicking on individual segements of bar chart.
Below is the image of the charts. 

Here is the fiddle link    http://jsfiddle.net/Lihkinrian/J7TYD/

![capture](https://cloud.githubusercontent.com/assets/7201610/3746105/5d5a4f92-17b9-11e4-9e21-387f5525cd59.JPG)

gordonwoodhull:
You would have to do this with a renderlet right now.

gordonwoodhull:
Although it would be great to add this feature to dc.js core, it's not horribly difficult to implement it outside the library.

I've added [an example of filtering stacks](http://dc-js.github.io/dc.js/examples/filter-stacks.html) in v2.1.4.

The key is to key your dimension on both the x axis value and the stack key:

```js
          function multikey(x,y) {
              return x + 'x' + y;
          }
              runExptDim = ndx.dimension(function(d) { return multikey(d.Run, d.Expt); }),
```

Then build a conventional multivalue fake group from the multikey group. The keys are the x values, and the values are objects keyed on the stack key:

```js
          function stack_second(group) {
              return {
                  all: function() {
                      var all = group.all(),
                          m = {};
                      // build matrix from multikey/value pairs
                      all.forEach(function(kv) {
                          var ks = splitkey(kv.key);
                          m[ks[0]] = m[ks[0]] || {};
                          m[ks[0]][ks[1]] = kv.value;
                      });
                      // then produce multivalue key/value pairs
                      return Object.keys(m).map(function(k) {
                          return {key: k, value: m[k]};
                      });
                  }
              };
          }
              runExptGroup = runExptDim.group().reduceSum(function(d) {
                  return d.Speed;
              }),
              stackedGroup = stack_second(runExptGroup);
```

We'll disable built-in brushing and implement the click event ourselves:

```js
         chart
              .brushOn(false)

          chart.on('pretransition', function(chart) {
              chart.selectAll('rect.bar')
                  .classed('stack-deselected', function(d) {
                      // display stack faded if the chart has filters AND
                      // the current stack is not one of them
                      var key = multikey(d.x, d.layer);
                      return chart.filter() && chart.filters().indexOf(key)===-1;
                  })
                  .on('click', function(d) {
                      chart.filter(multikey(d.x, d.layer));
                      dc.redrawAll();
                  });

[...]

PR #1470, opened 2018-07-17
Grouped or Stacked bar chart with sensor bars

jaklub:
Hey @gordonwoodhull ! 

Played around some more with grouped bar charts (after #1459) and tried a version inspired by @gazal-k PR #1043. Also included a version of the sensor-bars as discussed in #1428. 

One "feature" of this PR is that you can easily switch between grouped and stacked bars.

One drawback is that this PR changes how the bars are "grouped/stacked" in the canvas, and their class names. Maybe something you would like to avoid?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "dc",
-  "version": "3.0.6",
+  "version": "3.0.6-grouped-or-stacked",
   "license": "Apache-2.0",
   "copyright": "2017",
   "description": "A multi-dimensional charting library built to work natively with crossfilter and rendered using d3.js ",
--- a/spec/bar-chart-spec.js
+++ b/spec/bar-chart-spec.js
@@ -77,17 +77,17 @@ describe('dc.barChart', function () {
             });
 
             it('should generate labels with positions corresponding to their data', function () {
-                expect(nthStack(0).nthLabel(0).attr('x')).toBeWithinDelta(405, 1);
-                expect(nthStack(0).nthLabel(0).attr('y')).toBeWithinDelta(104, 1);
-                expect(nthStack(0).nthLabel(0).text()).toBe('1');
+                expect(nthBarGroup(0).nthLabel(0).attr('x')).toBeWithinDelta(405, 1);
+                expect(nthBarGroup(0).nthLabel(0).attr('y')).toBeWithinDelta(104, 1);
+                expect(nthBarGroup(0).nthLabel(0).text()).toBe('1');
 
-                expect(nthStack(0).nthLabel(3).attr('x')).toBeWithinDelta(509, 1);
-                expect(nthStack(0).nthLabel(3).attr('y')).toBeWithinDelta(104, 1);
-                expect(nthStack(0).nthLabel(3).text()).toBe('1');
+                expect(nthBarGroup(3).nthLabel(0).attr('x')).toBeWithinDelta(509, 1);
+                expect(nthBarGroup(3).nthLabel(0).attr('y')).toBeWithinDelta(104, 1);
+                expect(nthBarGroup(3).nthLabel(0).text()).toBe('1');
 
-                expect(nthStack(0).nthLabel(5).attr('x')).toBeWithinDelta(620, 1);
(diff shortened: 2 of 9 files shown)


gordonwoodhull:
I see. So we currently have the choice to be able to switch between grouped and composite easily, or switch between grouped and stacked easily.

I'd really like to support grouped + stacked at the same time, so I somewhat prefer the other PR (#1459). But it sounds like there are good arguments either way. 

What deficiencies in the other approach caused you to move to this one?

jaklub:
Hey @gordonwoodhull !

Well, I started with this approach a litle bit as an exercise to get to know the library better. 

One thing that "confused" me with the other approach (at least while working on the functions that calculated the bar widths and x coordinates :)), was how the bars are "grouped" in the dom. 
It felt more intuitive that a "group" of bars would share a common parent element (see the screenshots bellow). 

To accomplish that in this PR, I might have made some violence on how you are suppose to handle the stacked data coming from d3. As in these rows in plotData(): https://github.com/jaklub/dc.js/blob/da8f8748f5fc2b8cb2e6baf963c366aa74ec7561/src/bar-chart.js#L78-L115
Might be a better way to do that type of data pivot, or another way to achieve the same end result.

But after the pivoting I thought it was easier to work with the placement of the bars, and the injection of the sensor bars also felt natural. Also the change in behavior when hoovering over and selecting a group felt like it was easier to implement when the bars where grouped in the dom.

At first I set up the toggling between stacked and grouped bars in the examples to ease the tweaking of the width and x calculations. But after a while it felt kind of natural. I mean in a way the two ways of presenting the data could be considered to be really closly related. And treating them similarly in the implementation and the dom felt like an advantage. 

But I agree one disadvantage of this approach is the support of grouped + stacked bars at the same time. That feature would probably be more difficult to implement.

Example of the dom in bar-grouped-stacked2:
bar-grouped-stacked2-dom

Example of the dom in bar-or-stacked:
bar-or-stacked-dom

gordonwoodhull:
Those are good points. I agree, to truly do it right, we want the DOM organized nicely and not just the API. I think that would be pretty difficult with the composite-based API, so I guess we would need something in the charts that understand more than 2D of data.

cwolcott:
Has this feature stalled out.  I have a need for the Grouped Bar Chart.  I understand how long it can take to complete a new feature.  I finally submitted the enhanced Box Plot a couple of months go, but actually started the enhancement 9 months prior.

gordonwoodhull:
Hi @cwolcott. I think the code is good. Unfortunately I stalled out from indecision between this approach and #1459.

The implementation here is much cleaner and IMO will be easier to maintain, and as argued above, it produces a cleaner DOM. However, this approach also pretty much shuts out the possibility of doing grouped+stacked bars for the foreseeable future.

Going either way is going to make people unhappy. I should probably pull the trigger and decide one way or the other, but it's going to cause me some pain too, either trying to maintain the other one or hurting my head trying to make grouped+stacked work with this one.

cwolcott:
Understand. I am actually looking for group+stacked.

gordonwoodhull:
There's a branch mentioned in the other ticket, which has a reasonably up-to-date version (3.0.5) of dc.js with grouped+stacked. If you're okay with the limitations of using a branch, it could be a solution for you.

Frozenlock:
I'll just chip in and describe how I'm currently using the other branch.
I split a timeseries into multiple different dimensions (say weekly) and compose them all into the same chart.

For example, each color represents a different week : 
image

Is this something that could be done with this PR?

gordonwoodhull:
Hi @Frozenlock, I think that scenario can work with either grouped bar implementation, at least the drawing part.

The stack feature just takes any group (no matter what dimension it belongs to), and grouped bars in this PR are implemented using the stack interface, so that should work.

The filtering behavior might be different, though. Currently there is no support for selecting individual segments of a stacked bar chart (#657), and that applies to the other implementation of grouped bars too. (It's really a data problem, since there is usually no dimension representing the stacked-ness). With this implementation, each bar in a group is in its own chart, so it is possible to filter it.

(I haven't verified this, just going on my general understanding of how stuff works.)

Issue #1474, opened 2018-08-19
controlsUseVisibility documentation is incorrect in Getting Started guide


ialarmedalien:
An easy documentation fix I spotted...

According to the [dc.js Getting Started guide](http://dc-js.github.io/dc.js/docs/stock.html#section-9):

> `dc.js` >=`2.1` uses `visibility: hidden` to hide/show controls without disrupting the layout. To return the old `display: none` behavior, set `chart.controlsUseVisibility(false)` and use that style instead. 

However, the API docs and the dc behaviour is the opposite: the default is to use `display: none` and `chart.controlsUseVisibility` has to be set to `true` to trigger the behaviour. Suggest changing the docs to

> By default, `dc.js` uses `display: none` to control whether or not chart controls are displayed. To use `visibility: hidden` to hide/show controls without disrupting the layout, set `chart.controlsUseVisibility(true)`.

gordonwoodhull:
Thanks! I think I originally thought I would just change the behavior, then went back and made it backward-compatible instead. But I forgot I had documented it in the Getting Started guide.

gordonwoodhull:
Thanks @ialarmedalien! fixed via #1475 in 3.0.7.

PR #1475, opened 2018-08-20
Filter printer does not correctly print out multi-element filters

ialarmedalien:
When using a hierarchical chart (the sunburst chart) where the filtered dimension is an array, the filter printer only prints out the first two members of the array. This alters the printer so every element of the array is printed out.

Apologies for the odd commit history; I accidentally reused an old branch.

Added a fix for [#1474](https://github.com/dc-js/dc.js/issues/1474), altering the documentation in `stock.js` for `chart.controlsUseVisibility`.

### DIFF
--- a/spec/utils-spec.js
+++ b/spec/utils-spec.js
@@ -46,6 +46,9 @@ describe('dc utils', function () {
         it('print zero', function () {
             expect(printer(0)).toEqual(0);
         });
+        it('print a multi-element array', function () {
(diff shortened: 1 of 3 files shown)


gordonwoodhull:
Makes a lot of sense. Will cherry-pick the relevant commits and merge soon!

gordonwoodhull:
Thanks @ialarmedalien! Merged for 3.0.7

PR #1477, opened 2018-08-31
add a single select example using addFilterHandler()

moononournation:


gordonwoodhull:
Thank you, this is one of those things that is totally not obvious, and the question comes up a lot.

I think I will just shorten the name to "bar single select" but this looks great otherwise! Will merge soon.

gordonwoodhull:
Ah, I see we had an example by that name before, but it was using a very difficult method to do the same thing. For 3.0.7 I'm replacing the old example with yours, which is much nicer. Thanks @moononournation!

Issue #1467, opened 2018-07-11
Sunburst : filtered elements should be animated


Frozenlock:
Filtered elements instantaneously disappear without animation.

See the [Sunburst example](https://dc-js.github.io/dc.js/examples/sunburst-cat.html).

Without animation, it's not immediately clear what is being removed from the chart.

The pie chart calls `dc.transition` which is missing for the sunburst chart.
Could it be as simple as that?

Pie chart : https://github.com/dc-js/dc.js/blob/develop/src/pie-chart.js#L124
Sunburst : https://github.com/dc-js/dc.js/blob/develop/src/sunburst-chart.js#L128

ialarmedalien:
There are various bugs in the code that are stopping the transitions from being performed—the main problem is that the existing code talks about `x`, `dx`, `y`, and `dy`, which are from the d3 v3 era, but v4/v5 uses `x0`, `x1`, `y0`, and `y1`. I am working on a branch that fixes those errors and adds some new functionality to the sunburst chart.

gordonwoodhull:
Thanks @ialarmedalien! 

Please let us know how it goes. Even a partial solution would be a big help.

Frozenlock:
Appears to be fixed in the last version.
Thank you very much!

gordonwoodhull:
Yes @ialarmedalien fixed it in #1481 for 3.0.7

PR #1481, opened 2018-09-02
Sunburst fixes

ialarmedalien:
This fixes [#1467](https://github.com/dc-js/dc.js/issues/1467), the absent transitions in the sunburst chart. Includes an edit to `safeArc` to reorder params and remove the unnecessary `i` argument.


tttp:
Hi,
Why did you upgrade grunt? is there something incompatible with the existing version?

ialarmedalien:
I ran `npm audit fix` after installing the various node modules as there were a number of security issues flagged; the updated `grunt` is probably a result of that. (Rerunning it reveals a further 27 security vulnerabilities, including six serious ones.)

gordonwoodhull:
This works very well! Thank you, I hope to merge this soon.

I find the history of this branch a little confusing, so I think I will just take the last commit (ccbf7c2), and do the npm updates separately. Seems to work!

Although it is important to stay up-to-date, we generally aren't exposing our users to any security vulnerabilities - these flaws invariably arise on the `devDependencies` side, so we may have insecure testing environments, but that's not a big deal. I don't think anyone is putting our `grunt server` into production. (I hope not!)

gordonwoodhull:
Thanks @ialarmedalien! This looks amazing, and really helps see what just happened when you filter.

Merged for 3.0.7

PR #1491, opened 2018-10-09
Fix Sunburst centering on redraw

### LINKED ISSUE #1490: Sunburst : doesn't recenter when redrawn
### Behavior

Expected : the sunburst chart should recenter itself like the pie-chart when redrawn width different width/height.

Actual : the sunburst chart doesn't move.

### Potential cause
The sunburst chart doesn't have a cx/cy based transition in the `drawChart` function like the pie-chart does.
Pie chart : https://github.com/dc-js/dc.js/blob/develop/src/pie-chart.js#L124
Sunburst chart : https://github.com/dc-js/dc.js/blob/develop/src/sunburst-chart.js#L94

Frozenlock:
Fixes #1490

### DIFF
--- a/src/base-mixin.js
+++ b/src/base-mixin.js
@@ -432,6 +432,8 @@ dc.baseMixin = function (_chart) {
         }
         if (dc.instanceOfChart(parent)) {
             _anchor = parent.anchor();
+            if(_anchor.children) // is _anchor a div?
+                _anchor = '#' + parent.anchorName();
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
I'm confused. It seems to get into the right position but not resize (change radius)?

gordonwoodhull:
Thanks @Frozenlock! Merged and released as 3.0.8

PR #1492, opened 2018-10-17
Example - Table with other widget

dadokkio:
Expanded table pagination example with interaction with other widget and filtered|total item count

based on: https://www.intothevoid.io/data-visualization/sortable-table-dc-js/

### DIFF
--- a/web/examples/index.html
+++ b/web/examples/index.html
@@ -78,6 +78,7 @@ here</a>.</p>
     <td><a href="table-pagination.html">table pagination</a></td>
   </tr>
   <tr>
+    <td><a href="table-with-other-widgets.html">table with other widgets</a></td>
     <td><a href="text-filter-widget.html">text filter widget</a></td>
     <td><a href="time-intervals.html">time intervals</a></td>
   </tr>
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
This looks like a nicer-styled version of the [table pagination](http://dc-js.github.io/dc.js/examples/table-pagination.html) example, with a pie chart filtering it.

Maybe we should just replace the earlier example?

BTW, I noticed an issue. If you select two categories from the donut chart, then go the last page of the table and deselect one of the categories, you end up with a blank table, even though it says it is displaying data:

image

Paging up and then down again, the three rows show up.

I am not sure if this was always a problem with pagination or if it is specific to this example. (I also haven't tried to troubleshoot it yet.)

gordonwoodhull:
I was able to fix this by splitting `update_offset` and `display`, calling `update_offset` on `preRedraw` and `display` on `pretransition`. @HamsterHuey's code keeps `beginSlice` in bounds but it wasn't getting called in the right order in this PR.

I'm replacing the existing table-pagination example in 3.0.9, on a release branch for now, available sometime soon.

Thanks @dadokkio!

Issue #1500, opened 2018-11-19
Render gridlines opacity values keeping changing


louMoxy:
Hi guys,  👋  

Bit of an odd issue, the opacity value of the grid lines (horizontal or vertical) appear to be random every time. Here is a fiddle of the issue: 

https://jsfiddle.net/vpk5cu9n/

Having a little dig through the code, I am assuming its because it hasn't finished its the transtion on enter, although I can't see an obvious issue with the code: https://dc-js.github.io/dc.js/docs/html/coordinate-grid-mixin.js.html#sunlight-1-line-723

louMoxy:
Thanks @gordonwoodhull for the detailed response, I have opened a PR following your instructions.

PR #1501, opened 2018-11-24
Fix missing animation for grid-lines opacity

louMoxy:
* Remove opacity from gridlines css (this was overriding the opacity animation on render)
* Update grid-chart-spec to test that the opacity value after render is 0.5
* Update gridMixin to display the gridlines with opacity 0.5 and remove overriding transition

Fixes: #1500


gordonwoodhull:
Ah, having them fade in looks really nice. Thanks @louMoxy, and double thanks for the tests!

Merged for 3.0.9.

Issue #1506, opened 2018-11-30
Reimagining crossfilter and dc.js interactions


kum-deepak:
@gordonwoodhull many thanks for your continuous work on making dc.js available to the humanity. 👍 

I had a look at https://github.com/att/xfilter. It pretty much achieves what I needed.

I have few ideas to make it ready for the prime time:

- Currently dc code is highly tailored for what [Crossfilter](https://github.com/crossfilter/crossfilter) provides and their API. `dc` does not need much, which you have already almost separated in [xfilter](https://github.com/att/xfilter). I think to make xfilter (or anything else) need to be brought to same playing field as crossfilter.
- We should define what dc actually needs - including things cross filter does not provide (like anything that needs a function to be passed as filter). Then we can create an adapter to bridge between crossfilter and dc that brings in these additional functionality as well.
- This will also imply that some of filter related code will shift to adapters.
- At this stage `xfilter` (and others) will be able to target the same specs.
- I believe it can be done in a backward compatible way.

I addition there can be some minor improvements which I can describe as we move along. At this stage wanted to check your views.

kum-deepak:
Thanks for your detailed note. I will get some code going, that will give me ideas as well 😄

kum-deepak:
This task is completed across several PRs #1765, #1773, #1774 and #1779.

PR #1510, opened 2018-12-12
Rethinking filters - data adapters

kum-deepak:
As a first step I have extracted filter related code and attached it with dimension. The current code is similar in logic as the previous code.

- Ideally we should offer a wrapper onto Crossfilter so that returned dimension objects are already blessed with the extra functionality.
- The special filters (from dc.filters) to be converted to plain JS objects with all the info. The specialized code to actually filter should move to the data adapter.
- Not tried to run fix test cases yet - once we agree on the basic outline I will get onto those.
- Range/focus - does not fully work - will look once we agree on basic structure.

See #1506

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -529,6 +529,7 @@ module.exports.jsFiles = [
     'src/logger.js',
     'src/config.js',
     'src/events.js',
+    'src/crossfilter-adapter.js',
     'src/filters.js',
     'src/base-mixin.js',
     'src/margin-mixin.js',
--- a/src/base-mixin.js
+++ b/src/base-mixin.js
@@ -1,84 +1,3 @@
-dc.wrapDimension = function (_dimension) {
-    var _filters = [];
-
-    var applyFilters = function () {
-        if (_filters.length === 0) {
-            _dimension.filter(null);
-        } else if (_filters.length === 1 && !_filters[0].isFiltered) {
-            // single value and not a function-based filter
-            _dimension.filterExact(_filters[0]);
-        } else if (_filters.length === 1 && _filters[0].filterType === 'RangedFilter') {
-            // single range-based filter
-            _dimension.filterRange(_filters[0]);
-        } else {
-            _dimension.filterFunction(function (d) {
-                for (var i = 0; i < _filters.length; i++) {
-                    var filter = _filters[i];
-                    if (filter.isFiltered && filter.isFiltered(d)) {
-                        return true;
-                    } else if (filter <= d && filter >= d) {
-                        return true;
-                    }
-                }
-                return false;
-            });
-        }
-    };
-
-    var removeFilter = function (filter) {
-        for (var i = 0; i < _filters.length; i++) {
-            if (_filters[i] <= filter && _filters[i] >= filter) {
-                _filters.splice(i, 1);
-                break;
-            }
-        }
-    };
-
-    var addFilter = function (filter) {
-        _filters.push(filter);
-    };
-
-    _dimension.hasFilter = function (filter) {
-        if (filter === null || typeof(filter) === 'undefined') {
-            return _filters.length > 0;
-        }
-        return _filters.some(function (f) {
-            return filter <= f && filter >= f;
-        });
-    };
-
-    _dimension.exFilter = function (filter) {
-        if (!arguments.length) {
-            return _filters.length > 0 ? _filters[0] : null;
-        }
-        if (filter instanceof Array && filter[0] instanceof Array && !filter.isFiltered) {
-            // toggle each filter
-            filter[0].forEach(function (f) {
-                if (_dimension.hasFilter(f)) {
-                    removeFilter(f);
-                } else {
-                    addFilter(f);
-                }
-            });
-        } else if (filter === null) {
-            _filters = [];
-        } else {
-            if (_dimension.hasFilter(filter)) {
-                removeFilter(filter);
-            } else {
-                addFilter(filter);
-            }
-        }
-        applyFilters();
-    };
-
-    _dimension.filters = function () {
-        return _filters;
-    };
-
-    return _dimension;
-};
-
 /**
  * `dc.baseMixin` is an abstract functional object representing a basic `dc` chart object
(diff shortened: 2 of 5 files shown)


gordonwoodhull:
The idea makes sense and will help clean things up and connect dc.js with other backends.

However, this implementation isn't backward compatible (and a lot of tests are failing). There are one or two places where we replace the value in the setter, and return something different in the getter from what was set, but it's generally kind of confusing and should be avoided.

Rather than wrapping `chart.dimension()` in another object with a different interface, maybe we could have some sort of "policy" or "strategy" object which provide methods which take the dimension and perform these actions? That way the dimension is left unchanged, but the user could replace this policy with one that does different things.

The charts would still not directly access the dimension, so it still serves the purpose of isolating dc.js from crossfilter. I guess for backends which don't have dimensions per se, the user could pass a dummy value, or something that is meaningful to the backend (e.g. the name of a column).

kum-deepak:
My thinking pattern is slightly different. I was intending to clearly define an abstract interface that dc needs from a data provider. Currently dc uses two interfaces - dimension and group (with exception to dataCount widget). We can determine if we need to additionally abstract the main data provider itself (corresponding to current crossfilter).

We can enumerate methods for `dimension` (seems `hasFilter`, `filter`, `bottom`, and `filters`) and `group` (seems `all`) that dc actually needs.

The methods `filter` and `hasFilter` have been overloaded quite peculiarly. These can be split into multiple simpler methods. The `chart.filter` method, in some cases, does not return what was set. We also need to think about complex filters (for example from dc.filters).

Backward compatibility can be maintained for most of the cases. My current assessment suggests that supporting all the Handler methods may lead to quite clumsy code. Instead we should provide a guide advising how to migrate such cases.

Currently I am not focusing on test cases. Once we agree on an outline, I will work on those. I have been testing using the stock example on the home page.

gordonwoodhull:
Yes, I agree with all of that, I think.

I'm just emphasizing the need to maintain backward compatibility. All existing dc.js user code should be unaffected by the change. Users should not need to wrap crossfilter, dimension or group in anything, and what they get back from any getters should be a normal dimension or group.

I think I would be happy with any change that fulfills that requirement. I was suggesting one way to do it, which is for the adaptor/policy to take crossfilter objects as arguments to its methods. I'm sure there are other ways to do it.

gordonwoodhull:
Basically, every time dc.js currently accesses a crossfilter object, it can do it through the adaptor ([or adapter](https://grammarist.com/spelling/adapter-adaptor/)). But the user does not see the objects as wrapped, and the user can specify a different adaptor.  

If you want to go ahead and build this according to a different plan, that's ok, but it just means that I will be the bottleneck when we finally try to merge this (unless you want to maintain a fork of dc.js). I'd rather spend a little time with you now working out an interface we agree on, rather than winding up with something that is really difficult to merge later.

IMO there is no need for the adaptor to own any crossfilter objects. It can be purely functional.

gordonwoodhull:
Also IMO any issues with the `filter` and `hasFilter` interfaces are orthogonal to this discussion. First, isolate crossfilter without changing any interfaces. Then, worry about any stupid designs in dc.js later.

Most of the handlers manipulate lists of filters, which I don't think will change. Then there is `filterHandler` itself, which currently expects to take a crossfilter dimension and the current list of filters. We can have our default implementation use the adaptor, but maintain compatibility with user code which works directly with the crossfilter dimension.

I know, it's annoying to maintain compatibility with something that was not always thought out. I often dream of a <strike>3.0</strike> 4.0 with everything nicely thought out. But if we want to bring our current users along, it has to be considered.

gordonwoodhull:
I had another look at this today, and unless I'm mistaken, I think you are pulling out three unrelated (but all complicated) pieces of functionality:

 1. the default `filterHandler`, which has some complicated logic applying filters in different ways depending on the type of filter
 2. the has/add/remove/reset filter handlers
 3. the implementation of `filter`, which applies filters by toggling them

I agree that Feature 1 has to be separated in order to use different backends from crossfilter. In a sense, it already is separate because it can be replaced, for instance with the current [xfilter.filter_handler](https://github.com/att/xfilter/blob/6ec8f299f4fe4258b3af14421150a3cb2887fa97/src/filter-handler.js). The main issue with that implementation is it does not cover all filter types, but it could be expanded easily.

Features 2 and 3 just manage the list of filters, and although they could be improved, they already do this in a fairly generic way. I don't see how it benefits a backend to abstract this part.

In my view, any backend should apply a list of filters and, either synchronously or asynchronously, it should provide new results through some `group.all()` (and `dimension.top()` / `.bottom()` if dimensions are used with `dc.dataTable`; `dimension.size()` for `dc.dataCount`). 

It's dc's responsibility to manipulate the list of filters, and the user can customize how that's done using the has/add/remove/reset filter handlers. 

It's maybe unfortunate that dc's standard filter behavior is to toggle filters, but that basically works and there is always a simple (if not intuitive) way to customize it.

I just looked and dc.js does not set filters on a dimension anywhere except for the `filterHandler` - so as long as a backend can apply a list of dc.js filters it should be all set. Except for this one handler, dc.js does not ever *set* any crossfilter object - everything else is *reads*, as described above.

@kum-deepak, could you please explain if you still think that the "filter list manipulations" - Features 2 & 3 above - need to be abstracted out? My belief is that the current interface, as messy as it may be, covers all the needed functionality here.

gordonwoodhull:
I'm closing this due to the reasons above. 

@kum-deepak, if you really think this will help, please ping me and I will reconsider.

kum-deepak:
Sorry for a late comment. Was off for about a month. I agree with your reasoning. 👍 

Please let me know when you think it is good time to start work on dc v4. I have time free this year, so, if it happens this year, I will be honored to contribute.

gordonwoodhull:
Thanks @kum-deepak! I'd really like to continue merging in some of the trickier PRs.

None of them are easy though. Some "just" need tests, which is difficult enough. Some have design issues. And others just need a thorough review and/or port to d3v5, which shouldn't be all that hard but still takes a lot of time.

Just today somebody asked after a "pyramid" / "paired row" chart. This is a case where we actually have two good contributions, but the design is quite different. I was able to rebase the branch but haven't delved into the review.

Sort of like grouped & stacked bars, but probably not as difficult.

PR #1524, opened 2019-03-25
Add a custom defined time interval to switching-time-intervals example

christianMurschall:
This PR adds a custom defined time interval (quarter of a year) to the `switching-time-intervals` example.


web/examples/switching-time-intervals.html
```diff
@@ -39,6 +39,41 @@
...
+            // and returns the number of boundaries between the start (exclusive) and end (inclusive).
+            function (start, end) {
+                return (end.getMonth() - start.getMonth()) / 3 + (end.getFullYear() - start.getFullYear()) * 4;;
+            },
+            // The optional field function takes a date, already floored to the current interval, and returns the
+            // field value of the specified date, corresponding to the number of boundaries between this date
+            // (exclusive) and the latest previous parent boundary. For example, for the d3.timeDay interval,
+            // this returns the number of days since the start of the month.
+            function (date) {
+                return date.getMonth() * 3;
```
gordonwoodhull: My reading is that this should return `Math.floor(date.getMonth() / 3)` for correctness - do you agree?

(dc.js won't use the field function, so it's sort of academic, but still would be nice to get it right in case anyone uses the code for other purposes.)

gordonwoodhull:
Thanks @christianMurschall, this looks great. I noted one potential problem which won't affect the example but everything else looks correct.

I'll merge this next release.

christianMurschall:
Thanks for your review @gordonwoodhull . I misunderstood the field functions purpose and have changed the PR accordingly.

gordonwoodhull:
Thanks @christianMurschall! Merged for 3.1

PR #1529, opened 2019-06-13
Updated resource urls, fixed mixed content loading

ron4stoppable:
resolves issue with mixed content loading, http content over https; in example

### DIFF
--- a/web/examples/replacing-data.html
+++ b/web/examples/replacing-data.html
@@ -8,21 +8,21 @@
 </head>
 
 <body>
-<link rel="stylesheet" type="text/css" href="http://dc-js.github.io/dc.js/css/dc.css"/>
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Thank you @ron4stoppable. I see the error in Firefox but not Chrome, and it prevents the page from running.

Wonder why it's not an issue with the rest of the examples, which load dc.js over http in header.js. Maybe because the same script is also loaded normally. I'm seeing a [warning about document.write](https://developers.google.com/web/updates/2016/08/removing-document-write) so we should at least remove the scripts from the `document.write`.

ron4stoppable:
@gordonwoodhull I got the error on chrome (73.0.3683.86), might be because of poor network. I'll check `document,write` in other examples too.

gordonwoodhull:
Thanks @ron4stoppable, merged for 3.1 (also fixed the one in header.js). 

I was able to repro on Chrome as well.

PR #1534, opened 2019-07-01
Remove console.log on mousemove and mouseout for line-chart dots

adrm:
I suppose this has been a slip and the library should not log this by default on production.


gordonwoodhull:
Thanks @adrm! Indeed, looks like a slip.

adrm:
Thanks to you!

Issue #1521, opened 2019-03-19
the way we save the stack in errors defeats the stack


gordonwoodhull:
I did this in bbee2c435:

https://github.com/dc-js/dc.js/blob/81082815b830ffc442fc8b57cc2cc7085db7da63/src/errors.js#L11

I presume it helped in some old browser, but now it replaces the usual stack with a stringy stack that doesn't work well in debuggers.

gordonwoodhull:
I'm not even seeing an error in the browser console in Firefox, so I wonder if throwing objects derived from Error is a good idea.

gordonwoodhull:
There are no try-catch blocks in the code at all, so for better or worse, removing the classes would only potentially affect clients of the library.

A brief search across repos on github did not reveal anyone using `InvalidStateException` but I guess the rest should be searched too... if that's the right approach. Maybe it's possible to fix the current derivation.

kum-deepak:
It has already been updated. This issue can be closed.

gordonwoodhull:
fixed in #1548

Issue #1520, opened 2019-03-13
dc.js 4.0: d3 6.0 and ES6


gordonwoodhull:
There has been a lot of interest in the dc.js community for an ES6 version, but it was not a priority to me as long as we had to transpile and run on IE. (*)

However, d3 6.0 will finally drop IE - see d3/d3-array#87. Glory day! We can use ES6 natively!

I'd be happy to work with someone or a team of people who want to port dc.js to Rollup, **NO BABEL or transpilation**, and ES6 modules. I think we should keep the code in one repo but have lots of modules in it. ES6 modules should allow anyone to mix and match their own dc.js charts or versions of dc.js charts with those in the official repo, and only take what they need.

I don't have the time to do all this work myself, but I have the time to do code review, make decisions and commit PRs for this project.

#### Challenges:
1. The dependency between modules is complicated. I think we will save ongoing effort by staying in one repo with one version number, but we still need to figure out the dependencies beyond [the class hierarchy](https://dc-js.github.io/dc.js/docs/html/). I'll bet there are many places where the dependencies are not pure, yet we want tree-shaking to work, etc.
2. Currently dc.js does not use prototype inheritance, so the path to ES6 classes is not completely clear. I doubt it will lead to significant performance improvements, but it would be nice to use idiomatic ES6 classes. There may be things we are doing, like adding and changing methods on other classes, which will not work or will cause subtle bugs.
3. As much as possible, we should be able to keep code written for DCv3 working with v4, but where ES6 helps the dc.js API, we should use it. ES6 classes may break the interface a little. I don't know if we'll have trouble with the places where people replace methods like `yAxisMin` in order to get the right effect. 

We can't fix everything at once - let's try to leave drastic interface changes for later, wherever possible.

I bet a lot of people out there have already gone through the ES5->ES6 transformation and know the best practices. Anyone have time to take this up? Anyone have experience with Rollup the way it is used to bundle D3? 

Please leave a comment if you're interested in working on this: doing the translation, review PRs, help debug, or just offer advice.

I know some people worked with @mtrayham's fork #1175. I closed the PR because I didn't want to transpile & still don't want to. I also want to use rollup instead of webpack because it is simpler. Nonetheless it's the right direction - did anyone try to merge that fork with the D3v4/5 changes in 3.x?

[...]

PR #1548, opened 2019-08-12
Clean up Exceptions

kum-deepak:
Ref #1520 and #1521


gordonwoodhull:
Great! Much cleaner. 

Thanks @kum-deepak, rebased and merged!

PR #1549, opened 2019-08-12
Remove polyfill code fragments

kum-deepak:
In ES6 modules it is not desirable to keep code that causes side effects during load. This implies that polyfill code should not be inside a module.

The current polyfills only target IE browsers, so, these can be removed. If someone does want to use this with IE, we can put documentation in how they can add required polyfills.


gordonwoodhull:
Yes! Great!

No we will not need to document them. We will be using arrow functions and other features that cannot be made to work on IE.
 
If someone else wants to transpile for IE that'll have to be a separate project. The only supported way will be to stick with 3.x

Issue #1543, opened 2019-08-11
ES6 - assignments to exported members


kum-deepak:
Ref #1520 

Rollup generated js does not allow reassigning any exported members. Currently test cases fail that use reassigning:

- disableTransitions (one spec)
- redrawAll - two specs put a spy on this call (which try to replace this function)

gordonwoodhull:
There will probably be a long tail of things that no longer work because they rely on reassignment.

There are a few global settings on the dc namespace. We might have to make these into global setters. 

Not sure what to do about spies.

kum-deepak:
My thinking is not firmed yet. However my current thinking is like the following:

- Currently there are few settings which are somewhat global
- Package all these settings into a class - say Config or DcConfig
- Attach an instance of this object (with default values) to the Chart Group
- A chart will always know which chart group it is linked to
- The config can be assigned to a chart group or to a chart
- If no config has been assigned to a chart (which is likely to be the case), it will use config from the chart group

However before this we will need to discuss how much refactoring and breakage of existing API is acceptable.

kum-deepak:
As of now commented out these 3 specs via https://github.com/dc-js/dc.js/pull/1540/commits/b741b1a44afd1397de0cca53641b647be7064ac1

kum-deepak:
I have been reading about ES6 and tree shaking. It seems a proper structure would be like following:

- No top level exported API should manipulate state or depend on state.
- No mutable state should be a top level exported symbol.
- Mutable state should be part of an Object (preferably of a defined Class). If not possible than wrapped inside a class as static members (e.g., `utils.uniqueId`).
- All stateless helper methods can become top level exports (e.g., `utils.arraysEqual`)
- Some larger, limited use, methods can go to their own modules (e.g., `utils.toHierarchy`)

gordonwoodhull:
Seems like there needs to be a third level: look for config on chart then chart group then on `dc` itself, as demonstrated in #1553. 

Or does this violate "No mutable state should be a top level exported symbol"?

I hope not, since it would be handy. I guess if all else fails we could make `dc.config` immutable and only contain defaults, but that's not as convenient for users.

gordonwoodhull:
Probably would not want to cascade individual settings, just look for a config object at each level, which will have some value for every setting.

kum-deepak:
Makes sense.

kum-deepak:
I think we can close this task for this version. More changes than the current state would cause interface breakage.

[...]

PR #1553, opened 2019-08-13
Refactoing

kum-deepak:
Please treat this as sample for discussion. These commits include the following:

- Refactoring real core stuff to `core` folder
- Create a `Config` class
- Move `dateFormat` to `Config`

Will proceed further based on feedback.


gordonwoodhull:
Nice to see the new directory structure unfolding. The `Config` class looks good to me; glad to see the infrastructure you built for ordinal colors proving itself!

In #1543 you suggest putting the config on the chart or chart group. Seems like it will also be necessary to have one on `dc` itself, in case neither of the others is set.

This looks like a good path to get there.

Only thing I will continue to ask: is there any way to make setting `dc.dateFormat` an error so that users are given a nudge in the right direction? Currently they will set it and it will be ignored, leading to incorrect labels which might not be noticed for a while.

kum-deepak:
I am still thinking heavily. Not very sure what the final structure would look like. I am targeting to have least hidden dependencies across modules. So far whatever I understood that if someone uses the bundled code they would get a `dc` object (a namespace) in their browser. However if they choose to directly use modules (e.g. if they use rollup/webpack) to generate bundle as part of their build they may not see a `dc` object.

It is already too much for me to get my head around. I am hoping to figure out more as we progress.

I will now attempt a rough conversion of one actual chart :smile:

gordonwoodhull:
I think your understanding is correct. I've started using functions from d3 modules in a (react / parcel) project, and that's how they have it set up. I hope to follow their lead as much as possible (without having a ton of repos or separate versions, which seem to create a lot of work).

There is also the usage 

```js
Import * as d3 from 'd3-scale';
```

Which ends up looking like the old bundle but only incidentally.

PR #1555, opened 2019-08-15
Independent Karma

kum-deepak:
I use IntelliJ RubyMine with node, nom, and karma add-ons. Keeping a karma config in a file of its own allows run and debug any test case from the IDE, without interacting with the browser. A big speedup during development.


gordonwoodhull:
Sounds neat. I still use DevTools for everything but this shouldn't interfere with that.

PR #1557, opened 2019-08-15
Remove js files list - no longer need to be manually maintained

kum-deepak:


gordonwoodhull:
It's still used by `docs/coverage.js` but that hasn't worked in a long time and didn't work very well.

It was supposed to tell how well do docs cover the API. Please remove it as part of this PR. 

We can always look it up in history if we want to try that again.

gordonwoodhull:
Otherwise, this is great! Hated updating that list.

kum-deepak:
I will create a new issue for docs-coverage. Once the ES6 conversion is proper, we might be able to look for some newer tool that accomplices docs-coverage.

gordonwoodhull:
Thanks @kum-deepak!

Issue #1559, opened 2019-08-16
Should crossfilter.quicksort be replaced by Array.sort


kum-deepak:
It seems `crossfilter.quicksort` is the only function used by dc itself from `crossfilter`. If this can be replaced by `Array.sort` (or something similar), `crossfilter` can become a dev dependency.

gordonwoodhull:
Yes, in fact we recently had a discussion about this on the crossfilter repo. 

It shouldn't be using a hand rolled sort either: crossfilter/crossfilter#1559

gordonwoodhull:
Let's do this in 3.1.x as well as 4.0. I can backport if you fix it in the es6 branch.

kum-deepak:
I have created a pull request. The change should work without ES6 as well.

gordonwoodhull:
fixed by #1560

PR #1560, opened 2019-08-17
 Replace crossfilter.quicksort.by with Array.sort 

kum-deepak:
Ref #1559 

I do not think any new test case is needed as it seems to be covered by existing test cases.


gordonwoodhull:
Great! I'll backport soon and release as 3.1.2.

gordonwoodhull:
released in 3.1.2, with a fix to the documentation to reflect the change

PR #1578, opened 2019-09-18
added background drawing example

sven-h:
Based on a [stackoverflow question](https://stackoverflow.com/questions/57748011/rendering-in-the-background-of-a-dc-js-chart-with-renderlet)  I added an example which shows how to draw in the background of a chart. I manually edited the index.html in the web/examples folder to include the example (hopefully that was correct). Just let me know in case something should be changed.

### DIFF
--- a/web/examples/background-drawing.html
+++ b/web/examples/background-drawing.html
@@ -1,0 +1,102 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>dc.js - Background Drawing Example</title>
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Awesome! Thanks so much @sven-h! I love this example.

`index.html` is generated automatically by `grunt docs` (specifically `grunt fileindex`), and normally doesn't need to be checked in for PRs, but it looks like you did exactly what it would have done.

I merged this to develop and it will be included in 3.1.4.

PR #1581, opened 2019-09-21
Clean up `const chart = this`

kum-deepak:
Removed `const chart = this`, rest all seem to be needed, at least for now.


gordonwoodhull:
Thanks @kum-deepak!

PR #1582, opened 2019-09-22
Cleanup .call/.apply

kum-deepak:
I have been reviewing all usage of `.call` and `.apply`. Majority of such calls are to invoke some d3 method which expect specific value of `this`.

There are some cases within dc code as well. Majority of these set `this` to a d3 element. In general I am finding code difficult to understand and maintain with .call/.apply.

I have restructured a function in https://github.com/dc-js/dc.js/commit/fb3f9db1cd55a8ce522364bad6c817722f3403b0. In this now the d3-element is passed a parameter and `this` is set to the chart. This makes the function very similar to other functions.

I think the same restructuring will help in cleaning up getColor in color-mixin.

This is for your review and discussion, then I will convert similar cases in other charts.

### DIFF
--- a/docs/dc-v4-upgrade-guide.md
+++ b/docs/dc-v4-upgrade-guide.md
@@ -38,7 +38,7 @@ for your case.
 - The previous was of instantiating a chart is still supported.
   However new way is recommended. For example:
   
-    - `dc.pieChart = (parent, chartGroup)` --> `new dc.PieChart(parent, chartGroup)`
+    - `dc.pieChart(parent, chartGroup)` --> `new dc.PieChart(parent, chartGroup)`
 
 - In dcv4, inside a `dc` chart functions set and expect `this` to be the chart
   instance. However `d3` sets it to the d3 element.
--- a/src/base/base-mixin.js
+++ b/src/base/base-mixin.js
@@ -351,7 +351,7 @@ export class BaseMixin {
      */
     data (callback) {
         if (!arguments.length) {
-            return this._data.call(this, this._group);
+            return this._data(this._group);
         }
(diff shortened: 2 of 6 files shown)


gordonwoodhull:
This makes a lot of sense in the`cboxMenu`, since 
 * the function can use `this` where it used to use `_chart`
 * the function is completely private and changing it won't break the public interfacs
 
You'll find the same function in `selectMenu` since cbox was based on select.

Of course the `baseMixin` case is very clear too; that falls out of storing private members on the chart object instead of the closure.

I wonder if the other cases are as clear-cut. `getColor` is definitely more complicated, since it is called both with the chart and the *layer* as `this` (the element as well?) And the bar chart and line chart have different layer objects IIRC.

I *think* it was just a lazy shortcut someone took, but it's going to be tricky to prove it. We can't rely 100% on the tests for that one, since it could silently break a color that isn't tested. 

And since it's public, it needs to be documented in an upgrade manual. Presumably it affects `colorCalculator` as well.

I completely agree that using `this` for anything other than the chart is going to be confusing in this new ES6 world. Yet we do have the D3 precedent to contend with, and it doesn't look like that will change: https://github.com/d3/d3/issues/2246#issuecomment-158186135

So I'm cautiously optimistic about this approach and I'd like to review the other cases.

kum-deepak:
I have not touched getColor so far. Other than that I have cleaned up. Please review.

Merge it if you find it is sufficient for now. For getColor I will create a separate PR.

gordonwoodhull:
Completely agree - all of these free functions are much improved as member functions.

Thank you for your careful, conscientious work, @kum-deepak!

PR #1583, opened 2019-09-22
Cleanup and refactor core

kum-deepak:
I am targeting to cleanup and refactor core in coming days. I will let you when this is complete and ready to merge.


gordonwoodhull:
Wow I didn't know about `dc.round` - I don't see the utility of wrapping these functions & agree it should go. Careful, it is still used in some of the examples, like the stock example.

kum-deepak:
Thanks for the stock.js tip, updated the commit (it was used in one more example).

gordonwoodhull:
Awesome. Have you started a migration guide? Although this is trivial it is part of the public interface...

kum-deepak:
I am keeping a note of changes, high time to start writing migration guide :+1:

kum-deepak:
Added first version of the upgrade guide. Please check if I missed anything.

docs/dc-v4-upgrade-guide.md
```diff
@@ -0,0 +1,62 @@
...
+    - `dc.dateFormat` --> `dc.config.dateFormat`
+    - `dc.disableTransitions` --> `dc.config.disableTransitions`
+    - `dc.errors.InvalidStateException` --> `dc.InvalidStateException`
+    - `dc.errors.BadArgumentException` --> `dc.BadArgumentException`
+    
+- Functions from `dc.round` have been removed.
+  Please change as follows:
+
+    - `dc.round.floor` --> `Math.floor`
+    - `dc.round.round` --> `Math.v`
```
gordonwoodhull: typo: s/v/round

kum-deepak: Fixed, updated the commit. Thanks :+1:

docs/dc-v4-upgrade-guide.md
```diff
@@ -0,0 +1,62 @@
...
+    - `dc.round.round` --> `Math.round` 
+    - `dc.round.ceil` --> `Math.ceil`
+
+- The previous was of instantiating a chart is still supported.
+  However new way is recommended. For example:
+  
+    - `dc.pieChart = (parent, chartGroup)` --> `new dc.PieChart(parent, chartGroup)`
+
+- In dcv4, inside a `dc` chart functions set and expect `this` to be the chart
+  instance. However `d3` sets it to the d3 element.
```
gordonwoodhull: This an important point. I wonder if anyone is relying on passing dc member functions as callbacks like this.

gordonwoodhull:
Thanks @kum-deepak! You know better than I if there are interface changes. 

I think this covers everything we've discussed, as well as a couple of things I wasn't aware of. 👍

gordonwoodhull:
Looking good!

I'd argue that the *All functions (`renderAll` etc) belong in the chart registry module, since those are the main point of the registry.

kum-deepak:
Makes sense, will shift.

kum-deepak:
I am double mind about something, so, need your opinion.

Current `dc.pluck`, it may have been `dc.utils.pluck`, what do you say?

docs/dc-v4-upgrade-guide.md
```diff
@@ -0,0 +1,62 @@
...
+  Please change as follows:
+
+    - `dc.round.floor` --> `Math.floor`
+    - `dc.round.round` --> `Math.round` 
+    - `dc.round.ceil` --> `Math.ceil`
+
+- The previous was of instantiating a chart is still supported.
+  However new way is recommended. For example:
+  
+    - `dc.pieChart = (parent, chartGroup)` --> `new dc.PieChart(parent, chartGroup)`
```
gordonwoodhull: there is an extra ` = ` here

gordonwoodhull:
`dc.pluck` gets a lot of use, so I don't think we can move it and say "Great care has been taken to ensure API compatibility. Typical usage of this library should see very little differences."

The other changes are necessary for technical reasons. I don't think we should break the interface just to clean up. 

Truth be told,
```js
dc.pluck('x')
```

is longer than

```js
d => d.x
```

and less expressive. There would be even less reason to use `dc.utils.pluck`.

The only reason `dc.pluck` existed was because it was shorter than

```js
function(d) { return d.x }
```

So if anything, we should deprecate it.

kum-deepak:
Thanks makes sense.

kum-deepak:
Please merge this branch now. :+1:

gordonwoodhull:
Oh right, there is also

```js
dc.pluck('x', f)
```

which would become 

```js
(d,i) => f.call(d, d[n], i)
```

I think that one is only used internally, and my oh my is there some confusing code around it, like `stackMixin.title` - totally unintelligible:

```js
        title (stackName, titleAccessor) {
            if (!stackName) {
                return super.title();
            }

            if (typeof stackName === 'function') {
                return super.title(stackName);
            }
            if (stackName === this._groupName && typeof titleAccessor === 'function') {
                return super.title(titleAccessor);
            }

            if (typeof titleAccessor !== 'function') {
                return this._titles[stackName] || super.title();
            }

            this._titles[stackName] = titleAccessor;

            return this;
        }
```

gordonwoodhull:
Merged. Thanks @kum-deepak!

PR #1585, opened 2019-09-28
Convert deprecation and direct method assigment to ES6 style.

kum-deepak:
I have carried out sample conversion of a deprecated method and a case of direct method assignment. Please see if this approach seems acceptable.

For logger.annotate we can take same approach.


gordonwoodhull:
Yes, looks great. 

In both cases it is not as clever as the old solution, which is probably a good thing!

PR #1587, opened 2019-09-29
Several ES6 cleanup

kum-deepak:
I have converted deprecation, annotation and direct method assignment to ES6. Replaced all method deletions with Errors. Also reviewed ES6 markers and cleanup ones that do not need any further action.

In this process added few new test cases as well.

### DIFF
--- a/spec/composite-chart-spec.js
+++ b/spec/composite-chart-spec.js
@@ -110,6 +110,14 @@ describe('dc.compositeChart', function () {
         expect(chart.valueAccessor()).not.toBeNull();
     });
 
+    it('should not allow calling yAxisMin', function () {
+        expect(chart.yAxisMin).toThrowError();
+    });
+
+    it('should not allow calling yAxisMax', function () {
+        expect(chart.yAxisMax).toThrowError();
+    });
+
     describe('rendering the chart', function () {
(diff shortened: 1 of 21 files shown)


gordonwoodhull:
Thanks @kum-deepak! I will review tomorrow.

This replaces #1585 right?

kum-deepak:
Yes!

Sorry, I should have committed to that one. I will close the other one.

gordonwoodhull:
Sorry, I may not have time to review this or #1588 today - things are a bit tight at work. This PR in particular is very important, so I will be sure to take care of it in the next day or two.

Also, a heads-up, I will be on vacation next week. So let's prioritize things so that I don't hold you up while I am away.

gordonwoodhull:
This is great! I think we are near the end of the tunnel. I agree that wrapper functions, while more verbose, are the clearest way to deal with synonyms and deprecations.

I made some copy edits to the migration guide and added the list of removed symbols (synonyms from 1.* IIRC).

I wasn't sure about the general advice to use arrow functions, so I removed that. It didn't seem to have to do with migration, but if it did, feel free to add it back.

This is all merged to the es6 branch. Thank you @kum-deepak!!

PR #1588, opened 2019-09-29
Generate output in `esm` format as well.

kum-deepak:
First take on this. It turns out that rollup is flexible in generating esm. It allows a single file generation at the same time splitting it into multiple chunks. See: https://levelup.gitconnected.com/code-splitting-for-libraries-bundling-for-npm-with-rollup-1-0-2522c7437697

Currently just the basic configuration - a single chunk.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -38,7 +38,7 @@ module.exports = function (grunt) {
             }
         },
         rollup: {
-            js: {
+            es6: {
                 options: {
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
That guide is ... comprehensive ... Not sure I can absorb all of that today. 

But yes, it does sound promising.

I have merged this PR to the es6 branch, thanks @kum-deepak!

PR #1589, opened 2019-10-02
ES6 compliant JSDoc

kum-deepak:
Preview of changes.


gordonwoodhull:
Nice! I take it it’s a bigger win for chart classes because the mixins are functions returning classes and jsdoc doesn’t understand that.

I wonder if the mixins really need to be functions, given that (as I’ve said too many times) they are really base classes.

PR #1591, opened 2019-10-03
Mixins to classes

kum-deepak:
I have converted Stack and Margin mixins to proper classes. Do you reckon any of the other could be converted as well?

### DIFF
--- a/src/base/coordinate-grid-mixin.js
+++ b/src/base/coordinate-grid-mixin.js
@@ -2,7 +2,6 @@ import * as d3 from 'd3';
 
 import {ColorMixin} from './color-mixin';
 import {MarginMixin} from './margin-mixin';
-import {BaseMixin} from './base-mixin';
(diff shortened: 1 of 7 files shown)


gordonwoodhull:
Sorry.... I forgot that the major problem solved by mixins in dc.js is not "mix and match" (which we don't have), but multiple inheritance. 

All the solutions I can find for MI in JavaScript are not prototype-based. 

I am inclined to leave the hierarchy as it is, with the clever "dynamic inheritance" you discovered. The only problem it is causing is slightly more complicated jsDoc. (#1589)

kum-deepak:
Please merge this commit. With these two, 4 of the 'mixins' are classes (Base and Coordinate apart from these two). It does help.

gordonwoodhull:
Okay, I'll do that. Thanks for your patience!

gordonwoodhull:
Rebased and merged.

PR #1593, opened 2019-10-11
ES6 compliant docs

kum-deepak:
All src files has been modified one round. Mostly used semi automated conversions with sampled checks. Seems alright for now.

The library is no longer necessarily namespaced under `dc`, this will need updating in main entry point.

### DIFF
--- a/docs/welcome.base.md
+++ b/docs/welcome.base.md
@@ -2,7 +2,12 @@
 
 Welcome to the dc.js documentation.
 
-The entire library is scoped under {@link dc the dc namespace}. Its child namespaces contain utilities.
+This library may be used as ES6 modules or as a JS file (UMD) in your webpage.
+
+When used as UMD in your webpage, the entire library is scoped under dc namespace.
+
+When used as modules you will see the classes and functions as per your import
+statements.
 
 A class diagram is shown below - mixins are blue and chart classes are green. (Relations between mixins are somewhat subjective.)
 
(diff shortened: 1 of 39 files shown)


gordonwoodhull:
Hi @kum-deepak, I'm back from vacation today.

This cleans up the documentation nicely. But it's a little frustrating that it breaks all links into the documentation. When I answer questions on SO I often link to the particular class and method under discussion, and it would be disappointing if all those links died.

Any thoughts about a decent way to redirect each of the files while preserving the fragment id `#`, e.g. `dc.barChart.html#gap` to `BarChart.html#gap`?

kum-deepak:
I was thinking to leave the current documentation on its location and host documentation of this version under a new URL.

I do not expect everyone to shift to this version immediately. So, keeping both versions may be useful.

gordonwoodhull:
I'd rather not keep old documentation alive and I don't think it will be too confusing for people to look at the new documentation, since the external API isn't changing much.

But I like the idea of generating the new docs to a new location since this makes a huge impact on the documentation filenames.

If you know the new location, please update this PR accordingly. Then I can generate the redirect pages sometime before the release.

gordonwoodhull:
Or if you prefer, I can merge this and open an issue to change the path before release.

Thanks for all your hard work, @kum-deepak!

kum-deepak:
I want to update the entry point of the documents. I will let you when it is ready to merge.

kum-deepak:
This is now ready to be merged. The folder restructure will be taken as a different PR.

gordonwoodhull:
Thanks @kum-deepak. I filed issues for each of us for the missing bits.

Issue #1592, opened 2019-10-08
dc.override remaining in examples


gordonwoodhull:
Ran into a crashing example by accident just now. 

```bash
$ grep override web/examples/*
web/examples/focus-dynamic-interval.html:            dc.override(chart, "yAxisMin", function() {
web/examples/stacked-bar.html:          dc.override(chart, 'legendables', function() {
```

Additionally, the removal of `dc.override` is not mentioned in the v4 upgrade guide.

We should probably run through all of examples (including transitions/resizing/zoom directories) and/or grep `web/` for the items in the upgrade guide.

kum-deepak:
Fixed in #1600

PR #1600, opened 2019-10-19
Remove dc.override

kum-deepak:
Ref #1592. I additionally realized that some Chart classes were not exported. corrected that as well.

### DIFF
--- a/docs/dc-v4-upgrade-guide.md
+++ b/docs/dc-v4-upgrade-guide.md
@@ -37,7 +37,7 @@ Please raise an issue on GitHub if you run into problems not covered here!
     - `dc.round.round` --> `Math.round`
     - `dc.round.ceil` --> `Math.ceil`
 
-- The previous was of instantiating a chart is still supported.
+- The previous way of instantiating a chart is still supported.
   However, it is recommended to use the `new` operator instead. For example:
 
     - `dc.pieChart(parent, chartGroup)` --> `new dc.PieChart(parent, chartGroup)`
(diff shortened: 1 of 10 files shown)


web/examples/focus-dynamic-interval.html
```diff
@@ -1,195 +1,134 @@
...
+    <script type="text/javascript" src="../js/crossfilter.js"></script>
+    <script type="text/javascript" src="../js/dc.js"></script>
+    <script type="text/javascript">
+
+        (() => {
+            class MyLineChart extends dc.LineChart {
+                yAxisMin () {
+                    const min = d3.min(chart.data(), layer => d3.min(layer.values, p => p.y + p.y0));
+                    return dc.utils.subtract(min, chart.yAxisPadding());
+                }
```
gordonwoodhull: the function should use `this` instead of `chart`

docs/dc-v4-upgrade-guide.md
```diff
@@ -70,3 +70,11 @@ chart.on('renderlet', function (_chart) {
     - `dc.coordinateGridChart` --> `dc.coordinateGridMixin`
     - `dc.marginable` --> `dc.marginMixin`
     - `dc.stackableChart` --> `dc.stackMixin`
+
+- `dc.override` has been removed.
+   It was used to override a method in an object (typically a chart).
+   You will now need to extend the appropriate chart class.
```
gordonwoodhull: You can either create a derived class extending the chart class, or you can override specific methods with e.g.

```js
            chart.yAxisMin = () => {
                const min = d3.min(chart.data(), layer => d3.min(layer.values, p => p.y + p.y0));
                return dc.utils.subtract(min, chart.yAxisPadding());
            };
```

since methods on the object also override methods on the prototype chain.

web/examples/focus-dynamic-interval.html
```diff
@@ -1,195 +1,134 @@
...
-        }
-        // e. Increase k by 1.
-        k++;
-      }
+    <script type="text/javascript" src="../js/d3.js"></script>
+    <script type="text/javascript" src="../js/crossfilter.js"></script>
+    <script type="text/javascript" src="../js/dc.js"></script>
+    <script type="text/javascript">
+
+        (() => {
```
gordonwoodhull: This IIFE was just to get indentation to work... It could be replaced with a simple pair of braces now.

gordonwoodhull:
Looks good! Glad to see `dc.override` gone! 

Just a few nits above.

kum-deepak:
Updated, enhanced example on upgraded notes. In example added a comment that in inheritance approach it is easy to access the parent implementation.

gordonwoodhull:
Excellent docs! Thanks @kum-deepak!

Issue #1545, opened 2019-08-11
use specific D3 ES6 modules?


gordonwoodhull:
Instead of `import * as d3 from 'd3';` we can import the specific D3 functions and/or modules we are using. 

This ought to cut down bundle sizes. It's a big task though!

gordonwoodhull:
This also means instead of depending on the d3 package, we'd depend on submodules we actually use, like d3-scale, d3-axis, etc. 

This would allow us to target the components that will be in D3v6 before D3v6 is actually released.

kum-deepak:
Yes, it indeed should be to the roadmap to ES6.

kum-deepak:
As preparation `utils.add` and `utils.subtract` need to be updated to not use `d3[funcName]` to call a function. I will create a separate PR for this.

gordonwoodhull:
The latest d3-selection and d3-transition have not landed yet, but we could try the `two` branch for those ones.

No word yet if/when D3 will do a release candidate - see d3/d3#3333. That would be a little easier than pulling individual dependencies (although we may still want to do that since it's lighter weight).

I hope that the choice between the d3 bundle and the individual modules doesn't cause another split in our build artifacts - that might be one reason to choose one or the other. My guess is that UMD would stay on the bundle and ESM/ESF would go to specific modules, but I'm not sure about that.

kum-deepak:
I am planning to experiment. I am unable to get my head around (without actually writing code) how to tell the build system that in some cases dependencies will come from `d3` while in other from smaller modules.

gordonwoodhull:
Yeah I don’t see it either. Each source file needs to import from particular modules, either `d3` or `d3-stuff`... seems like we’d have to abandon UMD if we went this route.

I’d love to be proven wrong.

gordonwoodhull:
More precisely, we would force our users to use the unbundled D3 modules rather than d3.js.

All of the unbundled D3 modules are available in UMD form, so UMD would be okay.

The interfaces would get complicated because the `d3` dependency would go away and the third line in `dc.js` would resemble

```js
define.amd ? define(['d3-selection','d3-transition',...], factory`
```

and the body would be like 

```js
}(this, (function (exports, d3selection, d3transition, ...) { 'use strict';
```

It would mean a lot more HTML `<script>` includes for folks who are not bundling their sites, potentially slower if the connection is not HTTP2.

[...]

PR #1601, opened 2019-10-20
Remove d3[funcName] for time add/substract functions.

kum-deepak:
Will be required to move to named imports from `d3`. Ref #1545


kum-deepak:
It seems https://github.com/dc-js/grunt-jsdoc-to-markdown has gone missing (probably no longer needed). It was causing tests to fails at Travis.

gordonwoodhull:
Merged! Thanks @kum-deepak!

Issue #1595, opened 2019-10-17
use terser instead of uglify-es


gordonwoodhull:
Apparently I made a bad suggestion in #1541.

uglify-es is no longer maintained and the thing to use is [terser](https://github.com/terser/terser).

Also it looks like we still have a dependency on uglify-js which we aren't using.

kum-deepak:
This is a good idea. We can use rollup plugin to use it.

PR #1602, opened 2019-10-20
Use terser, enhance rollup

kum-deepak:
Ref #1595 

This has package.json change. So, please run `npm i`. Also, I was not sure if I should checkin the package lock file, so, committed that as a separate commit.

This PR starts using terser, but does some more.

I have created a rollup config which does the following:

- UMD and corresponding minified file in `dist/es6`
- Flattened ESM and corresponding minified file in `dist/esf` (useful for importing as modules)
- Split ESM in `dist/esm`
- Minified ESM split modules in `dist/esm-min`

Currently it is not integrated with grunt. To test it, run the following:

```bash
$ rm -rf dist/*; npx rollup --config
```

I have included https://github.com/dc-js/dc.js/commit/b0b66a7e652c9ff2ae31f1a69a0c556db2726f18 which is part of #1601. I wanted to verify that with this change, the ESM version does not have `import * from 'd3'`.

Please see the output. For me also it was first such attempt.

It is clear from the output what rollup module splitting gives. We need to decide which of these six different output formats we need to include in our distribution.

The ESM output also shows that while our code imports everything from d3, the rollup output uses named imports.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,6 @@
       "version": "7.0.0",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
       "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
-      "dev": true,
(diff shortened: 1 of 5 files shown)


kum-deepak:
Closed in favor of #1603, simpler PR with fewer more relevant commit.

PR #1603, opened 2019-10-20
Use terser, enhance rollup

kum-deepak:
Ref #1595 

This has package.json change. So, please run `npm i`. Also, I was not sure if I should checkin the package lock file, so, committed that as a separate commit.

This PR starts using terser, but does some more.

I have created a rollup config which does the following:

- UMD and corresponding minified file in `dist/es6`
- Flattened ESM and corresponding minified file in `dist/esf` (useful for importing as modules)
- Split ESM in `dist/esm`
- Minified ESM split modules in `dist/esm-min`

Currently it is not integrated with grunt. To test it, run the following:

```bash
$ rm -rf dist/*; npx rollup --config
```

I have tested by including https://github.com/dc-js/dc.js/commit/b0b66a7e652c9ff2ae31f1a69a0c556db2726f18 which is part of #1601. I can confirm that with this change, the ESM version does not have `import * from 'd3'`.

Please see the output. For me also it was first such attempt.

It is clear from the output what rollup module splitting gives. We need to decide which of these six different output formats we need to include in our distribution.

The ESM output also shows that while our code imports everything from d3, the rollup output uses named imports.


gordonwoodhull:
Let's include all of them for starts, and then we can see what gets used. I am still not clear if this is all of the formats (where is the famous Michael Jackson Script?) or which ones people will want to use.

Previously because we weren't careful what we put in the NPM package, the package was 24M.

Since we'll just be publishing `dist/` from now, the new size is 5.6M. 

So please go ahead and fix the `grunt` build, removing uglify and restoring `grunt-rollup`. 

Or however you want to invoke rollup from grunt - whatever is most clear.

gordonwoodhull:
As for package-lock.json, yeah it's annoying. Sometimes I check it in, sometimes I don't. 

It only really matters for releases, I think.

kum-deepak:
I am planning to use grunt-shell to integrate. The current grunt-rollup plugin is causing more pain than it is worth.

kum-deepak:
Ready for your review and merge.

kum-deepak:
@gordonwoodhull ready for your review and merging.

kum-deepak:
I am happy that it is now using grunt-shell, the previous version was eating away all the warnings.

gordonwoodhull:
Nice! Thanks @kum-deepak, rebased and merged.

Issue #1458, opened 2018-06-28
bower.json: wrong d3 dependency for v3.04


silentmantra:
v3.04 uses d3.timeFormat, that was introduced in d3 v4.x
however the bower.json contains a dependency to d3 v3.x.

devleaks:
Please keep bower.json in sync with package.json
(as today: package.json has "crossfilter2": "^1.4.5", "d3": "^5.4.0", while bower still "d3": "3.x", "crossfilter2": "~1.4")
Thank you.

seantai:
Is there any plan for this to be updated? It seems like the bower.json is still out of date in its d3 dependency.

gordonwoodhull:
Sorry, bower is not on my radar, obviously. 

I'll correct this in the next release, and see if it can be generated from package.json somehow.

seantai:
Cool, thanks!

gordonwoodhull:
Sorry for the extreme delay. 

I'm updating `bower.json` in version 3.1.7 and deprecating Bower support - dc.js version 4 will not support Bower. Please make sure you have configured for version 3 if you are still using Bower.

PR #1606, opened 2019-10-27
Drop bower support.

kum-deepak:
Bower does not seem to be maintained https://bower.io/blog/2017/how-to-migrate-away-from-bower/.


gordonwoodhull:
Sigh, I’ve not been keeping it up to date either, see #1458

kum-deepak:
None of the new versions of npm, rollup, webpack, yarn are going to play nice with bower. I do not see why we should still maintain it.

kum-deepak:
If we decide to keep it, we should automate generation of `bower.json`.

gordonwoodhull:
I agree. I’ll update it in v3, and before merging this for v4 I just want to verify there aren’t any further comments or workarounds referring to bower. (Not at a real computer atm :)

gordonwoodhull:
Please add a note to the migration guide!

kum-deepak:
Done!

gordonwoodhull:
We are following [How to drop Bower support](https://bower.io/blog/2017/how-to-drop-bower-support/) perfectly with the version change. 

Will update the versions on develop for another 3.0 release, and the changelog can serve as notice that we're dropping Bower.

Thanks @kum-deepak, rebased and merged.

PR #1609, opened 2019-10-28
Remove polyfills

kum-deepak:
During the process also found one example that was not using local js/css files (fixed as a separate commit).


gordonwoodhull:
Thanks @kum-deepak! It's a relief to have those gone. Rebased and merged.

PR #1610, opened 2019-10-30
License banner in output JS files

kum-deepak:
This was done by uglify. Now it uses a rollup plugin.

Currently all JS files have it enabled. It can be controlled though. Please share your views.


gordonwoodhull:
Yes, this is the right thing to do - the license banner should be at the top of each distributed source. Wish it were a bit shorter but it's fine.

gordonwoodhull:
Rebased and merged. Thanks @kum-deepak!

PR #1611, opened 2019-10-30
Remove bower from install methods.

kum-deepak:
Cleanup README


gordonwoodhull:
Thanks @kum-deepak.

This will go live on the main GitHub code page when es6 is merged to develop, which is the right timing.

PR #1612, opened 2019-10-31
Module sample

kum-deepak:
While creating the samples the ideal structure for modules have started emerging. At present I have made each source file to be a module in the output as well. This seems to be ok for now. In future we may rearrange code based on usage patterns.

The commit https://github.com/dc-js/dc.js/commit/8ceec9e428426102ae500cb8525321be37856c45 has the new module structure. I have also set the `esm` to be the default module as per package.json. We will figure out if this works better during development of samples.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -8,5 +8,11 @@
 !LICENSE
 !dist/**/*
 
+# ES6 modules
+!src/**/*
+
+# Exclude index-with-version from ES6 modules
+src/index-with-version.js
+
 # ignore the source map files, these have no use without source files
 dist/**/*.map
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -79,11 +79,11 @@ module.exports = function (grunt) {
                 tasks: ['test']
             },
             reload: {
-                files: ['<%= conf.dist %>/es6/<%= conf.pkg.name %>.js',
+                files: ['<%= conf.dist %>/<%= conf.pkg.name %>.js',
                         '<%= conf.dist %>/style/<%= conf.pkg.name %>.css',
                         '<%= conf.web %>/js/<%= conf.pkg.name %>.js',
                         '<%= conf.web %>/css/<%= conf.pkg.name %>.css',
-                        '<%= conf.dist %>/es6/<%= conf.pkg.name %>.min.js'],
+                        '<%= conf.dist %>/<%= conf.pkg.name %>.min.js'],
                 options: {
                     livereload: true
                 }
@@ -117,7 +117,7 @@ module.exports = function (grunt) {
                 src: [
                     '<%= conf.web %>/js/d3.js',
                     '<%= conf.web %>/js/crossfilter.js',
-                    '<%= conf.dist %>/es6/<%= conf.pkg.name %>.js'
+                    '<%= conf.dist %>/<%= conf.pkg.name %>.js'
                 ]
             }
         },
@@ -133,7 +133,7 @@ module.exports = function (grunt) {
                     // source files, that you wanna generate coverage for
                     // do not include tests or libraries
                     // (these files will be instrumented by Istanbul)
-                    '<%= conf.dist %>/es6/<%= conf.pkg.name %>.js': ['coverage']
+                    '<%= conf.dist %>/<%= conf.pkg.name %>.js': ['coverage']
                 },
 
                 // optionally, configure the reporter
@@ -199,7 +199,7 @@ module.exports = function (grunt) {
         },
         jsdoc2md: {
             dist: {
-                src: '<%= conf.dist %>/es6/<%= conf.pkg.name %>.js',
+                src: '<%= conf.dist %>/<%= conf.pkg.name %>.js',
                 dest: 'web/docs/api-latest.md'
             }
         },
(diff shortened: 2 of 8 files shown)


kum-deepak:
This also has artifacts as I am using this as package location for npm in the sample. Please access the sample at https://github.com/kum-deepak/dc-sample

kum-deepak:
It turns out, at least, with the stocks example with rollup, using modules through the index.js in esm (which the default) or from specific module files, makes no difference in bundle size.

To regenerate bundles, switch to a folder, and run `npx rollup --config`

kum-deepak:
I am not attempting to write sample that would use modules directly from browser.

gordonwoodhull:
Yes, I took a look at your sample repo. Thanks for putting that together! 

This shows that tree-shaking works just as well either way. For example, you can search bundle.js for a chart which is not used in the stock example, like "heatmap" or "sunburst", and it is gone either way. I think the modules are included or excluded whole, so there are still some of the utilities for each of these charts, but not the charts themselves.

I think the different configurations would mostly be of interest to people who are not using bundlers, or who are using different bundlers. I say we put them all out there and wait for feedback from users.

kum-deepak:
Sorry for the long gap. Been traveling all over.

I tried using modules directly from browser. Eventually, quite clumsy, but successful. Using dc as modules becomes tricky as the modules use d3 and browser won't know d3 as it does not point to a URL. So, I needed to re-process all dc modules with path of d3. Then I realized that d3 just includes all other d3-modules. So, needed to reprocess d3 and all sub modules as well. So, my final conclusion was that if one has to do so much reprocessing to convert module id to paths (URLs), why not one would just bundle.

Also realized that unpkg has an option for dynamic rewrite of module id to URLs, however, it may cause multiple copies of a a module get loaded (sometimes because of few references with .js or without), this will break modules that has state (like our Chart Registry). So, quite risky to recommend to users.

gordonwoodhull:
I agree, it sounds too cumbersome to use modules directly from the browser, at least with tools we know.

Is the module list commit useful to merge? 

Want to move dc-sample to maybe github.com/dc-js/modules-demo?

It's interesting that we will no longer be able to install github branches of dc.js via npm once the artifacts are gone. 

That has been useful in the past for features that are in demand but not ready to merge, so I'll have to figure out a new way to do that.

I think we are almost ready to release an alpha on npm, which would be a better solution for modules-demo.

kum-deepak:
I will let you know when it is ready. I have going to tidy up certain things.

kum-deepak:
I have automated the list of modules using glob - should make it easier to maintain. I want to test it again before merging.

kum-deepak:
After attempts to create samples, I have realized only few of the formats make sense as of now:

- Modules can not be easily used by the browsers directly.
- If we use any of the modern bundlers (with tree shaking), it does not matter whether we directly reference flattend file, index file, or, the individual module.
- If we use a bundler it does not matter if the included code is already minified.

Suggestion:

- Drop output format esf.
- Drop minified versions of output format esm.
- Place the UMD bundle (and minified copy) directly in dist/

We can bring back whichever formats are needed.

On a side note, only when using modules directly from a browser we need the files exposed through URLs. In case of bundlers people can use non npm versions of this library through GitHub (provided dist folder is checked in).

gordonwoodhull:
That makes sense, thanks. 

I think instead of esf we might publish `src/` to npm - this is how crossfilter and D3 support ES6 modules. The source is already modules and folks can bundle it however they want.

As for checking in the `dist/` folder, I think a reasonable compromise might be to check it in only for experimental branches. The `master` and `develop` branch won't need artifacts because releases are better accessed through npm, but any unofficial releases can go on branches that do include dist, so people can use those through GitHub.

kum-deepak:
Yes, just using src for es6 module would work. Only issue would be in using/exporting version which comes from package.json and relies on a rollup plugin.

Decision point:

- Do we not include version in es6 modules (but include in UMD)?
- Do we drop version altogether?

A solution can be worked out depending on our decision.

gordonwoodhull:
In crossfilter, we decided to keep the version in the bundled UMD module but drop it in the es6 modules.

The main purpose of the version is to verify what bundle you are looking at, when you receive dc.js from somewhere (or looking at some web site that uses it).

If you install the npm module, there will be a directory under `node_modules/` with package.json, as well as other ways to verify the version like `npm ls dc` - so it's redundant there.

I guess it's unlikely that someone would copy dc.esm.js and use it outside of `node_modules`, but I'd be inclined to keep the version in there just to be safe.

kum-deepak:
That makes sense. Just to ensure that we are on the same page:

- dist will have UMD bundle and minfied copy. This will have version included.
- src can be used as ES6 module, the default index here will not have version included.

gordonwoodhull:
That sounds right. Crossfilter is doing basically the same thing you have in your commit. Version 1.5 has only been out for a week, but so far the scheme seems to be holding up.

kum-deepak:
Please check, should be ready for merge now.

gordonwoodhull:
Much simpler!

I didn't realize we were also talking about dropping the esm build, but you're right, if we distribute the source then why distribute esm? Seems to me rollup must be addressing some use-case which isn't relevant to us (like bundling CommonJS into ESM or something).

gordonwoodhull:
Looks good!

Rebased, removed unused glob dependency, and pushed. Thanks @kum-deepak!

PR #1616, opened 2019-11-18
d3 modules

kum-deepak:
This rollup config works. Generated code is not optimal but free of hacks.

Please check generated dc.js as well.


gordonwoodhull:
I like the idea of this a lot, but it looks like it is only replacing d3-selection with d3 for the no-modules case in dc.js:

```js
(function (global, factory) {
/* 1 */  typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports, require('d3'), require('d3-selection')) :
/* 2 */  typeof define === 'function' && define.amd ? define(['exports', 'd3', 'd3-selection'], factory) :
/* 3 */  (global = global || self, factory(global.dc = {}, global.d3, global.d3));
}(this, (function (exports, d3, d3Selection) { 'use strict';
```

1. CommonJS
2. UMD
3. Globals / no modules

I would hope that it would supply (require or define) d3 to the factory twice for CommonJS or UMD, as it does for the global case. 

There aren't a lot of CommonJS users of dc.js, but as of 3.* UMD users only need to pull in d3.js and dc.js, not d3-selection.js.

kum-deepak:
For 1 and 2, `d3` is going to be installed via npm locally, that will cause all d3 modules to be installed as node packages. So, I guess 1 and 2 should work fine.

There is a potential issue until full conversion has been carried out, some classes can get loaded twice in 1 and 2. This will not be the case after full conversion as d3 will no longer be directly included.

gordonwoodhull:
In UMD, every module load is its own asynchronous http request, so maybe it would work but it would be slow.

It also means that users would have to copy that many more files, since you don't want to serve node_modules in a production server.

So maybe that's the price of progress but it's a downgrade for UMD users.

kum-deepak:
I have an idea, we can do a two step translation with rollup:

- in which first we rewrite the d3-module to d3 to generate esm, and,
- in the next pass we generate the UMD bundle.

Let me check if the hypothesis works.

gordonwoodhull:
I guess I'm okay with deprecating UMD in this way if we can't figure out how to fix this. Or I can see a one-line `sed` fix for this. 😈

kum-deepak:
The following actually works. However, kills the sourcemap. So, I think we should go down this route.

```javascript
import {terser} from 'rollup-plugin-terser';
import json from 'rollup-plugin-json';
import license from 'rollup-plugin-license';

const jsonPlugin = json({include: 'package.json', preferConst: true});
const licensePlugin = license({
    sourcemap: true,
    banner: {
        content: {
            file: 'LICENSE_BANNER'
        }
    }
});

const umdConf = {
    file: 'dist/dc.js',
    format: 'umd',
    name: 'dc',
    sourcemap: true,
    globals: {
        d3: 'd3'
    }
};

const umdMinConf = Object.assign({}, umdConf, {file: 'dist/dc.min.js'});

export default [
    {
        input: 'src/index-with-version.js',
        external: ['d3', 'd3-selection'],
        plugins: [
            jsonPlugin
        ],
        output: {
            file: 'tmp/dc.esm.js',
            format: 'esm',
            sourcemap: true,
            paths: {
                'd3-selection': 'd3'
            }
        }

    },
    {
        input: 'tmp/dc.esm.js',
        external: ['d3'],
        plugins: [
            terser({include: [/^.+\.min\.js$/]}),
            jsonPlugin,
            licensePlugin
        ],
        output: [
            umdConf,
            umdMinConf
        ]
    }
];
```

kum-deepak:
If we also add the following:

```javascript
    paths: {
        'd3-selection': 'd3'
    }
```

The output becomes:
```javascript
(function (global, factory) {
  typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports, require('d3'), require('d3')) :
  typeof define === 'function' && define.amd ? define(['exports', 'd3', 'd3'], factory) :
  (global = global || self, factory(global.dc = {}, global.d3, global.d3));
}(this, (function (exports, d3, d3Selection) { 'use strict';
```

Which one do you think is better - adding the path section or without it?

gordonwoodhull:
The module header looks good! I'm confused though, does this kill the sourcemap?

gordonwoodhull:
BTW

1. I'll be on vacation next week. I'll try to review and merge stuff but I might be slow.
2. I always rebase PRs before merging, and I just realized that I could force-push the rebased branch so that the PR refers to the correct commits and it shows up as a merge rather than close. The drawback is that this modifies your repo - is that cool with you?

kum-deepak:
In the two stage generation, https://github.com/dc-js/dc.js/pull/1616#issuecomment-555829777, the final stage uses 'tmp/dc.esm.js', so, the source map refers to this file instead of the actual source files.

We may only use https://github.com/dc-js/dc.js/pull/1616#issuecomment-555835008 with our current rollup config - the will have 'd3' repeated many times, otherwise the sourcemap will refer to the original sources.

We can park this decision. Maybe, we can ask a support question in rollup to find out best way to do this.

kum-deepak:
> BTW
> 
>     1. I'll be on vacation next week. I'll try to review and merge stuff but I might be slow.
> 
>     2. I always rebase PRs before merging, and I just realized that I could force-push the rebased branch so that the PR refers to the correct commits and it shows up as a merge rather than close. The drawback is that this modifies your repo - is that cool with you?

This actually works better for me. Please do that.

gordonwoodhull:
Cool - I didn't understand the "also" in https://github.com/dc-js/dc.js/pull/1616#issuecomment-555835008. You meant we can use `paths` independently. 

I have tried it and it works great. Thanks for researching all this.

Repeating the same dependency N times is ugly but it will be cached at the UMD level and cost nothing. The user only has to worry about one file.

I'm sure it's fine for CommonJS too, although people only use that when they want to render dc.js server-side (rarely).

I'm sure it will be a slog to find the module for all those`d3.*` but hopefully soon we can move onto the fun part of trying the new D3 modules. I don't think there are so many API changes that will affect us, but `d3.event` is gone from d3-selection, and it will be gone from d3-brush when 6.0 is out.

kum-deepak:
Well it turns out converting to `d3-*` was not that much work. `rollup` had converted the imports to specific imports from `d3` when generating split modules. That helped.

Work was needed to rename variables to avoid shadowing. Time to test all samples before merging.

gordonwoodhull:
Wow! This is fantastic!

kum-deepak:
Finishing some rough edges in rollup config, but should be there soon.

kum-deepak:
Completed for now. Ready for review :smile:

gordonwoodhull:
Squashed two of the commits, rebased and merged. Thanks @kum-deepak!

gordonwoodhull:
And yeah, the UMD header is absurd, but I like how the source still distinguishes between the modules even though they resolve to the same thing.

PR #1621, opened 2019-11-22
Cleaning up eslint rules and making stricter

kum-deepak:
I am realizing that in process of moving to es6, the code has become cleaner. This makes me believe that many rules that were configured to be "off" will now pass as is or with little effort.

I am going to start with "eslint:recommended" and proceed incrementally.


kum-deepak:
The following were off, so removed (it makes no difference):

- no-bitwise
- guard-for-in
- no-use-before-define
- no-plusplus
- strict
- max-statements
- semi
- no-cond-assign
- no-debugger
- no-eq-null
- no-eval
- no-var
- prefer-const
- no-extra-semi
- no-unused-expressions
- block-scoped-var
- no-iterator
- no-loop-func
- no-multi-str
- valid-typeof
- no-proto
- no-script-url
- no-shadow
- no-new-func
- no-new-wrappers
- dot-notation
- no-invalid-this
- require-yield

kum-deepak:
Add  `"extends": "eslint:recommended"`.

Needed to switch off following rules:

- no-useless-escape
- no-console

Also needed to remove globals declaration from stock.js as those identifiers have already been declared global in eslintrc.

kum-deepak:
Following are already implied by `eslint:recommended`, so, removed:

- no-irregular-whitespace
- no-empty
- no-undef
- no-extra-semi

kum-deepak:
Cleaned up. Consistent conventions.

kum-deepak:
Ready to merge.

kum-deepak:
I have split into smaller commits for easy verification of the changes. All the commits can be squashed while merging.

gordonwoodhull:
Nice cleanup! Thanks @kum-deepak, squashed and merged.

PR #1622, opened 2019-11-23
Template strings

kum-deepak:
Please check https://github.com/dc-js/dc.js/commit/bf1be40af9dec5c7e9a9266d1c6c94a4b68df9be and https://github.com/dc-js/dc.js/commit/1606ac7c837ed6aa13bd86e70f407a1b272e462e.

I personally like template strings better than concatenation. However, not sure, if it will be general consensus.

I have tried automated conversion using eslint. See if this makes sense.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -58,8 +58,8 @@
     "arrow-parens": ["error", "as-needed"],
     "space-before-function-paren": ["error", "always"],
     "no-useless-concat": "error",
-    "prefer-template": "off",
-    "template-curly-spacing": "off",
+    "prefer-template": "error",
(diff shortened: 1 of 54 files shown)


gordonwoodhull:
I like it too! It’s super helpful for stuff like `translate(x,y)`.

Some of the simpler ones like where it’s just appending a class name to an element name, it’s not as clear a win, but I think it’s still worth doing.

kum-deepak:
Please merge #1621 first. Then when it is rebased there should be only two commits here.

gordonwoodhull:
Thanks @kum-deepak!

Issue #987, opened 2015-08-14
elasticX(true) on Composite breaks rangeChart zooming


eteubert:
If `elasticX(true)` is enabled on a composite chart, connected to a rangeChart, it breaks rangeChart zooming. Setting `elasticX(false)` "fixes" the behavior.

The bug first occurs in `beta.9`.

jsfiddle: https://jsfiddle.net/ocxxuyL5/1/

alan-unravel:
Okay so beta.9 pushed the elasticX() from the parent chart down to it's children. As you can see [here](https://github.com/dc-js/dc.js/blob/develop/src/coordinate-grid-mixin.js#L1086)  focus will be ignored if elasticX is set to true.

eteubert:
Ah ok that makes sense. It should just be mentioned in the API docs. And/or setting elasticX on composite charts should not be available then, or generate an error.

alan-unravel:
We could probably generate a warning of some sort and it should definitely be in the docs. I don't see why it would be disabled though since some people use composite charts without range charts.

gordonwoodhull:
I don't understand. What's the use case for a focus chart with `elasticX`? Its domain will always be set by the range chart.

Or alternately, is there a use case where some of the subcharts are elasticX but others are not? Don't they always share the X axis.

I think I must be missing something here

alan-unravel:
I think you are right, I can't really think of a use-case.

gordonwoodhull:
I guess it could be useful to have the first render use elasticX, and then subsequently just get the domain from the range chart.

gordonwoodhull:
Okay, I'll consider this a documentation issue: the fact that `elasticX` will disable the range/focus feature is mentioned only [in `.focus()`](https://github.com/dc-js/dc.js/blob/master/web/docs/api-latest.md#coordinategridmixinfocus), not in `.rangeChart()` or `.elasticX()` and it should be mentioned those places.

PR #1623, opened 2019-11-25
added caution message for mouse-zoomable conflict with elasticX

privateOmega:
@gordonwoodhull I found that I was unable to zoom-in to my graph even with mouseZoomable(true) and documentation didn't explicity specify this case.

After some googling I found your answer in [SO](https://stackoverflow.com/questions/56550825/mouse-zooming-not-working-for-dc-js-line-chart/56568963#56568963), and I thought users in future shouldn't be wasting time on this. The commit contains update in documentation and rebuild docs.

This can be removed once the below is the case.

> dc.js should only "elast" the X domain once (on render) if both options are set


gordonwoodhull:
Thanks @privateOmega! Related: #987, `elasticX` is not compatible with focus chart.

I am tempted to do what is probably expected and turn off `elasticX` after the first render, in either case. 

(Or perhaps, since it would also be surprising for a parameter to change on its own, define elasticX as only active on render not redraw, if either of these features are enabled.)

privateOmega:
@gordonwoodhull Since you have plans to fix it. I am abandoning the PR.

gordonwoodhull:
I’m going to reopen this as a reminder to fix this. I’m currently on vacation but I think it will be a quick fix when I get back.

Otherwise I think we would have to document it in at least three places!

privateOmega:
@gordonwoodhull Sure. Sounds perfect.

gordonwoodhull:
I started this on the `elasticX-only-render` branch, if anyone wants to try it out. 

So far I am seeing focus charts getting set to `[x1, 1]` instead of `[x1, x2]`, and getting reset to either `[0, 1]` or `[1, 1]` when the selection is cleared, so I clearly missed something. 

Hope to return to this soon.

PR #1624, opened 2019-11-25
Cleanup npm denedencies

kum-deepak:
Removed unused npm packages.

While at it, I realized that there was a dependency conflict between eslint and grunt-eslint. So, switched to grunt-shell for the integration. This also allowed adding one more task for `eslint --fix`.


gordonwoodhull:
Thanks @kum-deepak! Looks good!

Issue #1598, opened 2019-10-18
put docs in new location


gordonwoodhull:
As discussed in #1593.

I know there may be larger restructuring of the website, but in case that doesn't happen, this is just a reminder to @kum-deepak to move the docs to a new directory before release of 4.0. 

I'll handle the redirects in #1597.

gordonwoodhull:
@kum-deepak, what are your thoughts here? #1627 did not change the documentation folder; instead we went with moving the sources and then generating to the old locations.

I am okay with generating redirects within the same folder. Should we close this as outdated?

kum-deepak:
I think we can mark it outdated and close it. All the intents, that I had, have been fulfilled.

gordonwoodhull:
Cool, I misunderstood. 

Moving the source of the docs and web while keeping the same URLs is a great move, thanks!

Issue #1579, opened 2019-09-19
eliminate and ignore more artifacts


gordonwoodhull:
Artifacts go into dist/ now, and we ignore them for npm.

We could also `.gitignore` this directory, and I think we could also ignore a lot of `web/`

We could probably have done this in the past but since we were checking in the root-level artifacts it wasn't as noticeable.

kum-deepak:
I was thinking of rearranging docs/web in such a way to keep source and final layout (current web) in separate folders. This way we can totally ignore the output folder.

Let me know if it makes sense?

gordonwoodhull:
I like it. Most if not all of `web/js` could be copied from `node_modules`. (We'd need to make sure the build fails if anything is missing.)

I'd want to keep all URLs for pages the same, so as not to break links. Another constraint is that a lot of people use github.io as a CDN (despite being told not to). I don't want to break a lot of old demos.

This means that all the key js and css assets should stay at the same URLs. 

Here is the github.io documentation for redirects:
https://help.github.com/en/articles/redirects-on-github-pages

Since it relies on Jekyll for the destination page it does not look like we can redirect assets. We would have to redirect pages manually (using an HTML stub at the old URL).

kum-deepak:
Jekyll generates an html file with meta refresh in the head, which causes the redirect. If needed these files can be hand generated.

Once the new structure is finalized, either we can copy the critical assets in original paths or generate redirect files (same way as Jekyll) would have done.

The documentation site can also move to http://dc-js.github.io/ from http://dc-js.github.io/dc.js/.

gordonwoodhull:
Fixed by #1627 - the sources have been moved, but we still generate to the old locations.

Issue #1565, opened 2019-08-30
Proposal - update folder layout


kum-deepak:
The proposed changes and their rationale:

- rename `web` to `docs` - this will allow easier management of github pages, by changing the project settings in github. A separate branch and additional steps will no longer be needed.

- create a `dist` folder - this would allow us to create multiple sub folders, `es6` with rolled up file, `esm6` with each module in a file of its own, `css` with css files etc. Further this makes IDEs to clearly identify generated files against source files, making job of refactoring easier.

- creae an `.npmignore` file - this way contents of the npm package can be limited to only files that are needed by users. It will trim down the package size significantly.

gordonwoodhull:
We looked into the new github folders. The problem is now you need to check in all of the html artifacts of documentation, which are numerous. Currently we only check those into gh-pages, so it's not so messy. I'm open to discussion on this.

I completely agree on the other two points. I feel bad that the npm distro has been so huge for so long.

gordonwoodhull:
Personally I would rather not check in any artifacts at all.

kum-deepak:
Makes sense. Will work on only 2nd and 3rd points.

kum-deepak:
I have pushed some commits, currently in charts branch:

- https://github.com/dc-js/dc.js/pull/1561/commits/f0404e66437888432c55a6155d17ff78cd316c3b - dropped browserify task
- https://github.com/dc-js/dc.js/pull/1561/commits/4eb998f4f1bfe7791d1e7ee6233c9a94bf3502cc - generated js files now go to dist/es6 and css files go to dist/style folders
- https://github.com/dc-js/dc.js/pull/1561/commits/2d20b19a1c96d43dc4d81c1fb7d8d69f065e210e - first cut of npmignore. First all files are excluded then selectively included. Please check list of files included.
- https://github.com/dc-js/dc.js/pull/1561/commits/636f2046a182859dcec54c13fbc2f9c97a3cfb1f - updated paths as per new folder structure in package.json

kum-deepak:
Question: is watch:reload grunt task used?

gordonwoodhull:
Fixed by #1627 - we are keeping the old URLs but moving the sources so that `web/` and `dist/` are 100% generated.

PR #1627, opened 2019-11-30
Restructure folder structure

kum-deepak:
See #1598, #1579, #1565


kum-deepak:
I am planning to make test running independent of `web/`.

kum-deepak:
Please review and see everything makes sense. Completed work from my end.

kum-deepak:
I realized that link for source code link for examples examples will break when this version is released. Please let me know if I should fix those in this PR or in another PR when this one is merged.

gordonwoodhull:
This is a huge change! I like the approach of keeping everything in the same place on the external site, but removing the output directory from the sources.

Unfortunately this seems to have caused some eslint breakage:
1. The lintable files list includes `web/stock.js` which is an artifact which doesn't exist until `grunt copy` is called.
2. eslint is now checking a number of 3rd party files which fail:

```
/home/gordon/src/dc.js/spec/3rd-party/compare-versions.js
/home/gordon/src/dc.js/spec/3rd-party/crossfilter.js
/home/gordon/src/dc.js/spec/3rd-party/d3.js
/home/gordon/src/dc.js/spec/helpers/jasmine-jsreporter.js
```

As an aside, it also seems that the eslint update from last week requires a newer version of node (due to using newer regexp flags). I have node 8.9.1 and I am wary of upgrading. I backed out to eslint 6.5 and am up and running again - will look into this later.

I'm pushing a fix for the documentation links. Please look into the linting issues and update this PR.

kum-deepak:
Fixed list of files for linting.

I have started using `nvm` for my node version management. Different projects like different ones. For dc I have using node v11.

kum-deepak:
Our Travis build also use node v11.

gordonwoodhull:
Thanks for the `nvm` tip! The reason I have been fearful of changing node versions is `node-sass`, which is incompatible with many versions, especially on Linux. `nvm` makes it easy to switch: I tried node v13 and then dropped back to v11 when that didn't work.

It's kind of hard to keep up with all the consequences, but I think this PR cleans up the generation while being 100% backward-compatible and URL-preserving. My only question was about #1598, which I asked over there.

Thanks @kum-deepak!

Issue #1630, opened 2019-11-30
Convert chart registry and logger to ES6 classes


kum-deepak:

kum-deepak:
Implemented by #1631.

PR #1631, opened 2019-12-01
Convert Logger and Chart Registry to ES6 classes

kum-deepak:
I have also cleaned up methods `logger.deprecate` and `logger.annotate`. Ref #1630


gordonwoodhull:
Thanks @kum-deepak!

PR #1634, opened 2019-12-04
Need to explicitly set `chart` as `this`

kum-deepak:
An error was stopping this example to work correctly. I have tried to fix it, can you think a better way to solve it?

Accordingly we would need to put a note in the upgrade guide.

### DIFF
--- a/web-src/examples/cap-and-sort-differently.html
+++ b/web-src/examples/cap-and-sort-differently.html
@@ -57,7 +57,7 @@ var chart = dc.pieChart("#test");
     }
     items.sort((a,b) => d3.ascending(a.key, b.key));
     if (chart.othersGrouper()) {
-      return chart.othersGrouper()(items, rest);
+        return chart.othersGrouper().call(chart, items, rest);
     }
     return items;
 


gordonwoodhull:
Drat, looks like I added this example recently and didn't test it under v4/es6.

The solution is okay, but we should think about this and examine the source for other places where the problem might come up. 

I seem to recall other instances where the default for a property is a method which uses `this`.

I suppose when Safari gets [class fields support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Class_fields#Browser_compatibility) and we drop Edge or consider it to be the Chromium one, class fields would be a general and clean solution.

Alternately, we could simulate the effect of class fields by adding

```js
this.othersGrouper = this.othersGrouper.bind(this);
```

to the constructor. The problem comes up all the time in React programming.

gordonwoodhull:
I suppose the other possible solution is 

https://github.com/dc-js/dc.js/blob/75d4770fd36191fb11f071b3aa831feede20e02b/src/base/cap-mixin.js#L178-L181

```js
return this._othersGrouper.bind(this)
```

gordonwoodhull:
Closing in favor of a full investigation in #1638 

It's a shame we didn't have any tests or examples wrapping these functions, so I'm glad I accidentally threw a wrench in the works of our ES6 transformation. 

It's very important to be able to wrap things like `_othersGrouper` and `_defaultHeightCalc`, and it should be as easy as it was in DCv3.

PR #1636, opened 2019-12-05
Watch `web-src` and `docs` as well during `watch:test` grunt task.

### LINKED ISSUE #1635: The watch scripts should include web-src
Realized it hard way while working on #1634

kum-deepak:
Fixes #1635


gordonwoodhull:
Thanks @kum-deepak!

PR #1637, opened 2019-12-05
Move `Legend` and `HtmlLegend` to `charts` folder

kum-deepak:
While working with examples I started feeling that these classes belong better to the `charts` folder. This way regular users are not likely to import anything from `base` folder.

I also sorted imports in index.js.


gordonwoodhull:
I noticed that too. Thanks as always for cleaning it up!

PR #1639, opened 2019-12-17
Bind `this` in default accessor function by using =>

### LINKED ISSUE #1638: when a default of a property is a method, `this` should be bound
#1634 pointed out that it's annoying to wrap the default `capMixin.othersGrouper` because it uses `this`

The problem is endemic to the ES6 transformation. 

In DCv3, almost all of these methods referenced `_chart`, not `this`, so the user didn't have to worry about passing the chart as `this` using `.call()` or `.apply()`

All of these methods need to be identified and bound to `this` in the constructor. In a year or two they can become [class fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Class_fields#Browser_compatibility) with arrow functions, to clean up the mess.

There may be exceptions. 

I will do a thorough search for uses of `this` in DCv3. I expect they will be rare and accidental - almost everything used the closure for private fields and methods.

kum-deepak:
An approach to resolve #1638.

### DIFF
--- a/src/base/base-mixin.js
+++ b/src/base/base-mixin.js
@@ -79,9 +79,17 @@ export class BaseMixin {
         this._isChild = undefined;
 
         this._minWidth = 200;
+        this._defaultWidthCalc = element => {
+            const width = element && element.getBoundingClientRect && element.getBoundingClientRect().width;
+            return (width && width > this._minWidth) ? width : this._minWidth;
+        };
         this._widthCalc = this._defaultWidthCalc;
 
         this._minHeight = 200;
+        this._defaultHeightCalc = element => {
+            const height = element && element.getBoundingClientRect && element.getBoundingClientRect().height;
+            return (height && height > this._minHeight) ? height : this._minHeight;
+        };
         this._heightCalc = this._defaultHeightCalc;
         this._width = undefined;
         this._height = undefined;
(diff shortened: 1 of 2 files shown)


kum-deepak:
The example cap-and-sort-differently.html works with this change.

gordonwoodhull:
Nice! Much simpler than using `.bind(this)`, and almost identical to the class fields approach except using constructor scope.

The trick will be identifying all of these default accessor functions. There are probably not more than a dozen, but I see no automated way to find them.

kum-deepak:
Found few more such cases in base-mixin. I tried few greps across files, so far unable to find more such cases. Have you been able to locate any more such cases?

gordonwoodhull:
I am documenting my investigation, which came up clean, in #1638.

It's not necessary to transform `_defaultData`, because it doesn't use `this`. Otherwise LGTM!

kum-deepak:
I realized that similar pattern was already followed in most of the cases. So, converted `_defaultData` also to the same pattern.

All clear from my end to merge.

gordonwoodhull:
OK, I don't think we want all defaults to use this pattern, but it's harmless to implement `_defaultData` this way.

Thanks @kum-deepak!

PR #1640, opened 2019-12-17
Misclleneous

kum-deepak:
Few small housekeeping type commits. :)

Becoming lazy, created one instead of three PRs.


gordonwoodhull:
Thanks! It will take me a little while to get used to `web-src`. I'm not sure why git allowed me to check that in, given that `web/` is in `.gitignore`.

I recall seeing `grunt server` going into an infinite loop, hopefully the 3rd-party fix will take care of that.

kum-deepak:
I tested `grunt server` now, it seems to be working for me.

You can also try `grunt test-n-serve`, it runs tests, start the server, and enters into watch to run tests as needed.

PR #1641, opened 2019-12-20
Update contribution guidelines

kum-deepak:
Please see if it makes sense. Feel free to make changes.


gordonwoodhull:
This is great - very clear, and the reorganization helps a lot.

At the cost of repeating one sentence, I split out testing into its own section. There are a lot of methods (!) and this interrupted the flow of the development guidelines. Arguably it could be moved to the end.

I didn't feel right about plugging JetBrains. I haven't tried it but I assume other IDEs have this capability as well.

PR #1643, opened 2019-12-23
added build artifacts

privateOmega:
@gordonwoodhull This PR contains `dist/` with umd module files generated using rollup, which I think was missed by @kum-deepak accidentally.


gordonwoodhull:
Thanks @privateOmega.

As of 4.0 we don’t plan to check in artifacts anymore. 

We’ll distribute the builds via npm and friends, and we’ll only check in artifacts in the rare occurrence where we want a branch available through npm.

I think this is the current best practice; e.g. this is what D3 does. It cleans up the commits and it reduces mistakes.

Any thoughts or objections on this?

privateOmega:
@gordonwoodhull So what you are proposing is, the builds will be generated just before publishing to npm? Sounds like a good idea.

gordonwoodhull:
Exactly. We already build right before we publish to npm and github.io. 

The difference is that we won't check `dist/` or web/` into the repo anymore - these are purely generated.

(We were also publishing the entire repo to npm, which was a ridiculous 21M. When 4.0 is released, it will only include `dist/`, `src/`, license, package, and readme - should be under 3M.)

Issue #1650, opened 2020-02-07
3.1.9 crossfilter is not a function


Spacarar:
Recently I tried to run `npm update` on a vue website that uses dc.js and it changed the crossfilter import that is exposed by dc.js.

Here is a screenshot of the updated package list in case this was not a direct result of dc.js being updated

dc version from 3.1.2 -> 3.1.9
d3 version from  5.10.0 -> 5.15.0
crossfilter2 version from 1.4.7 -> 1.5.2 (problem?)
<details>
<summary>Screenshot of package.json diff</summary>

![Screen Shot 2020-02-07 at 3 28 02 PM](https://user-images.githubusercontent.com/24302366/74070635-cd961380-49be-11ea-803d-bdcad6b8e296.png)
</details>

after some quick inspection it appears that the dc.crossfilter wasn't imported correctly and can be fixed by either:

updating the reference on on `dc.js:13925` from `dc.crossfilter = crossfilter` to `dc.crossfilter = crossfilter.default`

or

just using `dc.crossfilter.default(dataArray)` without editing any dc.js source directly

I definitely feel that the workaround listed above isn't the correct way to solve the issue I've come across, but judging by the version numbers, it looks like dc.js should be compatible with the version of d3 and crossfilter being used but may not be imported correctly into dc 3.1.9.

gordonwoodhull:
Hi @Spacarar, sorry for the slow response. 

If you're still around, are you able to force crossfilter2@1.4? We changed the crossfilter2 build system from browserify to rollup and this may have confused your bundler. It could also be the upgrades to your bundler.

Unfortunately there is a fundamental problem here, which is that crossfilter is both a function and a module. This was legal JavaScript (since a function is an object) but it may not be legal ES6; some variants of Typescript allow this and some don't.

When I check the dc.js 3.1.9 examples in a browser, `typeof(dc.crossfilter) === "function"`. So I think this a bundler issue and not a JS issue. (There could be a better way to define the rollup module, I'm not sure.)

See also crossfilter/crossfilter#147: Angular doesn't like the Typescript syntax and once again you end up with a crossfilter module containing a crossfilter function/class.

Reductio also had [a similar problem](https://stackoverflow.com/questions/59960090/how-to-get-reductio-to-work-with-crossfilter2-using-webpacker-in-rails-6) which was solved by [switching from gulp to rollup](https://github.com/crossfilter/reductio/pull/60/files).

[...]

PR #1651, opened 2020-02-08
Mention that `dc.crossfilter` is not exported in v4

kum-deepak:
Issue https://github.com/dc-js/dc.js/issues/1650 made me to take notice of this.


gordonwoodhull:
Thanks @kum-deepak!

Good point. Additionally, as of 3.1.2 DCv3 also doesn't use crossfilter but it still needs to be removed from the AMD scaffolding.

PR #1655, opened 2020-02-18
implement sunburst.relativeRingSizes in dc 3 to provide more flexible…

### LINKED ISSUE #1511: Sunburst chart calculate radius not counted exists of root area and inner radius
Sunburst chart use d3.partition() to calculate the data size. This function reserved first level as root, but this area should not show in sunburst chart. Currently, second level shared this area, so the second level always larger that other level if inner radius is 0. In contrast, if setting inner radius too large, the first few levels cannot show.

So I suggest change the partition size setting to:
```
        var partition = d3.partition()
            // .size([2 * Math.PI, _radius * _radius]);
            .size([2 * Math.PI, 1]);

        partition(hierarchy);
        _rootOffset = hierarchy.y1;
```

And then change radius calculation in buildArcs() to:
```
            .innerRadius(function (d) {
                // return d.data.path && d.data.path.length === 1 ? _innerRadius : Math.sqrt(d.y0);
                return _innerRadius + ((d.y0 - _rootOffset) / (1 - _rootOffset) * (_radius - _innerRadius));
            })
            .outerRadius(function (d) {
                // return Math.sqrt(d.y1);
                return _innerRadius + ((d.y1 - _rootOffset) / (1 - _rootOffset) * (_radius - _innerRadius));
            });
```

the3ver:
… ring sizing; fixes #1511

### DIFF
--- a/spec/helpers/fixtures.js
+++ b/spec/helpers/fixtures.js
@@ -232,3 +232,12 @@ function loadIrisFixture () {
             "5.9,3,5.1,1.8,virginica"
     );
 }
+
+function loadSunburstData3CompleteRings () {
+    return [
+        {"x1":"a","x2":"aa","x3":"aaa","y":1},
+        {"x1":"a","x2":"aa","x3":"aaa","y":2},
+        {"x1":"a","x2":"aa","x3":"aaa","y":3}
+    ];
+}
+
--- a/spec/sunburst-chart-spec.js
+++ b/spec/sunburst-chart-spec.js
@@ -303,4 +303,124 @@ describe('dc.sunburstChart', function () {
         });
     });
 
+    describe('sunburst.relativeRingSizes', function () {
+
+        function buildSunburstChart3CompleteRings(id) {
+            data = crossfilter(loadSunburstData3CompleteRings());
+            var valueDimension = data.dimension(function (d) {
+                return [d.x1, d.x2, d.x3];
+            });
+            valueGroup = valueDimension.group().reduceSum(function (d) {
+                return +d.y;
+            });
+            var div = appendChartID(id);
+            var chart = dc.sunburstChart('#' + id);
+            chart
+                .dimension(valueDimension)
+                .group(valueGroup)
+                .width(width)
+                .height(height)
+                .transitionDuration(0);
+            return chart;
+        }
+
+        function getPieSliceBBoxY(chart, sliceNumber) {
+            return chart.select('.pie-slice._' + sliceNumber).node().getBBox().y;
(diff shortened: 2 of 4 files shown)


gordonwoodhull:
Thanks @the3ver! Very nice and clean way to factor out the changing behavior, and I'm glad it defaults to the old behavior. Thank you so much for writing tests!

Since it is a whole *ring sizes interface* that gets replaced, what do you think about going one step further and opening it up so that people can replace all of it if they want?

I am imagining someone showing up next year saying "ah but there is a third strategy!" The "relative ring sizes" strategy you came up with looks really general, but there is always someone who comes along with another idea. We always try to keep everything open and customizable in dc.js. (It is a *leaky abstraction by design*.)

So how about..?
- make a new public member `chart._defaultRingSizes` with the defaults currently on`_ringSizes`, then initialize `_ringSizes = this._defaultRingSizes`
- make a new public member `_relativeRingSizes` with the stuff that is currently initialized in  `relativeRingSizes`
- new getter/setter `ringSizes` that changes `_ringSizes` instead of the one that just changes the `relativeRingSizesFunction`

If desired, also have a constructor for relativeRingSizes. With ES6 users could just do

```js
chart.ringSizes({...chart._relativeRingSizes, relativeRingSizesFunction: r => { ... }})
```
but with ES5 is it would be the more annoying
```js
chart.ringSizes(Object.assign({},
                              chart._relativeRingSizes,
                              {relativeRingSizesFunction: function(r) { ... }}));

```

The change looks weird in DC3 but it is forward-compatible: in DCv3 there aren't members exposed publicly like this, but in DCv4 all the "private" members are exposed as `this._whatever`.

the3ver:
@gordonwoodhull nice suggestions, thanks, I think this makes the interface clearer and more consistent. Also the dc.sunburstChart.RelativeRingSizes constructor is very useful.

I think you meant _defaultRingSizes (and _relativeRingSizes) to be an object, not a function, now that i look at the code again. Please let me know, if thats the case.

gordonwoodhull:
I like this! It's as powerful as what I was suggesting, but a simpler interface.

As you show, there is no reason to expose the underlying objects, just provide constructors for the different ring sizing strategies. Then if people want to change the behavior they overwrite whatever they want on the strategy object.

I will make a few small changes (e.g. `RelativeRingSizes` is not a class, it's a function that returns an object, so it can be an ordinary method) and then merge for 3.2.0.

Thanks for your contribution!

gordonwoodhull:
Released in 3.2.0 and 4.0.2

PR #1660, opened 2020-03-05
Cleanup redundant code in Sunburst chart

kum-deepak:
While working on some other refactors, I came across this.

Removing this code does not seem to have any impact. Test cases still pass, the charts work. In addition, reading the code makes the believe that the BaseMixin's implementation covers the needs of this chart.

### DIFF
--- a/src/base/base-mixin.js
+++ b/src/base/base-mixin.js
@@ -24,8 +24,10 @@ const _defaultFilterHandler = (dimension, filters) => {
         dimension.filterFunction(d => {
             for (let i = 0; i < filters.length; i++) {
                 const filter = filters[i];
-                if (filter.isFiltered && filter.isFiltered(d)) {
-                    return true;
+                if (filter.isFiltered) {
+                    if(filter.isFiltered(d)) {
+                        return true;
+                    }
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
It's apparently a performance improvement: 601e723c727

Probably to avoid the second case in the loop of the default filter handler:

https://github.com/dc-js/dc.js/blob/a393078a1ef0208c55693a4176dba60daf8d5319/src/base/base-mixin.js#L24-L34

Correct fix for the workaround you propose removing would be

                if (filter.isFiltered) {
                    if(filter.isFiltered(d)) {
                        return true;
                    }
                } else if (filter <= d && filter >= d) {
                    return true;
                }

I don't think we ever want a filter object which has `.isFiltered` and should be second-guessed if it returns false.

kum-deepak:
Thanks! Makes sense.

kum-deepak:
I am working on an experimental PR to refactor filters, dimensions, and groups. I will commit an initial version and future thoughts soon.

gordonwoodhull:
Cool, thanks @kum-deepak!

PR #1662, opened 2020-03-05
fix rounding issue in sunbursts using _chart.equalRingSizes

### LINKED ISSUE #1661: sunburst: rounding problem using equalRingSizes
Using 3.2.0:
* make sunburst with chart.ringSizes(chart.equalRingSizes())
* use data that creates at least 8 rings
=>
dc.errors.BadArgumentException
message: "relativeRingSizes : portions must add up to 1, but sum was 0.9999999999999999"

the3ver:
fixes #1661

### DIFF
--- a/spec/helpers/fixtures.js
+++ b/spec/helpers/fixtures.js
@@ -249,3 +249,11 @@ function loadSunburstData3CompleteRings () {
     ];
 }
 
+function loadSunburstData10CompleteRings () {
+    return [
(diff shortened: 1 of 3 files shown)


gordonwoodhull:
I noticed this problem when reviewing #1655 but then I forgot to say anything about it.

I prefer to fix the issue by being lenient in `assertPortionsArray`:

```js
            if (Math.abs(portionsSum - 1) > dc.constants.NEGLIGIBLE_NUMBER) {
```

since this will come up for other relative ring sizes.

Took your tests, expanded to testing 2-27 equal rings, and applied the above fix for 3.2.1.

Thanks @the3ver!

Issue #1573, opened 2019-09-11
SunBurst and CapMixin


kum-deepak:
Currently SunBurst mixes in CapMixin. I am unsure if it can actually support capping in a logical and meaningful way.

@gordonwoodhull need your advice.

kum-deepak:
Removing CapMixin from SunBurst is not complicated and test cases still pass. I will make it as a commit into charts. We will drop it if needed.

gordonwoodhull:
I am not sure how useful it is, but it works almost as I would expect - it limits the number of top-level slices.

It has a bug where it splits the title "Others" across 6 cells:

<img width="321" alt="image" src="https://user-images.githubusercontent.com/1366709/64738660-ce815280-d4bd-11e9-9385-85d4fc7e1227.png">

I would expect just the innermost indigo slice to be titled Others, with no children. This bug isn't surprising because capped charts always need a lot of special cases to deal with the extra item.

I don't remember anyone complaining about it, so I don't think the feature is getting much use. I think it's mostly there because the author based the sunburst chart on the pie chart. 

I agree, it's safe to remove it, and we can revert it later if needed.

kum-deepak:
Thanks! I have made a commit. I will close this issue.

Issue #584, opened 2014-05-02
chainable .data() function


gordonwoodhull:
Originally, `.data()` was intended for pre-processing or filtering data. It takes a group and returns data in the same format as `group.all()`.

However, this does not work if the chart uses `.data()` internally, because one will override the other.

So the current best way to do this is to create a [fake group](https://github.com/dc-js/dc.js/wiki/FAQ#fake-groups) which wraps the group and implements `.all()` by fetching the data from the group and then returning its modified version of the data. This works fine and groups can be wrapped multiple times, but it's a bit confusing for new users, and it also looks like an ugly hack until you get used to it.

A simple solution would be to have an internal `.data()` and an external `.data()`, but this isn't very elegant and doesn't support the possibility that multiple levels of the chart might want to preprocess the data.

Better to figure out a way to chain them so that (I presume) the client processes the data first, then the base, then the children.

gordonwoodhull:
An alternate solution is to build some kind of data pipelining library that makes "fake groups" easier to write and deals with chaining in a more elegant fashion.

PR #1665, opened 2020-03-07
[Experimental] - refactoring dimensions, groups, and filters - maybe more

kum-deepak:
I have been studying the code for a while now. I have started thinking that it may be possible to extract input data along with associated transformation out of the charts themselves.

The following could be data concerns:

- computation of data (usually done by crossfilter)
- handling of filters (partially done by crossfilter)
- value/key accessors, including likes of radius accessor)
- sorting
- capMixin
- reasons people create fake groups - like hiding empty bars

Most of the charts have very simple interaction with `data concerns`. Charts like Bubble, Sunburst, Tables, Counts have different patterns.

I have been thinking that if we create a series of classes and mixins, each handling a one concern, we can untangle quite a lot. For example for a simple chart if we have a sequence like the following:

- Data adapter - a holder for `dimension` and `group`, key exposed methods `data`, `top` etc.
- FilterMixin - manage and apply filters, key exposed methods, `filter`, `hasFilter`, `isFiltered` etc.
- StandardTransformMixin - apply key/value accessors.
- CapMixin - carry out capping - optional - key methods `cap`

The above should cover the current features. However this will allow modifying any of the layers to create alternative strategies. For example:

- The FilterMixin can be replaced/extended to cover cases that would have needed setting any of the filter handler functions.
- By adding/inserting additional layers it will be possible to sort post capping, remove empty bars etc.

This also leads us to interesting thoughts - different charts have different data needs:

- Many charts need key, value pairs, with optional title, labels etc.
- Bubble chart additionally needs radius accessor
- Some charts like DataCount and NumberDisplay needs much simpler stuff.

If we start by defining alternate implementations for data requirements for these, we can support implementation of interesting ideas like capped BubbleCharts (as these will need special handling for the radius).

Other extreme example would be to create a Capped Sunburst - this might be more involved but possible.


gordonwoodhull:
I am in favor of clarifying the data model somehow. 

The filter mixin is a nice clean refactor and it might be nice to break up the base mixin just for the sake of organization and shortening the file somewhat.

dc.js is built on at least a dozen accessors and they are used in pretty much all of the code that calls D3 - besides key and value there are color, label, title, radius, symbol, etc.

So I'm not sure I see how a standard transform mixin would simplify things. But if you see a way to separate the data side from the chart side, with every chart having different data requirements, that would be a welcome improvement.

As for capping as a data transformation, we do have the `.data()` mechanism, but its flaw is that there can be only one transform this way. As you point out, the cap, stack, and bubble mixins all use this, and they can't currently be combined. Back in #584 I discussed a chainable `.data()` function - but I ended up going with fake groups for all user customizations, because that's already chainable and functionally the same thing.

Also if the interface to `.data()` were changed now, that might break user code. That might be okay in a major version, but we'd have to be careful and issue warnings if possible.

A somewhat hacky way that might work now is if the cap and bubble mixins explicitly wrapped the existing data accessor. Each mixin currently has 

https://github.com/dc-js/dc.js/blob/e391b99281b6c48733395c70a05c4a5489b3e4d3/src/base/cap-mixin.js#L41

but we could expose `baseMixin._data` so the the mixins could chain manually:

```js
  const lastData = this.dataFunction(); // returns baseMixin._data without calling it as .data() does
  this.data(group => {
      const all = lastData(group)
```

It doesn't help with chaining the stack mixin, because that one is pulling data from multiple groups. I suppose if you wanted a capped stacked chart, you'd need capping on each group, and the capping would need to be consistent between all the stacks, returning the same keys.

Capping the sunburst should be really easy and I outlined how to do it in #1573. That doesn't require the kind of changes we are talking about here.

kum-deepak:
Thanks for your detailed response. I am targeting this for next major release for sure.

I have been able to move quite a lot of data related logic into external classes. So far tried Key/Value accessors and Capping. In current approach the transform modifies the data such that following items are there in data:

- `key` - as returned by `group.all` - used for filters
- `value` - as returned by `group.all` - kept for reference (other accessors may use these).
- `_key` - output of `keyAccessor`
- `_value` - output of `valueAccessor`

Currently the `keyAccessor` and `valueAccessor` methods in chart propagate these functions to the data helpers. Charts are using fixed functions for key and value which correspondingly return `_key` and `_value`.

Capping has been implemented as a data transform, which collapses items as one entry, and modifies calls to filter. This has made Capping to be quite transparent to charts themselves. The capped accessors have no longer magical functionality.

Please see if the approach looks alright. Though I was not expecting, so far it is backward compatible.

I see following more work to take it to conclusion:

- Implementing for stacks
- Extracting other accessors
- Handling other chart types
- Cleanup code

gordonwoodhull:
Okay, looks interesting. I took a quick look. 

It definitely helps readability a lot and IIUC it implements chaining data using the class hierarchy.

A couple things I noticed were not backward-compatible:

* we try to avoid writing back to the key/value objects that come from `group.all()`. I'm not sure that I understand the purpose of adding `_key` and `_value`, please explain or maybe I need to read over it again. (It helps chaining somehow?)
* `.data()` used to take a `group => object[]` accessor function which was helpful to users if 

1. one of the chart's mixins hadn't already done this, or
2. they wanted to change what the mixin had done

This is the breakage I referred to above. It's hard to know how much people are using the "data accessor" since it has so many limitations. So it might be okay to remove, I don't know.

kum-deepak:
Yes, I am currently assuming that chaining would be implemented by inheritance. Classes are actually Mixins so that additional classes can be inserted if needed.

I am cloning output of group.all, before changing it. After cleanup the charts will directly use datum._key and datum._value. There is, at least, one case where the chart needs the original key - to set filters.

I am assuming that cases - where data callback was used - can be covered by replacing or extending one of the classes in data hierarchy.

gordonwoodhull:
Cloning does not prevent the original objects from being modified. You’d have to do a deep clone to avoid that. 

I would still like to understand what is improved by copying `key` to `_key` etc. It moves the `keyAccessor`, what other benefits?

src/base/stack-mixin.js
```diff
@@ -15,50 +16,46 @@ export class StackMixin extends CoordinateGridMixin {
 
         this._stackLayout = stack();
 
-        this._stack = [];
+        this._dataProvider = new StackedDataProvider();
```
gordonwoodhull, line 19: Why is the data provider reused by containment, while most other things are reused by inheritance?

src/data/basic-transform-mixin.js
```diff
@@ -0,0 +1,37 @@
+
```
gordonwoodhull, line 1: I've asked this a couple of times, but I'd still like to hear the rationale for transforming data into new fields. This seems complicated to me, but if it makes the data pipeline clearer in some way, I'd like to hear about it.

I will repeat my observation that there are literally dozens of accessors, and it seems messy to me to create new fields for them rather than just calling the accessors.

I guess as long as you are doing this every redraw pass, and there is a clear... what's the word I'm looking for? provenance? pipeline? data flow? then it could make sense. I just want to understand the rationale.

It does seem like chaining `.data()` from parent to child provides a clear flow, so maybe the risk of duplicated and inconsistent data is mitigated,

src/data/cross-filter-stack-adapter.js
```diff
@@ -0,0 +1,49 @@
+import {CrossFilterSimpleAdapter} from './cross-filter-simple-adapter';
+
```
gordonwoodhull, line 2: What is specific to crossfilter about this adapter? It looks like it just calls `group.all()` - I hope we are keeping that convention even when crossfilter is not used?

kum-deepak:
@gordonwoodhull, this PR was just for me to check that it is doable and that eventually, all test cases would pass. I do not think it was necessarily the best design or code to reach the end goal.

After converting to TypeScript I will redo the separation - maybe in smaller tasks. 

In parallel, I will go through your questions and answer the best of my current understanding.

gordonwoodhull:
Got it!

Yes, by all means, do not allow my questions to hold you up. Maybe you can provide a design document with the rationale when the transformation is ready to merge/release.

PR #1669, opened 2020-03-19
Cleanup color-mixin and color accessor

### LINKED ISSUE #1575: Usage of getColor by stack-mixin based charts
The stack-mixin and derived classes set `this` explicitly in addition to needing the `chart` instance inside `getColor`. This will prevent, in the current form, making `getColor` a member method in ColorMixin.

kum-deepak:
Fixes #1575

The color accessor code in StackMixin was bit peculiar. By tracing the code it became clear that ultimately in all cases the name of the layer is getting used as the color accessor. So cleaning up the code to ensure that in every case we pass the entity that has `.name` (as the name of the layer).

Also made few small internal changes:

- Layer name is quite important, so as soon a new unnamed layer is added, a name is generated.
- The layer visibility information is maintained in a separate variable. This makes it more consistent.

With all the above changes, `getColor` has been moved as an instance method.

I do not think it breaks compatibility.

### DIFF
--- a/src/base/stack-mixin.js
+++ b/src/base/stack-mixin.js
@@ -114,6 +114,8 @@ export class StackMixin extends CoordinateGridMixin {
         const layer = {group: group};
         if (typeof name === 'string') {
             layer.name = name;
+        } else {
+            layer.name = String(this._stack.length);
         }
         if (typeof accessor === 'function') {
             layer.accessor = accessor;
--- a/src/base/stack-mixin.js
+++ b/src/base/stack-mixin.js
@@ -19,10 +19,11 @@ export class StackMixin extends CoordinateGridMixin {
         this._titles = {};
 
         this._hidableStacks = false;
+        this._hiddenStacks = {};
         this._evadeDomainFilter = false;
 
         this.data(() => {
-            const layers = this._stack.filter(this._visibility);
+            const layers = this._stack.filter(l => this._visibility(l));
             if (!layers.length) {
                 return [];
             }
@@ -53,12 +54,13 @@ export class StackMixin extends CoordinateGridMixin {
     _prepareValues (layer, layerIdx) {
         const valAccessor = layer.accessor || this.valueAccessor();
         layer.name = String(layer.name || layerIdx);
+        const isLayerHidden = !this._visibility(layer);
         const allValues = layer.group.all().map((d, i) => ({
             x: this.keyAccessor()(d, i),
-            y: layer.hidden ? null : valAccessor(d, i),
+            y: isLayerHidden ? null : valAccessor(d, i),
             data: d,
             layer: layer.name,
-            hidden: layer.hidden
+            hidden: isLayerHidden
         }));
 
         layer.domainValues = allValues.filter(l => this._domainFilter()(l));
@@ -160,10 +162,7 @@ export class StackMixin extends CoordinateGridMixin {
      * @returns {StackMixin}
      */
     hideStack (stackName) {
-        const layer = this._findLayerByName(stackName);
-        if (layer) {
-            layer.hidden = true;
-        }
+        this._hiddenStacks[stackName] = true;
         return this;
     }
 
@@ -174,10 +173,7 @@ export class StackMixin extends CoordinateGridMixin {
      * @returns {StackMixin}
      */
     showStack (stackName) {
-        const layer = this._findLayerByName(stackName);
-        if (layer) {
-            layer.hidden = false;
-        }
+        this._hiddenStacks[stackName] = false;
         return this;
     }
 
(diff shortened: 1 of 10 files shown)


gordonwoodhull:
I like the idea here but I am wary about whether it is going to be backward-compatible.

Sounds crazy but it is probably a whole day's work to review and merge this. So I'm putting it off (but have not forgotten or rejected it!)

Off the top of my head I will need to look at
* every way that the strange `this` is used
* whether ensuring that layers have names has some other side-effect
* all the different ways that colors can be assigned
* whether this changes the precedence of color assignment

src/base/stack-mixin.js
```diff
@@ -45,20 +46,16 @@ export class StackMixin extends CoordinateGridMixin {
...
 
-        this.colorAccessor(function (d) {
-            return this.layer || this.name || d.name || d.layer;
-        });
+        this.colorAccessor(d => d.name);
     }
 
     _prepareValues (layer, layerIdx) {
         const valAccessor = layer.accessor || this.valueAccessor();
-        layer.name = String(layer.name || layerIdx);
```
kum-deepak: Remark 01

src/base/stack-mixin.js
```diff
@@ -45,20 +46,16 @@ export class StackMixin extends CoordinateGridMixin {
             return layers;
         });
 
-        this.colorAccessor(function (d) {
-            return this.layer || this.name || d.name || d.layer;
```
kum-deepak: Remark 02-01

src/base/stack-mixin.js
```diff
@@ -45,20 +46,16 @@ export class StackMixin extends CoordinateGridMixin {
...
 
     _prepareValues (layer, layerIdx) {
         const valAccessor = layer.accessor || this.valueAccessor();
-        layer.name = String(layer.name || layerIdx);
         const allValues = layer.group.all().map((d, i) => ({
             x: this.keyAccessor()(d, i),
-            y: layer.hidden ? null : valAccessor(d, i),
+            y: valAccessor(d, i),
             data: d,
-            layer: layer.name,
```
kum-deepak: Remark 02-02

src/charts/bar-chart.js
```diff
@@ -176,7 +176,7 @@ export class BarChart extends StackMixin {
         const enter = bars.enter()
             .append('rect')
             .attr('class', 'bar')
-            .attr('fill', pluck('data', this.getColor))
```
kum-deepak: Remark 03-01

src/charts/bar-chart.js
```diff
@@ -204,7 +204,7 @@ export class BarChart extends StackMixin {
             })
             .attr('width', this._barWidth)
             .attr('height', d => this._barHeight(d))
-            .attr('fill', pluck('data', this.getColor))
```
kum-deepak: Remark 03-02

kum-deepak:
I will put my leanings which may help your review:

- Before these changes, a name was assigned (see Remark 01) to each layer. However this function was called only for visible layers. Which makes me nervous - depending on which layers are hidden, it might generate same name for more than layer. So, made a scheme of consistent naming.
- 4 clauses in previous version of the accessor (Remark 02-01) referred to the same underlying attribute - name of the layer. See (Remark 02-02) - the `layer.name` is assigned as `datum.layer`. Changing this to `datum.name` gets rid of two of the clauses out of 4. See https://github.com/dc-js/dc.js/pull/1669/commits/3daef4b88fd22244f7487c7eeaa035e870761722
- There were cases (Remarks 03-01 and 03-02) where pluck was called - this was the reason for needing `this` in the accessor. With this change the accesor needed only one clause. See https://github.com/dc-js/dc.js/pull/1669/commits/d2b679c7014cb12a680b575fb043572113a8bc3e
- At this stage, the accessor is ready for same change as others which were done during ES6 conversion.
- So, all of the above makes me believe that there is no change in order or preference (as all were same entity anyway). If someone over-rode it by covering all 4 clauses, it would continue working (the 3 would never match and only one would) - though they can simplify of the want.

I have made small commits, after each commit (I do not remember precisely) all test cases should pass. This should make it easier to trace sequence of changes.

Please let me know if I can add more notes.

kum-deepak:
This PR can go in the next major release. If you feel more appropriate, please create a new branch say `v5` and merge into that.

kum-deepak:
Any luck with this PR?

gordonwoodhull:
Obviously I haven't found the time to review it. It doesn't really help to put it to another release because I still need to understand the impact before merging.

kum-deepak:
Already merged in dc-v5 branch so removing dc-v5 tag.

gordonwoodhull:
This one is on me to review before v5 is released.

kum-deepak:
It has already been merged into the current dc-v5 branch. For dc-v5, before release it would undergo few more refactors, so, I will suggest waiting.

Issue #1498, opened 2018-11-16
textFilterWidget should use redrawGroup, not redrawAll


gordonwoodhull:
Currently when the filter changes, `textFilterWidget` calls

https://github.com/dc-js/dc.js/blob/87cf58e670a34a988c2b7a81452b04f3432780f4/src/text-filter-widget.js#L63

This is incorrect because it redraws the default chart group, which might not be the same chart group where this filter resides. Instead it should call `chart.redrawGroup()` (or supply its own chart group name).

gordonwoodhull:
Fixed in version 4.2.1

PR #1678, opened 2020-04-10
Add chartGroup in redrawAll()

kriddy:
It took me some time to figure out why the text filter wasn't working anymore and then I remembered that I recently added groups to the charts. After some google magic I found https://github.com/dc-js/dc.js/issues/1498 en by changing redrawAll(); to redrawAll(chartGroup); it works again.


gordonwoodhull:
Sorry I missed this. Will put in next release.

gordonwoodhull:
Looks like this was a combination of dc.js 3.* and 4.* code; I changed it to 

```js
                chart.redrawGroup();
```

and release in version 4.2.1. Thanks @kriddy!

PR #1687, opened 2020-06-09
A strange interaction with crossfilter - subtle issue in Bubble Overlay

kum-deepak:
**Edit: now test case has been updated, so easier to review and merge.**

Most charts uses `d3.data`, however, Bubble Overlay uses `d3.datum`.

There are two tests to check if filtering another dimension sets few circles to zero radii:

- https://github.com/dc-js/dc.js/blob/develop/spec/bubble-overlay-spec.js#L129
- https://github.com/dc-js/dc.js/blob/develop/spec/bubble-overlay-spec.js#L142

The test cases pass - but due to a strange coincidence. To demonstrate the issue I have made a change in the test cases which should not have made any difference (https://github.com/dc-js/dc.js/pull/1687/commits/ae21b7a7804a9d7aa10574c93df3099f0c9d8d06). However, with this change the test cases fail.

Underlying issue appears to be the following:

- When we bind using `.datum`, children inherit it.
- When we update `.datum` of the parent node, it reflects in the children.
- However, if we set `.datum` to `undefined` in the parent node, the children retain the older value.
- The fix is simple - instead of setting `.datum` to `undefined` we should set it to `{}` (commit https://github.com/dc-js/dc.js/pull/1687/commits/1759e22c5bc1f2939f7a76cb0c9de37e78599151).

The unmodified code ties the crossfilter objects directly. When the filter is applied the tied objects get updated. This combined with the d3 behavior with `undefined`, makes the test cases succeed.

Not sure since which release the underlying issue appeared - it might be a d3 behavior change in d3v4.

I realize it is lot of explaining for a single line change.

### DIFF
--- a/spec/bubble-overlay-spec.js
+++ b/spec/bubble-overlay-spec.js
@@ -103,17 +103,17 @@ describe('dc.bubbleOverlay', () => {
         });
     });
 
-    function removeEmptyBins (grp) {
+    function cloneAndRemoveEmptyBins (grp) {
         return {
             all: function () {
-                return grp.all().filter(d => d.value !== 0);
+                return grp.all().map(d => Object.assign({}, d)).filter(d => d.value !== 0);
             }
         };
     }
     describe('filtering another dimension', () => {
         let regionDim;
         beforeEach(() => {
-            chart.group(removeEmptyBins(group)).render();
+            chart.group(cloneAndRemoveEmptyBins(group)).render();
             regionDim = data.dimension(d => d.region);
         });
         function expectRadii (expected) {
--- a/src/charts/bubble-overlay.js
+++ b/src/charts/bubble-overlay.js
@@ -145,7 +145,7 @@ export class BubbleOverlay extends BubbleMixin(BaseMixin) {
                 .attr('transform', `translate(${point.x},${point.y})`);
         }
 
-        nodeG.datum(data[point.name]);
+        nodeG.datum(data[point.name] || {});
 
         return nodeG;
     }


gordonwoodhull:
Do all the other charts allow cloning the crossfilter data? This surprises me, because I thought there were a lot of places that rely on the data being live-updated from crossfilter.

I think the actual bug here, if any, is that the bubble overlay chart is initialized with a particular set of points (California etc.) but the test uses `removeEmptyBins()`, which was probably a simple mistake. The bubble overlay is not designed to be used with removeEmptyBins; it expects to always have the same data coming in, with one crossfilter bin for every point.

The bubble overlay chart doesn't do any data joins; it doesn't have a way to add or remove points. Your fix causes it to substitute an empty object for any bins that have gone missing, and then presumably it coerces an undefined radius to zero. That make the test work but it's not really correct since the chart isn't designed to be used this way.

Would be happy to take a PR removing `removeEmptyBins`.

kum-deepak:
This is a fair point that this chart - the data should always include keys corresponding to each point. Should I change the test cases to not use `removeEmptyBins()`?

I have been trying various things with the code. One of them is to use without crossfilter. In our applications we routinely use simple charts, linked charts with cross filter, and linked charts where data is remote. Currently we use different libraries for each. Charts from these three libraries look and behave noticeably different. We are now starting to use dc for standalone charts as well. We achieve this by passing a fake group (with data from the backend) and a fake dimension.

On another note, cloning the cross filter data does not impact any other chart (barring Sunburst that has not been tested yet) - all test cases pass, examples seem to work. So, other than this, I have not come across any case that rely on bound objects getting modified by crossfilter. Are you aware of any?

gordonwoodhull:
Yeah, I think the problem with this test is the use of `removeEmptyBins`. The chart always loops over points, so it doesn't make sense to remove group data.

That's a good point that data from other backends will usually not be modified in place the way crossfilter data is. Maybe I am remembering old issues that have been resolved by now.

I completely agree this should be supported!

kum-deepak:
> Yeah, I think the problem with this test is the use of `removeEmptyBins`. The chart always loops over points, so it doesn't make sense to remove group data.
> 
I will do a PR

> That's a good point that data from other backends will usually not be modified in place the way crossfilter data is. Maybe I am remembering old issues that have been resolved by now.
> 
> I completely agree this should be supported!
> 
> What do you mean that charts look and behave noticeably different? Does it indicate bugs in dc.js?

Because we did not use `dc.js` in other two cases :smile:

Issue #667, opened 2014-08-07
Stack mixin charts' y-domain is always [0, n] when elasticY(true)


mtraynham:
##### Problem

The series chart (which is a composite mixin) draws all stack mixin children (typically line charts) incorrectly when `elasticY` is set to true.  I would expect the highest value to be at the top of the chart and the lowest value to be at the bottom, flush with the axis.

Here is a screen of the issue:
![screenshot from 2014-08-07 15 39 38](https://cloud.githubusercontent.com/assets/5741321/3847999/39ff6d44-1e6b-11e4-84e5-78cfd2a4eac3.png)

Because the series chart (composite mixin) will add multiple child charts, the stack mixin is effectively only rendering n number of single group stacks (where the data point has value `{y: value, y0: 0}`.  The issue arises when accumulating the y-axis min, in the child's `yAxisMin` function of `stack-mixin.js`:

```
    _chart.yAxisMin = function () {
        var min = d3.min(flattenStack(), function (p) {
            // Because y is typically greater than y0
            // (which in our case is 0), this returns 0.
            return (p.y + p.y0 < p.y0) ? (p.y + p.y0) : p.y0;
        });
        return dc.utils.subtract(min, _chart.yAxisPadding());
    };
```
##### Workaround

A workaround is changing the series-chart's `_chartFunction` to properly return the min value for the group, using the `chart` method:

```
var chart = new dc.seriesChart("#chart1", "group1")
      .chart(function (c) {
          var child = dc.lineChart(c);
          dc.override(child, 'yAxisMin', function () {
              var min = d3.min(child.data(), function (layer) {
                  return d3.min(layer.values, function (p) {
                      return p.y + p.y0;
                  });
              });
              return dc.utils.subtract(min, child.yAxisPadding());
          });
          return child;
      });
```

![screenshot from 2014-08-07 15 38 53](https://cloud.githubusercontent.com/assets/5741321/3848238/d77416b8-1e6d-11e4-8a90-2afd72a9ecb0.png)
##### Thoughts

The more I think about it, the less I like the fact that series is a chart type...  It really only applies to line charts.  It would be better if the stackMixin handled these cases, such as:
- a `seriesAccessor` function
- lines (stacked, series) & bars (stacked, grouped)
  - the animation between these seems to be a big missing feature

The list of bugs around stack & series has been growing for a while, so they definitely deserve some attention and it would be nice to provide an all in one true stack/series/group mixin.

gordonwoodhull:
[...]

Issue #661, opened 2014-08-01
Implement .elasticRadius for Bubble Overlay charts


hypsypops:
Circles should resize as data is filtered in other charts using [.elasticRadius([boolean])](https://github.com/dc-js/dc.js/blob/master/web/docs/api-latest.md#elasticradiusboolean) as done in the standard bubble chart.

Seems like it would be simple, but I'm still a little noob-ish and haven't been able to successfully do it myself. :)

Thanks for a great library!

gordonwoodhull:
Implemented in dc.js 2.1.2

PR #1688, opened 2020-06-12
Cleanup specs for BubbleOverlay

kum-deepak:
BubbleOverlay expects data corresponding to each of the points - removing empty bins interferes with this. See discussion in #1687.

For some reason - one of the circles are now different size. The new one seems correct. I am wondering was the old one incorrect?

### DIFF
--- a/spec/bubble-overlay-spec.js
+++ b/spec/bubble-overlay-spec.js
@@ -103,17 +103,9 @@ describe('dc.bubbleOverlay', () => {
         });
     });
 
-    function removeEmptyBins (grp) {
-        return {
-            all: function () {
-                return grp.all().filter(d => d.value !== 0);
-            }
-        };
-    }
     describe('filtering another dimension', () => {
         let regionDim;
         beforeEach(() => {
-            chart.group(removeEmptyBins(group)).render();
             regionDim = data.dimension(d => d.region);
         });
         function expectRadii (expected) {
@@ -142,7 +134,7 @@ describe('dc.bubbleOverlay', () => {
             it('filtering should lock the remaining bubbles to min and max radius sizes', () => {
                 regionDim.filter('Central');
                 dc.redrawAll();
-                expectRadii([0, 0, 0, 10, 60, 0]);
+                expectRadii([0, 0, 0, 45.7, 60, 0]);
             });
         });
     });


gordonwoodhull:
Thanks for pointing out the change in test result. This is why we have tests: not so much because we will immediately recognize what is the right result, but because when behavior changes we want to catch it and figure out why.

The test is for elastic radius. That is, when the chart is drawn, the extent of the input values from [radiusValueAccessor](https://dc-js.github.io/dc.js/docs/html/BubbleMixin.html#radiusValueAccessor) is calculated and used as the domain for the bubble radius scale [r](https://dc-js.github.io/dc.js/docs/html/BubbleMixin.html#r__anchor).

Now I vaguely recall that when I lifted elasticRadius to the bubble mixin in #661, I had to add `removeEmptyBins` to the test, or it wouldn't have the right effect. As you can see, the previous test observed the smaller bubble snapping to 10, the minimum radius. But without `removeEmptyBins`, the still-present zeros are snapped to 10, and the smaller visible bubble is around 45.

So to go to the root of the problem, #661 was not as simple as it looked. In order to implement elastic radius correctly on the bubble overlay chart, since it does not support `removeEmptyBins`, the zeros need to be excluded when calculating the elastic radius.

Interestingly, the bubble mixin already has a special case for zeros when it sets the SVG radius:

https://github.com/dc-js/dc.js/blob/64716dfcce893e0fcf48dd6b6772987b4b23a458/src/base/bubble-mixin.js#L114-L121

It passes the value into the radius scale, and then uses 0 if the result is NaN or the value is 0.

To fix this, `rMin()` needs to have a similar special case, removing the zeros before computing the minimum:

https://github.com/dc-js/dc.js/blob/64716dfcce893e0fcf48dd6b6772987b4b23a458/src/base/bubble-mixin.js#L106-L108

Something like 

```js
return min(this.data().map(this.radiusValueAccessor()).filter(value => value > 0))
```

gordonwoodhull:
Tested the change and reverted the change to the test!

Thanks @kum-deepak!

gordonwoodhull:
Rats, this broke the iris tests of the bubble chart.

I guess either choice might be valid, so I'm adding an option called `excludeElasticZero` and defaulting it true. As noted above, I think it is more consistent to exclude zeros when calculating elastic, because any zeros in the data are special cased to cause the radius to be zero.

However, there might be reasons why people like the old behavior, so a flag is the safest thing to do.

By analogy, consider #667, where the current behavior is to always include zero in elasticY, but the issue asks to optionally _not_ include zero, occasionally a valid choice.

kum-deepak:
Wow! So many things going on while I thought it was a simple one :smile:

gordonwoodhull:
Yeah, it seems to be the rule for charting libraries: features will interact in complex ways. 😔

PR #1691, opened 2020-06-27
Tighten eslint config further by removing unused globals

kum-deepak:
Further to https://github.com/dc-js/dc.js/commit/902736ad4436dbcad1d57badce14a5e485c59d7a


gordonwoodhull:
Ah, fantastic, yes I wondered if these were needed.

No globals for the core library! As it should be!

PR #1699, opened 2020-07-02
Typescript - crude conversion

kum-deepak:
Target is for the src files to have .ts extension and test cases should pass.


kum-deepak:
Figured out following pattern for chainable accessors (thanks to d3 typings):

```
    public valueAccessor (); // TODO: define ValueAccessorType
    public valueAccessor (valueAccessor): this;
    public valueAccessor (valueAccessor?) {
        if (!arguments.length) {
            return this._valueAccessor;
        }
        this._valueAccessor = valueAccessor;
        return this;
    }
```

kum-deepak:
Also noticed the typing for `d3-dispatch` is incorrect - our usage shows type error, most likely example on https://github.com/d3/d3-dispatch will fail as well. I will recheck and raise an issue in the relevant project.

kum-deepak:
As a syntactic choice, I am marking public methods and properties explicitly as `public`. It is the default and not needed, however, I like things to be explicit.

In this pass, I am making as little change as possible, once the entire code base is converted, deeper changes will follow.

gordonwoodhull:
People do use the private methods as well - dc.js is a leaky abstraction by design, and that's one of the way to do extra customization.

kum-deepak:
The generated code has all private members as well. So, these should work. If they tried to use type library they will get errors, which can be overridden.

gordonwoodhull:
I'm going to reserve judgment about Typescript for now. It's verbose, but I can see how it could be helpful for the heavy refactoring you are planning to do.

I tried debugging the output in Chrome, Firefox, and Safari. I was not too surprised to find that the source map works in Chrome and Firefox, but not in Safari. I have encountered this before. 

Gladly, Safari mostly works like (an older) Chrome, since they both came from WebKit. Hopefully we won't have too many Safari-specific issues to debug.

Looks like the Web Inspector tries and fails to load the sources:

image

gordonwoodhull:
Re d3-dispatch, yes D3 is notoriously difficult to type strongly. I'm sure that you will have to contribute to Definitely Typed if you want to get every type perfect.

kum-deepak:
> I'm going to reserve judgment about Typescript for now. It's verbose, but I can see how it could be helpful for the heavy refactoring you are planning to do.
> 
> I tried debugging the output in Chrome, Firefox, and Safari. I was not too surprised to find that the source map works in Chrome and Firefox, but not in Safari. I have encountered this before.
> 
> Gladly, Safari mostly works like (an older) Chrome, since they both came from WebKit. Hopefully we won't have too many Safari-specific issues to debug.
> 
> Looks like the Web Inspector tries and fails to load the sources:
> 
> image

Great to know. If we want to debug in Safari - we can try inlining the source map which might work.

kum-deepak:
> Re d3-dispatch, yes D3 is notoriously difficult to type strongly. I'm sure that you will have to contribute to Definitely Typed if you want to get every type perfect.

D3 magically works across many types - which the type library may not consider.

I have opened https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45869 for the dispatch typing.

Meanwhile, I have also found another issue with d3selection, the value function as per the typing must return a string. We already have one case in d3box that returns a number. I am getting ready to raise more PRs for the typings as we go along.

kum-deepak:
Crude conversion is complete - test cases still passing. Will merge it and start a new PR.

PR #1702, opened 2020-07-05
tslint - initial configuration

kum-deepak:
Start with a loose configuration and keep tightening as more refactoring is carried out.

This PR will cover only basic configuration and fixes.

### DIFF
--- a/src/charts/composite-chart.ts
+++ b/src/charts/composite-chart.ts
@@ -146,7 +146,10 @@ export class CompositeChart extends CoordinateGridMixin {
     }
 
     _calculateYAxisRanges (left, right) {
-        let lyAxisMin, lyAxisMax, ryAxisMin, ryAxisMax;
+        let lyAxisMin;
+        let lyAxisMax;
+        let ryAxisMin;
+        let ryAxisMax;
         let ranges;
 
         if (left) {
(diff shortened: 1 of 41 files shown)


src/charts/composite-chart.ts
```diff
@@ -146,7 +146,10 @@ export class CompositeChart extends CoordinateGridMixin {
     }
 
     _calculateYAxisRanges (left, right) {
-        let lyAxisMin, lyAxisMax, ryAxisMin, ryAxisMax;
+        let lyAxisMin;
+        let lyAxisMax;
+        let ryAxisMin;
+        let ryAxisMax;
         let ranges;
```
gordonwoodhull, line 153: This seems pretty extreme - is declaring multiple variables in a single statement really so problematic?

gordonwoodhull: I'd prefer to avoid needless verbosity - I guess [one-variable-per-declaration](https://palantir.github.io/tslint/rules/one-variable-per-declaration/) is default true, would prefer false.

kum-deepak: Very soon there will be type information for each of the variable declarations. In that case, it may make it easier to read and refer. That would be my only argument, if at all.

I will set it to false and undo this commit. If we change our opinion we can always change in the future.

gordonwoodhull: Oh... I get it. Never mind, I get it. I have seen this style guideline for C++ for the same reason. Sorry for the noise!

kum-deepak: Thanks, I will keep it for now.

kum-deepak:
I am done for this iteration. I will wait for a while before merging it.

Issue #1703, opened 2020-07-05
Explore Mixins - again


kum-deepak:
Investigate if we can do Mixins in a more Typescript friendly way. 

Initial lead: https://www.stevefenton.co.uk/2017/08/typescript-mixins-part-two/

kum-deepak:
The previous way of Mixins worked well for ES6. However, it was not enough for Typescript. 

- We need proper type checking in Mixins themselves - the compiler needs to figure out the methods a Mixin is using from possible Bases.
- We need proper type checking in derived classes, the compiler needs to figure out all combined methods.

There is some issue around generating declaration for anonymous classes. There are known workarounds based on https://www.typescriptlang.org/docs/handbook/mixins.html. One such workaround is at https://github.com/microsoft/TypeScript/issues/17744#issuecomment-468748859

After refactoring, we may have fewer Mixins. At that stage, the workaround would be implemented.

gordonwoodhull:
I took a look at these `defineProperty` loops. I guess it's a form of metaprogramming, generating the class by running a program.

It's too bad that it does not use language constructs, but I guess it's fairly clear what is going on. In a way it is a step back toward dc@3 in that it builds the classes dynamically instead of using classes and inheritance. 😉

kum-deepak:
Yes, it feels really sad having to go back.

Our usage is quite specific. So, I guess that we can try some simpler approaches as well. I will get back to it after a few rounds of refactoring.

PR #1704, opened 2020-07-07
Mixins - Typescript way - Take 1

kum-deepak:
Ref #1703

### DIFF
--- a/src/base/color-mixin.ts
+++ b/src/base/color-mixin.ts
@@ -5,6 +5,8 @@ import {max, min} from 'd3-array';
 import {config} from '../core/config';
 import {utils} from '../core/utils';
 import {ColorCommonInstance} from 'd3-color';
+import {BaseMixin} from './base-mixin';
+import {Constructor} from '../core/types';
 
 /**
  * The Color Mixin is an abstract chart functional class providing universal coloring support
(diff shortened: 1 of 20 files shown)


kum-deepak:
The entire code now supports basic type checking. All test cases pass. This phase seems done.

src/base/color-mixin.ts
```diff
@@ -14,174 +16,176 @@ import {ColorCommonInstance} from 'd3-color';
...
-    public _colorCalculator;
+export function ColorMixin<TBase extends Constructor<BaseMixin>>(Base: TBase) {
+    return class extends Base {
+        public _colors;
+        public _colorAccessor: (d, i) => any;
+        public _colorCalculator;
 
-    constructor () {
-        super();
+        constructor(...args: any[]) {
```
gordonwoodhull, line 25: does it take args?

src/charts/series-chart.ts
```diff
@@ -1,4 +1,4 @@
-import {ascending} from 'd3-array';
+import {ascending, Primitive} from 'd3-array';
```
gordonwoodhull, line 1: Primitive doesn't look like an export of d3-array?

kum-deepak: It is exported from @types/d3-array.

src/base/color-mixin.ts
kum-deepak: Typescript was giving error. I am planning to relook.

kum-deepak: It seems it is needed otherwise it gives following error:

```
error TS2545: A mixin class must have a constructor with a single rest parameter of type 'any[]'.
```

gordonwoodhull:
I'm confused. I didn't think that mixins were a feature of the Typescript language. These are templated functions that return a class, right?

Is it going by the name of the function?

kum-deepak:
I also did not expect the error message to be so specific. No idea how it is picking it. :smile:

PR #1709, opened 2020-07-12
Initial take at compat system. Skeleton code with partially converted DataCount

kum-deepak:
This is a sample PR demonstrating a way to shift compat code outside of the main code.

- The structure of the folder `src/compat` will allow moving compat code there.
- In the compat folder, all but `data-count.ts` are no-op pass-throughs.
- Please check the way compatibility code from `src/charts/data-count.ts` has been shifted to `src/compat/charts/data-count.ts`.

Currently, the UMD output `dist/dc.js` is built from the compat code. Two UMD options would be built - one with compat and one without.

Once this approach is more stable, the test cases would be split as well.


gordonwoodhull:
It seems like a useful system. My only concern is that it seems geared toward eventually removing the compatibility layer, while in most cases in dc.js we preserve backward compatibility unless we cannot.

Does the system run into issues if compatibility is preserved for multiple versions?

(Note: I’m on vacation through 7/21; I may be slow to review stuff.)

kum-deepak:
We can preserve as much as we want and can. There will be some cases that would make it very clumsy, otherwise no limitations.

Sometimes the augmentations would be in layers - we will see examples soon.

Noted your vacation details. I will be working in an interleaved manner on the following three:

- Moving code out to the compat layer.
- Adding more type information.
- Protoyping and working towards conf system.

kum-deepak:
I will close this PR and start a new one - long running.

PR #1711, opened 2020-07-15
Moving towards better typing

kum-deepak:
No functional changes. Figuring out type information. Occasionally minor code rearrangement to make typing work.


kum-deepak:
No rush for reviews, enjoy your vacation, I will let you know when I am done with this.

kum-deepak:
I have reached a logical stage. This now has basic typing in place. Now I will move onto refactors. With each refactor, the typings will also keep getting better.

Now I understand the d3 typings much better - after factoring out data related stuff, we can go for stricter d3 typings.

@gordonwoodhull, please review if you have time.

src/core/types.ts
```diff
@@ -1,3 +1,130 @@
...
+    range (): RangeType[];
+    range (newRange: [RangeType, RangeType]): this;
+}
+
+export type MinimalColorScale = MinimalScaleWithRange<any, string>;
+
+export type MinimalRadiusScale = MinimalScaleWithRange<any, number>;
+
+// Specifically created for scales in Coordinate Grid charts
+export interface MinimalXYScale extends MinimalScaleWithRange<any, number> {
```
gordonwoodhull, line 69: I wonder if it would make sense to separate the ordinal case from the continuous case, since we use very different parts of the interface. I am only thinking this because this is a lot of `?` to compensate for different usage, not sure if it would actually be helpful.

src/core/utils.ts
```diff
@@ -62,38 +62,38 @@ export const pluck = function (n, f?) {
  * @type {{}}
  */
 export const utils: IUtils = {
```
gordonwoodhull, line 64: These utils are confusing - do they still need to be filled in?

gordonwoodhull: Oh, I guess they get replaced below. I guess these dummies are a way of giving type information to the methods below?

src/base/base-mixin.ts
```diff
@@ -71,12 +81,12 @@ const _defaultResetFilterHandler = filters => [];
...
 export class BaseMixin {
     // tslint:disable-next-line:variable-name
-    private __dcFlag__: number;
-    private _dimension; // TODO: create an interface for what dc needs
-    private _group; // TODO: create an interface for what dc needs
-    private _anchor; // TODO: figure out actual type
-    private _root; // TODO: figure out actual type
-    private _svg;
+    private __dcFlag__: string;
+    private _dimension: MinimalCFDimension;
```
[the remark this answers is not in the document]
kum-deepak: Dimensions and Groups, I am planning to relook several times before final release. One of the major targets is to clearly identify what data charts need (https://github.com/dc-js/dc.js/issues/1718) and providing those using data providers.

src/core/types.ts
```diff
@@ -1,3 +1,130 @@
...
+    legendToggle: (d: LegendSpecs) => void;
+    legendReset: (d: LegendSpecs) => void;
+    legendHighlight: (d: LegendSpecs) => void;
+    filters; // function, TODO: signature
+    legendables: () => LegendSpecs[];
+
+    svg (): Selection<SVGElement, any, any, any>;
+}
+
+export interface LegendSpecs {
```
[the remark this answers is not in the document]
kum-deepak: Will change it.

src/core/types.ts
kum-deepak: Scales - especially when it comes to CoordinateGrid poses challenges. It is not just Ordinal scales, even `scaleBand` is significantly different. Currently, I have taken an approach that let us do better than what we have. After some more refactoring, we might see if we can use Generic type for x-scale.

No wonder the guys doing typings for d3 have created to no base class for a scale, they have mostly enumerated all the scales.

Once we do generics for the chart data, we can do a little better for scales.

I do realize that Ordinal charts behave differently than others and I am sure this has come up quite few times for discussion.

src/core/utils.ts
kum-deepak: This a temporary structure to get started, this pattern will totally go. In cases where functions are independent, these will be exposed as top-level (with backward compatibility). In other cases when they are related they will be rearranged to avoid this pattern.

PR #1725, opened 2020-07-24
Sunburst cleanup

kum-deepak:
Making `this.ringSizes().relativeRingSizes` to `this._relativeRingSizes` and `this.ringSizes().rootOffset` to `this._rootOffset`. This avoid unnecessarily modifying a user supplied object.


kum-deepak:
Localized impact, so, merging it now. Can reverse if needed.

gordonwoodhull:
Agree, this bothered me too.

PR #1726, opened 2020-07-25
Inline `pluck` and move it to compat layer

kum-deepak:
We had discussed this when doing ES6 conversion. Finally did it :smile:. Changes span multiple files, however, the pattern is similar. 

In three places 2 argument version was used, it was first fetching the title accessor for the stack and then using it to get the actual title (Line and Bar charts).

The original function has been moved to the compat layer.

### DIFF
--- a/src/base/base-mixin.ts
+++ b/src/base/base-mixin.ts
@@ -2,7 +2,7 @@ import {BaseType, select, Selection} from 'd3-selection';
 import {dispatch, Dispatch} from 'd3-dispatch';
 import {ascending} from 'd3-array';
 
-import {pluck, utils} from '../core/utils';
+import {utils} from '../core/utils';
 import {instanceOfChart} from '../core/core';
 import {deregisterChart, redrawAll, registerChart, renderAll} from '../core/chart-registry';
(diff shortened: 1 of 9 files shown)


kum-deepak:
I am unclear, so, need your inputs. For the two-argument version, which one you like better:

```typescript
            dot.append('title').text( (d, i) => this.title(d.name)(d.data, i));
```

or

```typescript
            dot.append('title').text( (d, i) => {
                const titleAccessor = this.title(d.name);
                return titleAccessor(d.data, i);
            });
```

kum-deepak:
I realized creating `_extTitleAccessor` in `StackMixin` gives best of both options.

kum-deepak:
I am ready to merge this.

gordonwoodhull:
Otherwise this looks like a nice improvement. Just want to be careful.

kum-deepak:
Valid points. Currently, I find it is alright to keep the two-argument version as it was. I have a sense that this code may be impacted by the data provider concept that I am planning to work on. After that, if needed, I will come back to this.

Issue #1717, opened 2020-07-18
Refactor: `_inputType` in CboxMenu


kum-deepak:
Currently, there is a member `_inputType` in CboxMenu. This should ideally be a local variable in `_renderOptions()`. No user-visible changes.

PR #1727, opened 2020-07-25
Make `inputType` a local variable in `_renderOptions`

kum-deepak:
Ref #1717


kum-deepak:
Localized impact, so, merging it now. Can reverse if needed.

gordonwoodhull:
Agree, it is completely dependent on `.multiple()` so there's no need to duplicate the state.

On the rare chance anyone was using this private, undocumented field, they can look at `multiple` instead.

Issue #1119, opened 2016-03-17
releasing charts from the chart registry


tlrobinson:
If you have a single-page application that renders charts, throws them away, renders more charts, etc (as we do over at https://github.com/metabase/metabase) it appears `dc.chartRegistry` holds onto a reference to every chart ever rendered by dc.js forever (and thus the data used by every chart as well).

This seems unfortunate. I think it would be better to explicitly create a chart group object that you add charts to (or not, if you don't need that functionality).

In the meantime, is the recommended solution to just call `dc.chartRegistry.deregister(chart)` if you know you're done with a chart?

Or if you don't use chart groups at all is it safe to just do `setTimeout(() => dc.chartRegistry.clear(), 10000)` or something?

XaserAcheron:
Thinking on it, I may have run into this as well. I certainly ran into a similar-sounding slowdown case in my SPA which I seem to have fixed by calling `dc.chartRegistry.clear` and `dispose()`-ing all of the `crossfilter` dimensions.

Would be nice to have an in-library solution, of course.

gordonwoodhull:
Do you have any suggestions for how this could be done automatically? It seems to me that the client creates the charts and only the client will know when they should be destroyed.

I doubt that you don't use the chart groups, because the chart groups are used to implement reaction to filters.

XaserAcheron:
I was thinking of suggesting something along the lines of "remove the chart if another one is defined in the same chart group with the same ID", but then I got to thinking: dc doesn't (AFAIK) have a concept of "chart ID" other than its parent selector (maybe). Can you put more than one chart on the same parent? I've never tried.

nordfjord:
Most SPA frameworks have some lifecycle hooks. A common one would be an event when a component is removed from the DOM.

just do something like this for your chart component:

```js
component.onunload = function(){
  dc.chartRegistry.deregister(component.chart, component.chart.chartGroup());
  component.chart.dimension().dispose();
}
```

@gordonwoodhull a way to reliably detect whether a chart is indeed active would be something along the lines of:

```
_chart.redraw = function() {
  if (!document.contains(_chart.anchor())) {
    dc.chartRegistry.deregister(_chart, _chart.chartGroup());
  }
}
```

Which is really expensive.

XaserAcheron:
[...]

Issue #988, opened 2015-08-16
generalize & spin off the chart registry


gordonwoodhull:
The proposal is to create a separate library (under this dc-js org or crossfilter org, though potentially it can be used without either), which takes care of the communication between charts that is currently handled by the undocumented (#676) `dc.chartRegistry`

It would probably use `d3.dispatch` (or a copy of it from the new modular d3 v4, since this has relevance without d3) to keep the array of listeners. So render and redraw become just events.

Also there are at least two common inter-chart events, various forms of brushing-and-linking that are not brushing-and filtering. One is just linked selection, where the charts have the same dimension or at least compatible dimensions (#681). The other is range/focus charts; in general there will be more than one focus chart (#820).

The more complicated idea is to abstract what a filter is, basically move `dc.filters` out from here, because many people use dc.js with a backend that is not crossfilter. Then there would be something that "renders" queries from those filters. (And we might hit #478 while we're at it.)

gordonwoodhull:
I made half an attempt here: [chart.registry.js](https://github.com/dc-js/dc.graph.js/blob/develop/chart.registry.js). 

It works okay but it is verbose. I came to the conclusion that d3.dispatch is probably too low-level a tool for this, since really we want a whole interface of methods (some of them optional) on each chart.

Maybe that is fine, since it would be better for the chart registry to be free of dependencies.

gordonwoodhull:
Note this solution by @Fil and @pierreleripoll: https://observablehq.com/@pierreleripoll/vegasync

It's a step in the right direction because it isn't bound to any particular charting library. 

On the implementation level, I have two critiques:
* d3-dispatch is somewhat verbose for this purpose, since you have to bind every method (render/redraw/etc is really an interface)
* The solution should not be bound to crossfilter, since the query provider might be something else (Apache Arrow, backend server, etc).

However it's a good starting point for discussion.

kum-deepak:
I think in the dc-v5 branch, the ChartGroup has achieved the goals of this task. It is generic enough that non-dc charts can interact as well. This combined with FilterStorage can work with Crosfilter - without having a hard dependency on Crossfilter. This works with remote data sources. I think we can close it.

gordonwoodhull:
Yes. The only part that is still missing is the “spin off” part. 

At the time I had the idea that the chart registry should be a separate library so that people could connect charts in a standard way without using dc.js at all.

[...]

PR #1729, opened 2020-07-28
Concept of ChartGroup

kum-deepak:
A simple and localized factoring out ChartGroup class. This class holds all information related to a chart group.

@gordonwoodhull I have a few questions which you might have answers to:

1.  We have two different ways to check if two charts are same, any idea why?
https://github.com/dc-js/dc.js/blob/82469c9c9cdc27bdca1d5b1b4f1e8ac79c41f9d9/src/core/chart-registry.js#L70
https://github.com/dc-js/dc.js/blob/82469c9c9cdc27bdca1d5b1b4f1e8ac79c41f9d9/src/core/chart-registry.js#L40
2. There is a call `has` (https://github.com/dc-js/dc.js/blob/develop/src/core/chart-registry.js#L138). Any reason this does not take a group as an argument?
3. The `deregisterAllCharts` behaves differently `*All` functions that it works across groups when no group is passed, is that intentional?

### DIFF
--- a/src/core/chart-group-types.ts
+++ b/src/core/chart-group-types.ts
@@ -1,0 +1,23 @@
+export interface IMinimalChart {
+    render(): void;
+
+    redraw(): void;
+
+    filterAll(): void;
+
+    focus?(): void;
+}
+
+export interface IChartGroup {
+    register(chart: IMinimalChart): void;
+
+    deregister(chart: IMinimalChart): void;
+
+    renderAll(): void;
+
+    redrawAll(): void;
+
+    filterAll(): void;
+
+    refocusAll(): void;
+}
--- a/src/core/chart-group-types.ts
+++ b/src/core/chart-group-types.ts
@@ -1,23 +1,23 @@
 export interface IMinimalChart {
-    render(): void;
+    render (): void;
 
-    redraw(): void;
+    redraw (): void;
 
-    filterAll(): void;
+    filterAll (): void;
 
-    focus?(): void;
+    focus? (): void;
 }
 
 export interface IChartGroup {
-    register(chart: IMinimalChart): void;
+    register (chart: IMinimalChart): void;
 
-    deregister(chart: IMinimalChart): void;
+    deregister (chart: IMinimalChart): void;
 
-    renderAll(): void;
+    renderAll (): void;
 
-    redrawAll(): void;
+    redrawAll (): void;
 
-    filterAll(): void;
+    filterAll (): void;
 
-    refocusAll(): void;
+    refocusAll (): void;
 }
--- a/src/core/chart-group-types.ts
+++ b/src/core/chart-group-types.ts
@@ -1,23 +1,15 @@
 export interface IMinimalChart {
     render (): void;
-
     redraw (): void;
-
     filterAll (): void;
-
     focus? (): void;
 }
 
 export interface IChartGroup {
     register (chart: IMinimalChart): void;
-
     deregister (chart: IMinimalChart): void;
-
     renderAll (): void;
-
     redrawAll (): void;
-
     filterAll (): void;
-
     refocusAll (): void;
 }
--- a/src/core/chart-group.ts
+++ b/src/core/chart-group.ts
@@ -12,7 +12,7 @@ export class ChartGroup implements IChartGroup {
     }
 
     public has (chart: IMinimalChart): boolean {
-        return this._charts.indexOf(chart) >= 0;
+        return this._charts.includes(chart);
     }
 
     public register (chart: IMinimalChart): void {
--- a/src/core/chart-group.ts
+++ b/src/core/chart-group.ts
@@ -1,0 +1,55 @@
+import {IChartGroup, IMinimalChart} from './chart-group-types';
+
+export class ChartGroup implements IChartGroup {
+    private _charts: IMinimalChart[];
+
+    constructor() {
+        this._charts = [];
+    }
+
+    public list(): IMinimalChart[] {
+        return this._charts;
+    }
+
+    public has(chart: IMinimalChart): boolean {
+        return this._charts.indexOf(chart) >= 0;
+    }
+
+    register(chart: IMinimalChart): void {
+        this._charts.push(chart);
+    }
+
+    deregister(chart: IMinimalChart): void {
+        this._charts = this._charts.filter(ch => ch !== chart);
+    }
+
+    public clear(): void {
+        this._charts = [];
+    }
+
+    renderAll(): void {
+        for (const chart of this._charts) {
+            chart.render();
+        }
+    }
+
+    redrawAll(): void {
+        for (const chart of this._charts) {
+            chart.redraw();
+        }
+    }
+
+    filterAll(): void {
+        for (const chart of this._charts) {
+            chart.filterAll();
+        }
+    }
+
+    refocusAll(): void {
+        for (const chart of this._charts) {
+            if (chart.focus) {
+                chart.focus();
+            }
+        }
+    }
+}
--- a/src/core/chart-group.ts
+++ b/src/core/chart-group.ts
@@ -3,49 +3,49 @@ import {IChartGroup, IMinimalChart} from './chart-group-types';
 export class ChartGroup implements IChartGroup {
     private _charts: IMinimalChart[];
 
-    constructor() {
+    constructor () {
         this._charts = [];
     }
 
-    public list(): IMinimalChart[] {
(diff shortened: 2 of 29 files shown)


gordonwoodhull:
1. I think the only reason is multiple authors. `has` / `hasChart` dates back to the original chart registry commit 55119930e by Nick Zhu; `deregisterChart` was written by Neal Ruggles and arrived a year and a half later in fc88bfab3f / 7d1d663ff.
I have to guess that Neal was being cautious and thought that object identity might not be trustworthy. (In JS it is.) I can't think of a reason why you would remove a chart by a different handle. It would be nice to get rid of the anchorName one, since this is the only one that returns a value; I believe the rest of the required interface for charts is pure pub/sub.
2. Considering that `has` / `hasChart` are not used internally, it's hard to guess what the intention was. We could add the parameter. A tricky bit is that the usual way of specifying the default group (by not mentioning it) would not work. However, a user could pass `dc.constants.DEFAULT_CHART_GROUP` to check the default group. We would need to document that.
3. It's inconsistent but yes, I think it's intentional, and sorta intuitive. I wouldn't want to break the interface. Again, a user would have to pass `dc.constants.DEFAULT_CHART_GROUP` to clear just the default group.

gordonwoodhull:
This is a nice change in itself. I am wondering if there is anything we can do in this area to help with #988.

Specifically I would like dc.js to (optionally) depend on a third-party pub/sub system, preferably one that has interfaces, since I found using d3-dispatch too messy. Then charts from different libraries could interact through crossfilter and this third-party library, instead of having to register external charts in the dc.js chart registry.

Not sure what would help here - interfaces for chart registry and chart group? Chart registry is just one global but it's a doozy.

kum-deepak:
Please also see #1119 (reference for myslef)

kum-deepak:
Thanks @gordonwoodhull 

>     1. I think the only reason is multiple authors. `has` / `hasChart` dates back to the original chart registry commit [5511993](https://github.com/dc-js/dc.js/commit/55119930e4693a00c07c8b26e2f1cc00397aaad7) by Nick Zhu; `deregisterChart` was written by Neal Ruggles and arrived a year and a half later in fc88bfab3f / [7d1d663](https://github.com/dc-js/dc.js/commit/7d1d663ffd19e1a77c0ca4bba8946b9a1f81baf7).
>        I have to guess that Neal was being cautious and thought that object identity might not be trustworthy. (In JS it is.) I can't think of a reason why you would remove a chart by a different handle. It would be nice to get rid of the anchorName one, since this is the only one that returns a value; I believe the rest of the required interface for charts is pure pub/sub.
> 
I will switch to comparing just object references.


>     2. Considering that `has` / `hasChart` are not used internally, it's hard to guess what the intention was. We could add the parameter. A tricky bit is that the usual way of specifying the default group (by not mentioning it) would not work. However, a user could pass `dc.constants.DEFAULT_CHART_GROUP` to check the default group. We would need to document that.
> 
We will see at the end of the refactor cycle and document accordingly.

>     3. It's inconsistent but yes, I think it's intentional, and sorta intuitive. I wouldn't want to break the interface. Again, a user would have to pass `dc.constants.DEFAULT_CHART_GROUP` to clear just the default group.
It makes sense, it is definitely useful to SPAs.

kum-deepak:
Giving a `string` as chart group limits flexibility in regards to replacing it with an alternate. At this stage, I am planning to allow an object to be passed as the chart group (backward compatibility will be maintained). Since these will be documented as interfaces, it should be possible to write adaptors for other systems.

This also offers an interesting way of using the Group. An SPA can create a Chart Group as a local object and connect all charts to it. This will allow the charts to be garbage collected when all these go out of scope.

kum-deepak:
I have completed my work for this iteration. In the last commit https://github.com/dc-js/dc.js/pull/1729/commits/eb833892300c99ed53016c3279c9c94ff8d860aa, the stock sample has been updated to work with an explicit group. We can omit this commit before merging.

@gordonwoodhull, I will move based on your feedback.

src/core/chart-group-types.ts
```diff
@@ -0,0 +1,23 @@
+export interface IMinimalChart {
+    render (): void;
+
+    redraw (): void;
+
+    filterAll (): void;
+
+    focus? (): void;
+}
```
gordonwoodhull, line 9: Just a style nit, but do interfaces need to be double spaced? The gaps help readability when methods have content, but are they helpful if there are just one line declarations?

src/core/chart-group.ts
```diff
@@ -0,0 +1,55 @@
...
+    constructor () {
+        this._charts = [];
+    }
+
+    public list (): IMinimalChart[] {
+        return this._charts;
+    }
+
+    public has (chart: IMinimalChart): boolean {
+        return this._charts.indexOf(chart) >= 0;
```
gordonwoodhull, line 15: Could use [.includes()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes)

src/core/chart-registry.ts
```diff
@@ -39,53 +40,32 @@ class ChartRegistry {
...
+            if (this._chartMap[group]) {
+                this._chartMap[group].clear();
+                delete (this._chartMap)[group];
+            }
         } else {
+            for (const chartGroupName in this._chartMap) {
+                if (this._chartMap.hasOwnProperty(chartGroupName)) {
+                    this._chartMap[chartGroupName].clear();
+                }
+            }
```
gordonwoodhull, line 68: We've had occasional complaints about memory leaks when deregistering charts; this careful approach may help. 

On the other hand, this PR introduces a cycle, because the chart will hold a reference to the chart group. I guess that's why you are being careful!

Might be worth testing. Not sure if it could be automated, but we could at least have a page in the examples that creates and removes a lot of charts in a loop, and use dev tools to confirm it's not leaking.

src/core/chart-registry.ts
```diff
@@ -96,9 +76,8 @@ class ChartRegistry {
      * @param {String} [group] Group name
      * @returns {Array<Object>}
      */
-    public list (group: string): BaseMixin[] {
-        group = this._initializeChartGroup(group);
-        return (this._chartMap)[group];
+    public list (group?: string): BaseMixin[] {
+        return this.chartGroup(group).list() as BaseMixin[];
```
gordonwoodhull, line 80: Does this convert correctly? `.list()` returns `IMinimalChart[]`

This method is probably unused internally.

src/core/chart-registry.ts
```diff
@@ -173,15 +149,8 @@ export function filterAll (group: string): void {
...
-export function refocusAll (group: string): void {
-    const charts = chartRegistry.list(group);
-    for (let i = 0; i < charts.length; ++i) {
-        // @ts-ignore
-        if (charts[i].focus) {
-            // @ts-ignore
-            charts[i].focus();
-        }
-    }
+export function refocusAll (group?: string): void {
```
gordonwoodhull, line 152: Another function which is unused internally; I forgot it existed! 

We should improve our test coverage sometime.

kum-deepak: I will see that we can add coverage test, it will good to know.

kum-deepak: Created #1731

src/core/chart-registry.ts
kum-deepak: Yes, it should be changed.

src/core/chart-group.ts
kum-deepak: Thanks, still getting used to all Array functions :smile:

src/core/chart-group-types.ts
kum-deepak: Changed in https://github.com/dc-js/dc.js/pull/1729/commits/395967eb7773a95d4efa6e34b959a33333fe9e39

src/core/chart-group.ts
kum-deepak: Changed in https://github.com/dc-js/dc.js/pull/1729/commits/395967eb7773a95d4efa6e34b959a33333fe9e39

src/core/chart-registry.ts
kum-deepak: Changed in https://github.com/dc-js/dc.js/pull/1729/commits/d8d308b55745bcb72db6860613ed1807f5316c94

src/core/chart-registry.ts
kum-deepak: Created a task https://github.com/dc-js/dc.js/issues/1732

PR #1730, opened 2020-07-31
Make Utils functions as top level

kum-deepak:
This is a fairly mechanical conversion. Backward compatibility has been maintained via the compat layer.

While doing the conversion I realized that the function `utils.constant` can be inlined easily (this has been put as a separate commit).


kum-deepak:
I am merging this. Feedbacks I will incorporate as needed.

gordonwoodhull:
LGTM! :+1:

Issue #1735, opened 2020-08-04
Policy discussion: do we allow conf settings with side effects


kum-deepak:
The majority of chart properties exactly do get and set. However, there are a few properties that do additional work.

Going forward, with the new conf system, do we allow properties with side effects?

Ref #1707.

kum-deepak:
If we allow properties with side effects we will insist that properties are always set with a call to `.configure`. This method will have specific checks and code for side effects.

If we do not allow properties with a side effect, we will decide on a case-to-case basis for the updated code. In this case, we can additionally allow setting properties by a direct assignment.

gordonwoodhull:
I'm confused what we do with the current code that has side effects, if we decide to ban them?

kum-deepak:
I think I can see the way forward, please see https://github.com/dc-js/dc.js/pull/1736#discussion_r471228908.

PR #1736, opened 2020-08-04
Creating Conf system with backward compatibility through compat layer

kum-deepak:
The following has been achieved:

- A boilerplate structure with mixin classes in the compat layer to allow compat code to be slotted into appropriate places. This has been done in a way that still allows type checking capabilities of the typescript compiler to work.
- Conf interfaces that extend the parent. BaseMixin accepts it as a generic parameter. This has been done in a clever way that allows each Mixin code to type-check with the appropriate conf interface. For example, `this._conf` within `ColorMixin` code will support all properties from conf for `BaseMixin` and `ColorMixin`. Similarly for conf in `PieChart`.
- It also lays out how the code needs to be moved to the compat layer.

As the next step, all simple properties (i.e., without side effects) will be moved as per this structure.

Once we have taken a decision in #1735, we will be ready to migrate the rest of the properties.

Reaching the current code structure was quite a lot of pain and angst. However, the end product that allows type safe way of working is reward enough.


gordonwoodhull:
It will take me a couple days to review this and #1735. 

Out tomorrow for my birthday. :tada:

kum-deepak:
Completed the first round of work for `BaseMixin`. Two properties - `group` and `title` have been not covered in this round - these have been inherited in `StackMixin` with altered semantics. These will be converted after working on `StackMixin`.

kum-deepak:
As it turns out, interfaces allow fields to be declared readonly. This is helping in ensuring that no configuration item is set unintentionally within dc and hinder conf sharing across charts.

Started with CoordinateGridMixin conf and will do across all confs.

kum-deepak:
Completed this round of work. There are some more properties that can move to the conf system after other planned refactors. Those will be taken up as PRs later.

This PR is ready from my end.

src/compat/base/cap-mixin.ts
```diff
@@ -1 +1,136 @@
...
+         * 2.1.5, the ordering defaults to sorting from greatest to least like `group.top(N)` did.
+         *
+         * If you want to cap by one ordering but sort by another, you can still do this by
+         * specifying your own {@link BaseMixin#data `.data()`} callback. For details, see the example
+         * {@link https://dc-js.github.io/dc.js/examples/cap-and-sort-differently.html Cap and Sort Differently}.
+         * @memberof CapMixin
+         * @instance
+         * @param {Number} [count=Infinity]
+         * @returns {Number|CapMixin}
+         */
```
gordonwoodhull, line 43: One consequence of this refactor is that most of the getter-setter documentation is moved to `compat/`

Do you have a long-term vision of where this documentation should go? Obviously it's not in `compat/` so that it can be removed! 😆

src/base/coordinate-grid-mixin.ts
```diff
@@ -34,47 +35,35 @@ const DEFAULT_AXIS_LABEL_PADDING = 12;
...
     private _lastXDomain: [number, number];
     private _y: MinimalXYScale;
     private _yAxis: Axis<any>;  // TODO: can we do better
-    private _yAxisPadding: number;
-    private _yElasticity: boolean;
     private _yAxisLabel: string;
     private _yAxisLabelPadding: number;
     private _brush: BrushBehavior<unknown>;
     private _gBrush: SVGGElementSelection;
     private _brushOn: boolean;
```
gordonwoodhull, line 56: It looks somewhat arbitrary which members got moved to the config here.

I am clear that stuff that is generated by the chart and belongs to the chart state, e.g. the brush, zoom, and `_parentBrushOn`, stay in the chart.

But there are a log of user-configurable things which were left behind, like `x`/`yAxisLabel` & padding, the x and y scales. Does this have to do with side effects #1735 and internal modifications #1739?

src/compat/base/cap-mixin.ts
kum-deepak: I have some idea, the first step would be to try and put the documentation in Conf interfaces. If still, it is not good enough, we can create a table, which will have all settings, explanations, and which charts support it.

I am planning to do this after a couple of more refactors, and when we have switched to tsdoc.

src/charts/series-chart.ts
```diff
@@ -43,23 +36,27 @@ export class SeriesChart extends CompositeChart {
...
     }
 
     public configure(conf: ISeriesChartConf) {
         super.configure(conf);
+
+        // TODO: This is defensive, looking at the code - 'seriesAccessor', 'seriesSort', 'valueSort' do not need it
+        if (['chartFunction', 'seriesAccessor', 'seriesSort', 'valueSort'].some(opt => opt in conf)) {
+            this._resetChildren();
+        }
     }
```
kum-deepak, line 60: Experimentation with simple side effect.

src/base/coordinate-grid-mixin.ts
kum-deepak: I think some cases are simple, where there is a simple side effect. I have experimented with SeriesChart in a similar case. See my comment in the respective commit.

In some cases where the properties are treated as an internal state, these must remain out of conf - for example, scales.

In no case modification of a conf property is allowed, as it will restrict sharing of conf across charts.

Few cases, like margins, are modified in complex ways in Coordinate charts, and even more complicated sharing through Composites. These cases would need refactoring before moving to conf. These would be taken on a case to case basis.

kum-deepak:
After incorporating #1742 I will merge this PR. After moving to integrate Prettier, I will take on more refactors before coming back to conf again.

src/base/coordinate-grid-mixin.ts
gordonwoodhull: Got it. Makes sense! 👍

Issue #1185, opened 2016-07-26
Accessibility


eenblam:
EDIT: It's possible that this could work out to be a small PR that just adds a few ARIA labels. If it seems like it would instead require breaking changes, I can look at providing these features as a fork or as a plugin.

I think dc.js could be made a bit more accessible. I have some ideas on where to start, and I'm willing to do the work. However, I'm just starting to get my feet wet in front-end, visualization, and accessibility, so I'll probably have to ping for help every now and then. At present, I'm trying to get a good look at the problem surface. Orca has been crashing for me on Linux, so I'm currently testing things with NVDA (with multiple browsers) on Windows.
## Motivation

A dashboard is just a tree of some div's and svg. Adding in some ARIA attributes would make the SVG generated by different chart objects accessible to keyboard-only users (think limited range of motion) and users with screen readers.

I expect this will be straightforward for pie, bubble, row, and bar charts. I don't see a way to provide similar functionality to line and area charts. However, just getting this sorted out in the former case would be a big win for both visualization and accessibility, as dc.js would provide a write-once, accessible interface. _(I'm not aware of this functionality elsewhere, but I'd be happy to learn that another project has already implemented something similar!)_
## Requirements
1. The chart must be easily traversed with the keyboard (I suspect switch interfaces should be able to benefit from this functionality without any additional work, but I haven't tried this yet.)
2. The underlying data from which the chart is generated must be available and clearly identified.
3. Presentational nodes, in particular those with `.axis`, should be obscured with `aria-hidden="true"`.

Item 1 is where I'm having the most trouble getting browsers to be consistent with their accessibility API.  See [this table on HTML5 accessibility](http://html5accessibility.com/) across browsers. This [accessible implementation of jQuery UI](http://hanshillen.github.io/jqtest/#goto_tree) is a good reference for the desired functionality, but their implementation has a lot of additional DOM manipulation going on as keyboard focus changes. That would probably be better suited as a plug-in for dc.js, and I would like to avoid that if possible.

Item 2 is partially solved by the existing `title` tags, which provide the data as `label: value`. The task, then, is to direct the SR to this tag. `aria-labelledby` might work here, but I'd like to avoid generating a unique id for each title.

Item 3 seems self-explanatory. Hide stuff that only provides visual cues to sighted users from the screen reader and keyboard, and make the data more explicitly available on keyboard focus (see 2).
## Bonus Points

[...]

PR #1738, opened 2020-08-05
Support for adding accessibility features at chart generation

gherka:
Currently, to add accessibility features to dc.js charts you need to do it as part of post-processing. It would be easier to simply specify what features you'd like to include together with other chart attributes, like width and height and labels. This idea has been brought up before (#1185) and after raising it again on the mailing list recently, @gordonwoodhull encouraged me to try to implement something.

I'm thinking of a series of commits / PRs to cover the basics:

- This one makes all SVG charts accessible by keyboard using tabindex. It also adds a description that screen readers can access. The configuration of a `<desc>` element and `aria-labelledby` attribute seems to have the best support according to [this](https://www.deque.com/blog/creating-accessible-svgs/). I tried adding `<title>`, but it fails a few tests looking at tooltips and besides, some screen readers will read it twice. It defaults to an empty string and ideally, I'd like it to reflect the default chart type - so a `dc.PieChart` will have "pie chart" in the description and so on. This will likely require modifying each chart file with some default attribute, however. Users can add a description for their charts by simply passing a string to `.svgDescription()`.

- Make internal chart elements, like `<circle>` in a bubble chart or `rect` in a bar chart keyboard-selectable. Also, make them "clickable" from keyboard.

- When I tested tooltips created using `<title>` in Narrator / NVDA, I got mixed results for a variety of browsers. In some, the tooltip text was read out, on others it wasn't. There is probably a way to make it a little bit more consistent.

- Optionally designating the `NumberDisplay` widget as `aria-live` [region](https://developer.paciellogroup.com/blog/2014/03/screen-reader-support-aria-live-regions/), meaning changes to it will be read out by screen readers.

Here's the summary of this PR as it currently stands:

- [x] root SVG can get focus and users can add custom description using `svgDescription` method

- [x] internal chart elements can get focus and filtering interactions will trigger on `Enter` or `Space` as if clicked

- [x] CSS will behave on focus in the same way as it behaves on hover

- [x] Individual chart types are accessible  

  - [x] Bar chart
  - [x] Box plot
  - [x] Bubble chart
    - [x] DOM order of bubbles can be changed based on x value
  - [x] Choropleth
  - [x] Heatmap
  - [x] Line chart
  - [x] Pie chart
  - [x] Row chart
  - [x] Scatter plot
    - [x] DOM order of bubbles can be changed based on x value
  - [x] Sunburst chart
 <html><br></html>

- [x] NumberDisplay can optionally be set as `aria-live` region

- [x] Legends can get focus and interactivity

- [x] Add example

### DIFF
--- a/spec/bar-chart-spec.js
+++ b/spec/bar-chart-spec.js
@@ -1280,12 +1280,80 @@ describe('dc.BarChart', () => {
 
     describe('accessibility bar chart', () => {
 
-        beforeEach(() => {
-            chart.render();
-        });
+        function removeEmptyBins (sourceGroup) {
+            return {
+                all:function () {
+                    return sourceGroup.all().filter( d => d.value !== 0);
+                }
+            };
+        }
 
         it('default description should match class name', () => {
+            chart.render()
             expect(chart.svg().node().firstChild.innerHTML).toEqual('BarChart');
+        });
+
+        it('internal elements are focusable by keyboard', () => {
+            chart.keyboardAccessible(true);
+            chart.render();
+            chart.selectAll('rect.bar').each(function () {
+                const bar = d3.select(this);
+                expect(bar.attr('tabindex')).toEqual('0');
+            });
+        });
+
+        it('initial internal elements are clickable by pressing enter', () => {
+            chart.keyboardAccessible(true);
+            const clickHandlerSpy = jasmine.createSpy();
+            chart.onClick = clickHandlerSpy;
+            chart.render();
+          
+            const event = new Event('keydown');
+            event.keyCode = 13;
+                     
+            chart.selectAll('rect.bar').each(function () {
+                const bar = d3.select(this).node();
+                bar.dispatchEvent(event);
+                expect(clickHandlerSpy).toHaveBeenCalled();            
+            });
+        });
+
+        it('newly added internal elements are also clickable from keyboard', () => {
+            chart.keyboardAccessible(true);
+            const clickHandlerSpy = jasmine.createSpy();
+            chart.onClick = clickHandlerSpy;
+
+            const event = new Event('keydown');
+            event.keyCode = 13;
+
+            const stateDimension = data.dimension(d => d.state);
+            const regionDimension = data.dimension(d => d.region);
+            const stateGroup = stateDimension.group().reduceSum(d => +d.value);
+            const ordinalDomainValues = ['California', 'Colorado', 'Delaware', 'Ontario', 'Mississippi', 'Oklahoma'];
+            
+            chart
+                .dimension(stateDimension)
+                .group(removeEmptyBins(stateGroup))
+                .xUnits(dc.units.ordinal)
+                .x(d3.scaleBand().domain(ordinalDomainValues))
+                .elasticX(true)
+                .barPadding(0)
+                .outerPadding(0.1)
+                .transitionDuration(0)
+                .render();
+
+            //force creation of new <rect> elements on existing chart
+            regionDimension.filterExact('West');
+            chart.redraw();
+            regionDimension.filterAll();
+            chart.redraw();
+                    
+            chart.selectAll('rect.bar').each(function () {
+                const bar = d3.select(this).node();
+                bar.dispatchEvent(event);
+                expect(clickHandlerSpy).toHaveBeenCalled();            
+            });
+
         });
 
     });
--- a/spec/bar-chart-spec.js
+++ b/spec/bar-chart-spec.js
@@ -1288,11 +1288,6 @@ describe('dc.BarChart', () => {
             };
         }
 
-        it('default description should match class name', () => {
-            chart.render()
-            expect(chart.svg().node().firstChild.innerHTML).toEqual('BarChart');
-        });
-
         it('internal elements are focusable by keyboard', () => {
             chart.keyboardAccessible(true);
             chart.render();
--- a/spec/bar-chart-spec.js
+++ b/spec/bar-chart-spec.js
@@ -1278,6 +1278,18 @@ describe('dc.BarChart', () => {
         }
     });
 
+    describe('accessibility bar chart', () => {
+
+        beforeEach(() => {
+            chart.render();
+        });
+
+        it('default description should match class name', () => {
+            expect(chart.svg().node().firstChild.innerHTML).toEqual('BarChart');
+        });
+
+    });
+
     function nthStack (n) {
         const stack = d3.select(chart.selectAll('.stack').nodes()[n]);
 
--- a/spec/bar-chart-spec.js
+++ b/spec/bar-chart-spec.js
@@ -1306,10 +1306,10 @@ describe('dc.BarChart', () => {
             const event = new Event('keydown');
             event.keyCode = 13;
                      
-            chart.selectAll('rect.bar').each(function () {
-                const bar = d3.select(this).node();
-                bar.dispatchEvent(event);
-                expect(clickHandlerSpy).toHaveBeenCalled();            
+            chart.selectAll('rect.bar').each(function (d) {
+                this.dispatchEvent(event);
+                expect(clickHandlerSpy).toHaveBeenCalledWith(d);
+                clickHandlerSpy.calls.reset();    
             });
         });
 
@@ -1343,10 +1343,10 @@ describe('dc.BarChart', () => {
             regionDimension.filterAll();
             chart.redraw();
                     
-            chart.selectAll('rect.bar').each(function () {
-                const bar = d3.select(this).node();
-                bar.dispatchEvent(event);
-                expect(clickHandlerSpy).toHaveBeenCalled();            
+            chart.selectAll('rect.bar').each(function (d) {
+                this.dispatchEvent(event);
+                expect(clickHandlerSpy).toHaveBeenCalledWith(d);
+                clickHandlerSpy.calls.reset();
             });
 
         });
--- a/src/base/base-mixin.js
+++ b/src/base/base-mixin.js
@@ -719,14 +719,14 @@ export class BaseMixin {
         // called from each chart module's render and redraw methods
         this._svg.selectAll('.dc-tabbable')
                 .attr('tabindex', 0)
-                .on('keydown', d => {
+                .on('keydown', (d, i) => {
                     // trigger only if d is an object undestood by KeyAccessor()
                     if (d3Event.keyCode === 13 && typeof d === 'object') {
-                        onClickFunction.call(this, d, ...onClickArgs)
+                        onClickFunction.call(this, d, i, ...onClickArgs)
                     } 
                     // special case for space key press - prevent scrolling
                     if (d3Event.keyCode === 32 && typeof d === 'object') {
-                        onClickFunction.call(this, d, ...onClickArgs)
+                        onClickFunction.call(this, d, i,  ...onClickArgs)
                         d3Event.preventDefault();                
                     }
                 
--- a/src/base/base-mixin.js
+++ b/src/base/base-mixin.js
@@ -72,7 +72,7 @@ const _defaultResetFilterHandler = filters => [];
 export class BaseMixin {
     constructor () {
         this.__dcFlag__ = utils.uniqueId();
-        this._svgDescription = ''
+        this._svgDescription = this.constructor.name || '';
 
         this._dimension = undefined;
         this._group = undefined;
--- a/src/base/base-mixin.js
+++ b/src/base/base-mixin.js
@@ -72,6 +72,7 @@ const _defaultResetFilterHandler = filters => [];
 export class BaseMixin {
     constructor () {
         this.__dcFlag__ = utils.uniqueId();
+        this._svgDescription = ''
 
         this._dimension = undefined;
         this._group = undefined;
@@ -521,8 +522,30 @@ export class BaseMixin {
 
     generateSvg () {
         this._svg = this.root().append('svg');
+
+        this._svg.append('desc')
+            .attr('id', `desc-id-${this.__dcFlag__}`)
+            .html(`${this._svgDescription}`);
+
+        this._svg
+            .attr('tabindex', '0')
+            .attr('role', 'img')
+            .attr('aria-labelledby', `desc-id-${this.__dcFlag__}`);
         this.sizeSvg();
         return this._svg;
+    }
+
+    /**
+     * Set description text for the entire SVG graphic to be read out by assistive technologies.
+     * @param {String} [description='']
+     * @returns {String|BaseMixin}
+     */
+    svgDescription (description) {
+        if (!arguments.length) {
+            return this._svgDescription;
+        }
+        this._svgDescription = description;
+        return this;
     }
 
     /**
--- a/src/base/base-mixin.js
+++ b/src/base/base-mixin.js
@@ -1,4 +1,4 @@
-import {select} from 'd3-selection';
+import {select, event as d3Event} from 'd3-selection';
 import {dispatch} from 'd3-dispatch';
 import {ascending} from 'd3-array';
 
@@ -73,6 +73,7 @@ export class BaseMixin {
     constructor () {
         this.__dcFlag__ = utils.uniqueId();
         this._svgDescription = this.constructor.name || '';
+        this._keyboardAccessible = false;
 
         this._dimension = undefined;
         this._group = undefined;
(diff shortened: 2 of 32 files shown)


gordonwoodhull:
Haven’t tested yet, but it looks good!

I think you can default `svgDescription` to something reasonable with `this.constructor.name`

https://stackoverflow.com/a/10314492

gherka:
Good idea! I've added it and set a basic test up on a bar chart. I also checked other chart examples (pie, line, row, composite) and it works as expected.

gherka:
To make significant internal chart elements like bars in a bar chart accessible, I've added a dc prefixed class in the chart's module, but since the effects of this feature are the same across all chart types (`tabindex=0`, triggers `onClick` on keypress), the function is part of `base-mixin.js`. This way, if a new chart type comes along, its creator would only need to think about which elements are "significant" and simply add the `dc-tabbable` class to them. 

Also, elements created at the initial render can disappear and new ones appear after filtering - so the new function would need to run on both `render()` and `redraw()`. 

@gordonwoodhull, if you approve of this approach, I'll go through all existing chart types and amend them in the same way as I did with the bar chart, and write some tests.

src/base/base-mixin.js
```diff
@@ -1,4 +1,4 @@
-import {select} from 'd3-selection';
+import {select, event as d3Event} from 'd3-selection';
```
gherka, line 1: Had to alias "event" because it's used as a function parameter elsewhere.

gordonwoodhull:
I hope to take a look at this today, but I’m out sick, so please bear with me if it takes a little longer.

gherka:
Of course, no worries at all. Hope you get better soon!

src/base/base-mixin.js
```diff
@@ -521,10 +523,45 @@ export class BaseMixin {
 
     generateSvg () {
         this._svg = this.root().append('svg');
+
+        this._svg.append('desc')
+            .attr('id', `desc-id-${this.__dcFlag__}`)
+            .html(`${this._svgDescription}`);
+
+        this._svg
+            .attr('tabindex', '0')
```
gordonwoodhull, line 532: Should this also be enabled/disabled by `keyboardAccessible`?

My concern is that it will probably change the behavior of existing applications.

src/base/base-mixin.js
```diff
@@ -663,11 +700,34 @@ export class BaseMixin {
...
         return result;
     }
 
+    _makeKeyboardAccessible () {
+        this._svg.selectAll('.dc-tabbable')
+                .attr('tabindex', 0)
+                .on('keydown', d => {
+
+                    if (d3Event.keyCode === 13) {
+                        console.log('clicked!');
```
gordonwoodhull, line 718: did you mean to check in the debug logs?

src/base/base-mixin.js
```diff
@@ -521,10 +523,45 @@ export class BaseMixin {
...
+        this._svg
+            .attr('tabindex', '0')
+            .attr('role', 'img')
+            .attr('aria-labelledby', `desc-id-${this.__dcFlag__}`);
         this.sizeSvg();
         return this._svg;
     }
 
+    /**
+     * Set description text for the entire SVG graphic to be read out by assistive technologies.
```
gordonwoodhull, line 540: Please document the default here (in prose)

src/base/base-mixin.js
gherka: Sorry, my fault for not explaining the reasoning. I think it's useful to be able to turn the accessibility of internal chart elements on and off separately from the accessibility of the chart as a whole - depending on chart design, tabbing through many, many elements might not be desirable. `keyboardAccessible` is the method for the former and rolls together focusing and clicking. `svgDescription` is for the latter and covers focus and description. Since description is only useful on focused elements, I included `tabindex` by default.

The only scenario I can think of when this would result in an unexpected change is if the application changed `tabindex` to a value other than 0 which isn't [recommended](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex). I think as is, it's a "free" improvement for existing dashboards that mouse users shouldnt even notice.

gordonwoodhull: I think I get the design. What I meant is, what if someone has an existing dashboard with form elements, but the charts are not currently accessible using the tab key?

I think it's pretty common to have just a text entry field and a couple of selects, or whatever. I know I am lazy and I use the keyboard instead of the mouse in a familiar application, when I have to do something repetitive.

In other words, not all keyboard users are screen reader users.

With this change, the charts will be tabbable even though there isn't any interactivity or screen reader enabled for them. My concern is that it will make these simple form/dashboard applications harder to navigate, even though no functionality has been added.

Here's a very simple example: [complex reductions example](https://dc-js.github.io/dc.js/examples/complex-reduce.html). Before the change, just the links and the radio button group are accessible with the tab (on Chrome - on Firefox the links are not tabbable). After the change, the charts are tabbable but there isn't any functionality when they are focused.

gherka: Thanks for taking the time to explain. I've made a few changes (and tests) so that we don't add any new tabbable elements unless the user explicitly tells us to by either setting `svgDescription` or `keyboardAccessible`. The latter will keep the class constructor name as default description, unless it's overwritten by `svgDescription`. The implementation feels a bit hacky, though.

src/base/base-mixin.js
```diff
@@ -521,10 +523,59 @@ export class BaseMixin {
...
+     * @returns {String|BaseMixin}
+     */
+    svgDescription () {
+
+        // overwrite for public API
+        this.svgDescription = description => {
+            this.svgDescription.userSet = true
+            this._svgDescription = description;
+            return this;
+        };
```
gordonwoodhull, line 556: This seems too complicated. When you call the method, it replaces itself with another method, that, when used as a setter, sets a flag on the function object? And then call it on initialization?

Why not do the following:

* Initialize `_svgDescription` to null
* When `.svgDescription()` is called as a getter, return `_svgDescription || this.constructor.name
* When called as a setter, assign to `_svgDescription` as usual
* When testing if it's been set, use `_svgDescription`, but when getting the value, call `.svgDescription()`

gherka: Well, this is embarrassing :) Was over-thinking it and saw a problem where there wasn't one, really.

gordonwoodhull:
Looks great, thank you!

gherka:
Had to modify the approach a little bit because the way `on('click', fn)` function is specified varies between concrete chart implementations - could be a straightforward `onClick(d)`, or `onClick(d, layerIndex)` or `_onClick(d)` or `onClick(d, i)` or just an anonymous function. Now in addition to adding a `dc-tabbable` class to the relevant elements, you also need to tell `_makeKeyboardAccessible` what to do on click.

A couple of other observations:
-  the DOM order of bubble chart circles is based on the key - what do you think of making it based on the x-value instead so that tab order goes predictably from left to right? We'd need to preserve the option of putting smaller-radius circles on top with `sortBubbleSize`
- since BubbleOverlay requires "external" SVG to function, `generateSVG` from BaseMixin isn't called so svgDescription isn't added, even if called. What's the best way to signpost this behaviour?

gordonwoodhull:
Hi @gherka, sorry for the slow reply.

Yes, I think that ordering the bubble chart circles by X value would be okay, as long as `sortBubbleSize` is still respected. I don't think it makes any difference except for Z and tab order.

That's an interesting question about BubbleOverlay. I don't think this chart is used very much, so it might not matter too much, but I suppose you could do the annotation on render (or `_initOverlayG`) instead? I know it's ugly to have a special case for this one chart, but that's just what is, a single chart that doesn't create its own `svg`.

gherka:
I think this now covers everything in the initial scope of features. There's just a couple of things I'm not too sure about - 
- What tests do I need to add? Some, like ensuring that `NumberDisplay` doesn't have any transitions when `keyboardAccessible` is `true` is quite tricky and haven't been able to figure out how to do it yet.
- What examples do I need to add? While developing, I've tested each feature on the `web-src` examples so they can all potentially work.
- What about resolving conflicts? Do I need to do it in my PR?

src/base/base-mixin.js
```diff
@@ -668,6 +715,28 @@ export class BaseMixin {
...
+    _makeKeyboardAccessible (onClickFunction, ...onClickArgs) {
+        // called from each chart module's render and redraw methods
+        const tabElements = this._svg
+            .selectAll('.dc-tabbable')
+            .attr('tabindex', 0);
+                
+        if (onClickFunction) {
+            tabElements.on('keydown', (d, i) => {
+                // trigger only if d is an object undestood by KeyAccessor()
+                if (d3Event.keyCode === 13 && typeof d === 'object') {
```
gordonwoodhull, line 727: Maybe test if the key accessor returns truthy value? Just to be careful.

src/base/base-mixin.js
```diff
@@ -668,6 +715,28 @@ export class BaseMixin {
...
     }
 
+    _makeKeyboardAccessible (onClickFunction, ...onClickArgs) {
+        // called from each chart module's render and redraw methods
+        const tabElements = this._svg
+            .selectAll('.dc-tabbable')
+            .attr('tabindex', 0);
+                
+        if (onClickFunction) {
+            tabElements.on('keydown', (d, i) => {
```
gordonwoodhull, line 725: Since you started this PR, D3 was updated to 6.0 and the signature of event handlers [changed](https://github.com/d3/d3/blob/master/CHANGES.md#changes-in-d3-60) - now the signature is `(event, d)`

We built an adapter for dc@4.1 called `adaptHandler` that smooths over the difference, to move to d3@6 but still support d3@5. I would suggest using it here.

gordonwoodhull (changes requested):
One small suggestion, and a bigger change to update to dc@4.1.

gordonwoodhull:
Hi @gherka!

Sorry for the slow reply - I am on vacation this week and next.

* We don't generally test animated transitions, so I think it is okay to skip that. The specific tests that you made for edge cases are great! It would also be good to test the general case for each chart, i.e. that each element has been modified. This will help to prevent regressions in the future when other people work on the code.
* I think it would be sufficient to add one example with a few charts - for example, maybe adding accessibility to the [filtering example](https://dc-js.github.io/dc.js/examples/filtering.html) - so that people can try the feature out.
* Yes, please rebase your branch onto develop and then force-push the result to resolve conflicts. The PR will update automatically.

Unfortunately, there was a big change in dc@4.1 which affects your PR. As I mentioned above, the signature of event handlers changed in d3@6, and we are using an adapter to continue support for dc@5 until we decide to drop it. Please use this adapter and test your code with d3@5 and d3@6 to make sure it functions correctly. We are no longer testing d3@5, so I'll take your word on this.

Thanks for your careful work on this feature - people have asked for this for a long time, and it was something that worried me personally. It's good to see that accessibility can be added without much disruption to other functionality.

gherka:
Hi @gordonwoodhull, thank you for your feedback! In this commit I've added automatic tests for each chart type that I had modified and also tested each chart type on the standard examples using v5 and v6. As you suggested, I've included `keyboardAccessible()` and `svgDescription()` in the `filtering.html` example. 

There are a few minor things that can be tweaked further, but the base functionality is now in place, I think.

src/base/base-mixin.js
gherka: I tried testing for it (`this.keyAccessor()(d)`), but it failed for bar and pie charts and `this.keyAccessor()` is always truthy because it returns a function. Is there another way?

spec/bar-chart-spec.js
```diff
@@ -1278,6 +1278,81 @@ describe('dc.BarChart', () => {
...
+            regionDimension.filterExact('West');
+            chart.redraw();
+            regionDimension.filterAll();
+            chart.redraw();
+                    
+            chart.selectAll('rect.bar').each(function () {
+                const bar = d3.select(this).node();
+                bar.dispatchEvent(event);
+                expect(clickHandlerSpy).toHaveBeenCalled();            
+            });
```
gordonwoodhull, line 1350:
```suggestion
            chart.selectAll('rect.bar').each(function (d) {
                this.dispatchEvent(event);
                expect(clickHandlerSpy).toHaveBeenCalledWith(d);
            });

```
1. It's better to check that the correct thing has been clicked, rather than just checking if the spy has been called. Also,  `toHaveBeenCalled` will always be true after the first iteration, because the spy is keeping a count of all calls since it was created.
2. `const bar = d3.select(this).node();` is the same as saying `const bar = this;`, and the fact that `this` is the bar should be apparent from the `selectAll` above.

gordonwoodhull (changes requested):
Fantastic work. Love to see thorough tests, and I finally tried it out via the new example and everything works flawlessly, as far as I can tell. The only thing I see you have not tested is newly added elements on other charts, but I think it's okay to skip that, given how tedious that is to test.

This is almost ready to merge, but I'd like to see the tests tightened up a little, and also the example clarified, as mentioned above.

src/base/base-mixin.js
gordonwoodhull: I agree, there probably is no better way. I am not sure that checking `typeof d === 'object'` provides all that much protection, since objects are pretty common in JS.

Were you seeing errors if you didn't check this?

spec/bar-chart-spec.js
gordonwoodhull: Hmmm, thinking about this further, `toHaveBeenCalledWith` will check all calls ever. We 

want to see that the last call equals the datum, so I guess we need 

```js
expect(obj.method.calls.mostRecent().args[0]).toEqual(d)
```
to be precise.

gordonwoodhull: I got it to work like this

```suggestion
            const clickHandlerSpy = jasmine.createSpy();
            chart.onClick = clickHandlerSpy;
            chart.selectAll('rect.bar').each(function (d) {
                this.dispatchEvent(event);
                expect(clickHandlerSpy).toHaveBeenCalledWith(d);
                clickHandlerSpy.calls.reset();
            });
```

Seems excessively complicated, but this clearly checks that the particular event caused the particular datum to get selected.

src/base/base-mixin.js
gherka: I thought you were on holiday! :) Yes. on charts like boxplot where you have a group with bound data and a clickable circle element. If you press enter on the circle, the event triggers, then propagates up to the group. If we don't check for `d` type, the `onClick` function will try to run twice and the first call (from circle) will error out with `TypeError: filters[0] is undefined`.

src/charts/number-display.js
```diff
@@ -121,36 +122,65 @@ export class NumberDisplay extends BaseMixin {
...
+                            let html = null;
+                            const num = chart.formatNumber()(interp(t));
+                            if (newValue === 0 && (chart._html.none !== '')) {
+                                html = chart._html.none;
+                            } else if (newValue === 1 && (chart._html.one !== '')) {
+                                html = chart._html.one;
+                            } else if (chart._html.some !== '') {
+                                html = chart._html.some;
+                            }
+                            node.innerHTML = html ? html.replace('%number', num) : num;
```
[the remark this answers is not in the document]
gherka: I agree. That was my first thought to just set ```            if (this._ariaLiveRegion) {
                span
                    .attr('aria-live', 'polite')
                    .transition()
                    .duration(0);
            }```, but the transition still plays out - very quickly, but enough for the screen reader to read out the incorrect figure. There is a [question](https://stackoverflow.com/questions/14104423/how-to-make-true-zero-duration-instant-transitions-in-d3) on SO touching on this issue, but I thought it safer to just forego transition altogether at the expense of some code duplication.

web-src/examples/filtering.html
```diff
@@ -64,12 +64,14 @@
...
     .dimension(yearDim)
     .group(spendPerYear)
+    .keyboardAccessible(true)
     .innerRadius(50)
     .controlsUseVisibility(true);
 
   spendHistChart
     .dimension(spendDim)
     .group(spendHist)
+    .svgDescription("Histogram")
```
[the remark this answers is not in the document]
gherka: Not that I've seen yet. Probably possible with some keyboard modifiers like `Ctrl + Arrow Left / Right`. Will need to investigate properly. [This](https://bl.ocks.org/Andrew-Reid/1ba11865ee0d0585f0274a0976324358) looks quite promising.

spec/bar-chart-spec.js
gherka: Thanks, that's good to know. I'll apply your fix to all the other tests that check this functionality.

src/base/base-mixin.js
gordonwoodhull: Thanks. Okay, let’s stick with the object check for now. Maybe we will figure something out (like another CSS class?) later.

src/charts/number-display.js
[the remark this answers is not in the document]
gordonwoodhull: Please try using `dc.transition()` here, like we do most places. That should do the trick. I’m sure it was not skipped intentionally.

spec/bar-chart-spec.js
gordonwoodhull: Great, thanks!

web-src/examples/filtering.html
```diff
@@ -64,12 +64,14 @@
 yearRingChart
```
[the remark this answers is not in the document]
gherka: Done!

spec/bar-chart-spec.js
gherka: Done!

src/charts/number-display.js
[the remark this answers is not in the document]
gherka: I must've set the transition duration to zero incorrectly initially when testing. Works now with much cleaner code.

gordonwoodhull:
Excellent work, thanks @gherka!

I squashed the commits and released this as 4.2.0.

PR #1744, opened 2020-08-16
Experiment with Prettier

kum-deepak:
I have been trying to use `tslint` to help maintain consistent code formatting. However, after going through all options, it still leaves formatting inconsistencies.

We have started using Prettier (https://prettier.io/) in some of our projects. On new projects, it seems quite natural. I wanted to experiment and see how does it look on this codebase.

The default settings cause too many changes. As a philosophy, Prettier supports very few options. I found the following goes inline with current code conventions:

```JSON
{
  "arrowParens": "avoid",
  "printWidth": 100,
  "singleQuote": true,
  "tabWidth": 4
}
```

It does bring consistent formatting. One departure from current code conventions (no space before `()` in function declarations):

```typescript
 public calculateRadiusDomain (): void {
```
to
```typescript
 public calculateRadiusDomain(): void {
```

I have put two commits here, first one is the setup and the second has some sample conversions. I will also add some comments against specific code lines to highlight some issues.

If we decide to go this route we can pass our tests and samples as well.

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,0 +1,6 @@
+{
+  "arrowParens": "avoid",
+  "printWidth": 100,
+  "singleQuote": true,
+  "tabWidth": 4
+}
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -406,6 +406,12 @@ module.exports = function (grunt) {
             },
             tsc: {
                 command: 'tsc'
+            },
+            'prettier-check': {
+                command: 'prettier --check src'
+            },
+            prettier: {
+                command: 'prettier --write src'
             }
         }
     });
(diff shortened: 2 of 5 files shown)


src/charts/bar-chart.ts
```diff
@@ -177,19 +184,23 @@ export class BarChart extends StackMixin {
...
             .attr('class', 'bar')
             .attr('fill', (d, i) => this.getColor(d, i))
             .attr('x', d => this._barXPos(d))
             .attr('y', this.yAxisHeight())
             .attr('height', 0);
 
-        const barsEnterUpdate: Selection<SVGRectElement, unknown, SVGGElement, any> = enter.merge(bars);
+        const barsEnterUpdate: Selection<SVGRectElement, unknown, SVGGElement, any> = enter.merge(
+            bars
+        );
```
kum-deepak, line 204: This formatting looks weird. These can be forced with `// prettier-ignore`

gordonwoodhull: Is this because the line was too long?

gordonwoodhull:
I'm okay with changing to prettier. Weren't we getting more out of tslint than just formatting, though? Like... actual linting? I'm confused.

src/charts/bar-chart.ts
kum-deepak: Prettier is different than linters. It tries to optimize for human readers, see thier note https://prettier.io/docs/en/options.html#print-width

kum-deepak:
`ts-lint` will still be used. Prettier does not support linting actually. I am also working on tightening `ts-lint` as much as I can. In that effort, I was hitting limitations on formatting and I thought to try prettier.

I will go ahead and configure these to work together. Once ready I will force push into this branch.

src/charts/bar-chart.ts
gordonwoodhull: I can't imagine they are doing this for readability. The original line is 106 characters.

kum-deepak: :smile: I think they would get better over time.

kum-deepak:
`grunt lint` will run prettier in check mode, and `grunt lint-fix` in write mode. Currently only covering `src`.

PR #1745, opened 2020-08-17
Prettified

kum-deepak:
Used some explicit ignores. There were a few cases of nested ternary ?:. These were making my head spin, so, I decided to rewrite it using simpler constructs.


kum-deepak:
I am merging this. In case we find any auto-formatting to be undone we can keep doing as we need.

PR #1746, opened 2020-08-19
Refactoring colors

kum-deepak:
Colors in dc have always stumped me. I wanted to disambiguate both the API and the documentation. This is an attempt in this direction.

I have figured out the following so far:

- Ordinal colors work with an ordinal type of usage. In this case, setting a domain is not required, setting a domain may actually be counterproductive.
- Linear Scales will work well where interpolation is needed. In all likelihood, the domain needs to be explicitly set (as the default is [0,1]) for it to work.
- The backward compatibility mode creates a `scaleQuantize` when an array is passed. This also needs a domain to be explicitly specified.
- LinearScales need the same number of points in domain and range (otherwise it discards), so, in case of dc, these are likely to be used with two values.
- `scaleQuantize` supports any number of points in range, however, they do not support additional interpolation.
- Apart from these, there is mode `colorCalculator` which bypasses the `colorAccessor`. (there are no test cases to cover this).
- In some cases, like bar charts, the color may actually come from the CSS.

I can see that all of the above have possible use cases. I am making an attempt to restructure the API to clearly indicate that these modes are different with different semantics and use cases.

This is WIP.

### DIFF
--- a/src/base/color-mixin.ts
+++ b/src/base/color-mixin.ts
@@ -80,15 +80,12 @@ export function ColorMixin<TBase extends Constructor<BaseMixin>>(Base: TBase) {
          * @returns {ColorMixin}
          */
         public calculateColorDomain(): this {
-            // TODO: use extent from d3-array which does exactly this
-            const newDomain = [
-                min(this.data(), this._conf.colorAccessor),
-                max(this.data(), this._conf.colorAccessor),
-            ];
+            const scale: MinimalColorScale = (this._colorHelper as ColorScaleHelper)
+                .scale as MinimalColorScale;
 
-            // @ts-ignore
-            const scale: MinimalColorScale = (this._colorHelper as ColorScaleHelper).scale;
-            scale && scale.domain && scale.domain(newDomain);
+            if (scale && scale.domain) {
+                scale.domain(extent(this.data(), this._conf.colorAccessor));
+            }
 
             return this;
         }
--- a/src/base/color-mixin.ts
+++ b/src/base/color-mixin.ts
@@ -1,71 +1,14 @@
-import { scaleLinear, scaleOrdinal, scaleQuantize } from 'd3-scale';
-import { interpolateHcl } from 'd3-interpolate';
-import { max, min } from 'd3-array';
+import { scaleQuantize } from 'd3-scale';
+import { extent } from 'd3-array';
 
 import { config } from '../core/config';
 import { BaseMixin } from './base-mixin';
 import { BaseAccessor, Constructor, MinimalColorScale } from '../core/types';
 import { IColorMixinConf } from './i-color-mixin-conf';
-
-export interface IColorHelper {
-    getColor(d, i?: number): string;
-    colorAccessor: BaseAccessor<string>;
-}
-
-export class ColorCalculator implements IColorHelper {
-    public colorAccessor: BaseAccessor<string>;
-    public getColor: BaseAccessor<string>;
-
-    constructor(colorCalculator: BaseAccessor<string>) {
-        this.getColor = colorCalculator;
-    }
-}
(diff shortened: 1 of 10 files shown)


kum-deepak:
I am in double minds about `calculateColorDomain` - it is applicable only to Linear and Quantize scales. Should it be moved to the comapt layer.

kum-deepak:
Initially I though that Linear colors would get used with only start/end. However, I can see cases using more points. For example, to use different shades for -ve/+ve in a Bubble chart:

Domain: -500, 0, 500
Range: red, gray, green

Wow, so much I do not know about colors :smile: 

I will update the typings accordingly.

gordonwoodhull:
* Setting the ordinal color domain is useful when you want control over which value will get assigned to which color. Otherwise you have little control over what order the chart will see the values, so it's hard to get consistent colors between charts for the same ordinal colors, for example.
* `calculateColorDomain` is how you avoid setting the color domain. Linear and Quantize scales are the most common color scales. Sure it might be specific, but there is no way this is legacy functionality!
* The reason why automatic creation of scaleQuantize should be legacy is because it takes an array instead of a scale. We want to encourage literacy of D3 Scales. However, this will probably never go away, because people are generating new examples every day that use `.colors(array)`.
* Yes, there are uses for linear color scales with more than two points.
* I wasn't aware that the colors can come from CSS, but I'm not surprised.

In general, this is one of those places where we want to leave everything open for people to plug in scales we haven't even heard of. It's similar to how X/Y scales are coded for linear scales, but we try to allow log scales or anything else.

Maybe the user even wants to invent their own scale and that's fine. We'll treat it as a function that goes from whatever their colorAccessor returns, to colors.

I took a quick look at the implementation. I like isolating the messiness of color calculator vs color scale in a helper. I am not sure I understand why `.share()` is needed. I guess it is a shallow clone, but it doesn't protect anyone from anything, since the underlying scale will still get modified. Why not just have multiple references to the helper?

As noted above, there are a lot of scales which are valid, and we want to allow arbitrary user scales, so while the helper subclasses `LinearColors` and `OrdinalColors` do tuck away a single messy constructor call, there is no way for it to scale to all the possibilities. That's fine, but I wouldn't want users to think they need to create a helper when passing an arbitrary scale is still fine.

kum-deepak:
All of your remarks make sense. 

`ColorScaleHelper` (base class of the other two specific classes) is designed to use any function, not necessarily a d3 scale, which takes the output of `colorAccessor` and returns a color. Maybe we can rename the class to make that obvious.

Also, if we agree that Quantize scales are going to be commonly used, we can make a concrete class for that.

Composite (and Series) charts have a mode to share colors. In that mode, the children use the same color scale but with their independent `colorAccessor`s (see: https://github.com/dc-js/dc.js/pull/1746/files#diff-55fde3a274919c517a38eb2651119a4dR287). For example, in the case of Ordinal colors, it relies on the same colors to be returned for a given key.

Spending time with this code, I have started getting exposed to more ways to use colors. I find that my team uses gradients as fills/strokes quite a few times. They either use CSS or a callback to select the required elements and set the gradient fill. The value for fill/stroke SVG attributes, apart from a color value, can be a URL specifying gradients (see https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Gradients). So, the scale can possibly return the appropriate URL. The way D3 scales work, these should work. I am going to experiment. :smile: 

I thnik we should also write a tutorial, possibly, in two parts about colors in dc.

kum-deepak:
For example, https://github.com/dc-js/dc.js/pull/1746/files#diff-fa5bc2bb53770f24adf91c1c398e6e11R98, in createion of a `ColorScaleHelper`, newScale can be a scale or just any function.

Issue #1748, opened 2020-08-24
d3@6 has been released


gordonwoodhull:
dc@4 may already be compatible; D3’s embrace of ES6 was our rationale to move on in #1520.

Or there may be other breaking changes in the API.

Now there is a pre-release PR https://github.com/d3/d3/pull/3438 so we should test it!

kum-deepak:
Tested briefly on develop branch. It is failing, at least two changes that affect:

- It seems the way `event` is passed into the `zoom` type of calls has changed.
- d3.nest has been removed.

Could not find any upgrade guide, if it has not been released yet, we can wait.

kum-deepak:
Changes are deeper, there are more failures. I think we should wait for the official upgrade guide. For example one of the failures:

```
Expected '−20' to be '-20'
```

It seems d3 now uses UTF8 (226, 136, 146) which looks like -. This was from test case related to Axis.

gordonwoodhull:
The changes are all documented in the release notes for the individual libraries, and I have been following the progress. I think I know what these problems mean.

Maybe it doesn't matter, but I'd like to be compatible before d3@6 is released. I'm willing to do the work (as long as I find the time).

kum-deepak:
The hyphen issue should only affect test cases, so that should be easiest. This should not affect backward compatibility.

I had created an issue for `d3.nest` #1720. It should be doable in a backward-compatible way.

The events related change - I remember when converting for d3v4, I wished if the event was not to be picked up from a global variable, as it was impacting nested events (e.g., in range/focus charts). This should make implementation simpler. It will impact out test infrastructure, where we simulate events. Not sure if it can be done backward compatible without writing conditional code. We will know when we dig deeper.

gordonwoodhull:
Good points. 

I guess there are two kinds of backward compatibility to be concerned about:
- compatibility with existing user code - as far as I can tell, these changes shouldn't break user code
- compatibility with d3@5 - I agree, we probably need some conditional code, both for events and for `d3.nest`. d3@5 does not contain the replacement group, rollup functions as these were introduced in d3-array@2 [which is ES6 only](https://github.com/d3/d3-array/releases/tag/v2.0.0).

Conditional code is annoying but probably better than a hard break. The alternative, making a hard break to d3@6 with dc@5, would be simpler code but would probably confuse users. 

(I am also not sure how we detect the difference, given that we import individual d3 modules and don't have access to the overall D3 version. But I'm sure there is a way.)

[...]

PR #1749, opened 2020-08-26
d3@v6 with dc@v4

kum-deepak:
Ref #1748 

Steps so far:

- Updated dependency on d3@v6
- Removed hardcoded d3 module dependencies.
- Added d3-collection as a dependency.
- It turns out if we add d3-colllection@1.0.7 with d3@v6, dc remains happy.

For dc@v5 we can remove d3-collection properly if we decide that it will only work with d3@v6.


gordonwoodhull:
I did a quick review. I like the general approach.

I wonder if we could simplify this further, and encapsulate the backward-compatibility code, by moving the logic detecting if `d3.event` exists into `cpt()`?

This would mean consistently using `cpt()` even for calls to `_brushing` and `_onZoom` and so on.

I am hoping this would mean we only do the ugly `if(event)` check in one place, and it might make it easier to remove d3@5- support in the future. (First change it to the identity function, and then remove it entirely.) More importantly, it will make our event handlers purely compliant with d3@6 standards.

I would also prefer to remove the new dependency on `d3-collection`, since that library is deprecated. TBH I haven't used `d3.group` / `d3.rollup` yet but I understand they cover the functionality of `d3.nest`.

kum-deepak:
Makes sense. Let me try.

kum-deepak:
Please check https://github.com/dc-js/dc.js/pull/1749/commits/d899c5cf80b55c5d924c5ce4ae34ae65f49c7568 and https://github.com/dc-js/dc.js/pull/1749/commits/e8f80fa152a32166b7a98b31a81c772547eb0365

kum-deepak:
For `d3.nest` a middle ground seems to be working. Please check https://github.com/dc-js/dc.js/pull/1749/commits/012ae33ef1dc079994fc7cf44c568e08443f6878

kum-deepak:
We also seem to be using `d3.set`. Let me get onto that.

kum-deepak:
`d3.set` is used only in specs. During the conversion, stumbled across a subtle difference in behavior vis-a-vis ES6 `Set`:

```javascript
const set = d3.set();
set.add('3');
set.has(3); // true, false with ES6 set
```

This happens as `d3.set` always converts keys to strings. On the other hand, ES6 Set allows a key to a number and considers 3 and '3' to be different.

gordonwoodhull:
Rebased and merged to develop. Thanks @kum-deepak!

I also took the opportunity to do an npm update and audit fix.

These are my remaining concerns before release. I am not sure if these matter, just want to be careful.

1. package.json now refers to d3-collection in its `devDependencies` and the old references to the rest of the specific d3 modules are gone. I'm not sure if these changes were just for testing, or permanent
2. Ideally d3-collection will not be used if d3 is version 6. However, we still import it into `d3compat.js` and don't use it. Not sure if this will cause trouble, or how to avoid it. (Dynamic module load, shudder?)

kum-deepak:
>     1. package.json now refers to d3-collection in its `devDependencies` and the old references to the rest of the specific d3 modules are gone. I'm not sure if these changes were just for testing, or meant to be permanent.
> 
This is intentional, we now depend on whatever d3 depends on. Sometimes compiler or linter may warn if we import from a dependency that is not directly added to package.json. However, no issues otherwise.

>     2. Ideally d3-collection will not be used if d3 is version 6. However, we still import `nest` from it in `d3compat.js`, and don't use it. Not sure if this will cause trouble, or how to avoid it. (Dynamic module load, shudder?)

In ES6 version, if we remove it, it should still compile and work, though I wanted to keep it clean. It is not used during testing as it is not copied into `spec/3rd-party`. Also, I have ensured that the code checks for `.gorups`, so, that in case both are available `.groups` gets used.

In summary, no harm in keeping it.

gordonwoodhull:
I see, it's just a change of strategy about how we deal with d3 modules. Makes sense!

I'll go ahead and release this.

PR #1750, opened 2020-08-28
Same as #1749 but using d3@v5

kum-deepak:
This PR skips last few commits from #1749. Check this to test against d3@v5.


gordonwoodhull:
I agree that we test this once and then stop testing d3@5 for the future.

It was fun to test d3@4 and d3@5 at the same time in the past, but it creates extra work and problems, and doesn't help our users much.

gordonwoodhull:
I pushed a simplification to the test - it was already confused whether it wanted to work with strings or numbers, so I resolved on numbers instead.

gordonwoodhull:
Looking good!

Let's give `cpt` a more descriptive name. I think it will be okay for it not to be short. What do you think of `adapt_handler`?

You have a typo in the new `core/d3comapt.js` filename.

kum-deepak:
Corrected the typo in the filename and renamed `cpt` -> `adaptHandler`. Let us test against both the versions :smile:

gordonwoodhull:
Okay, I have tested this; IMO it is enough to test against d3@5 this one last time.

Closing and will merge #1749 to develop in a minute.

PR #1754, opened 2020-08-30
GitHub actions - Ubuntu, Windows, and macOS 

kum-deepak:
The initial version, currently it runs our current test workflow on the latest Ubuntu, Windows, and macOS.

These should be possible to be extended to include OS/browser combinations that we intend to test.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -1,6 +1,6 @@
 name: MacOS Test
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   build:
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -14,4 +14,4 @@ jobs:
         with:
           node-version: '11.x'
       - run: npm ci
-      - run: npx grunt ci-pull
+      - run: npx grunt ci-macos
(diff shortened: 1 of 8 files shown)


kum-deepak:
Current config:

- Chrome, Firefox on Ubuntu
- Edge, Chrome, Firefox on Windows
- Safari, Chrome, Firefox on MacOS

Latest OS, Node 12

kum-deepak:
I have removed Saucelabs related settings.

kum-deepak:
Safari Headless is not supported with the NPM version of the plugin, so used from https://github.com/karma-runner/karma-safari-launcher/pull/22

gordonwoodhull:
Wow, this is great!

It's funny that it runs all the tests twice, once for push and once for pull_request, but I guess that is normal.

gordonwoodhull:
This is much faster too!

What a relief to be done with unreliable saucelabs. It was great the first few years but it really deteriorated recently.

kum-deepak:
We can actually drop one of Ubuntu or Travis as both are testing the same stuff.

gordonwoodhull:
Let's drop Travis in a month or two, when we're sure everything is working properly.

It's nice to have the redundancy (especially since it runs in parallel).

PR #1756, opened 2020-09-01
D3 v5 GitHub

kum-deepak:
Cherry-picking infra/test/build related commits from `develop` and adding to `dc-v5` branch.


kum-deepak:
Wow, it failed for no reason. MacOS tests are slowest to complete. Probably we can test only for Safari on macOS.

kum-deepak:
Rerunning the test - success on macOS.

gordonwoodhull:
Is it possible to see old runs? I couldn't find the failure.

Sure, I think it would be okay to test only Safari on macOS. I don't remember seeing too many OS-specific (as opposed to browser-specific) errors.

However, it looks like Windows is often slower than macOS.

kum-deepak:
I think it overwrote that record. I chose to rerun the failed test. In the details, Safari had disconnected twice before it could finish all the tests.

On another note, since Github allows multiple tasks, we can consider setting up one for linting as well.

PR #1757, opened 2020-09-01
dc@v5 with d3@v6

kum-deepak:
The code has moved quite a lot from the `develop` branch, so, git merge was not working. Some cherry-picking and remaining manual changes.

I will create one more PR, for the same code with d3@v5.


kum-deepak:
I am merging this. This will get testing many times, so, not a big concern.

PR #1758, opened 2020-09-01
dc@v5 with d3@v5

kum-deepak:
Same code as #1757 but with d3@v5.


gordonwoodhull:
Thanks for testing this case. I agree, does not need review.

PR #1760, opened 2020-09-02
Update crossfilter repo url

wtoscer:
The repo is crossfilter/crossfilter since 2015
See https://github.com/square/crossfilter/commit/0227e0e443d6426b35a3ccf4839c4456a98c6e0d


gordonwoodhull:
Thanks @wtoscer! Good point, it looks like all of these links need to be updated one way or another.

PR #1762, opened 2020-09-05
Mixins take3

kum-deepak:
Ref #1703, type definitions are now generated.

After considering multiple approaches, settling to this one. It involved one interface each that describes what each of the three Mixins expects. In addition, all members of the mixin themselves need to be public.

We will need to exclude the private members during typdoc configuration.


kum-deepak:
Please ignore it for now. The build fails for compat code.

kum-deepak:
I had been using Node 11. One strange issue, occasionally, the grunt build will succeed (the rollup part), even if it has an error (which would fail in CI environments). If I run `npm ci` and rerun the build it will fail. Not sure what is happening. I think this life of a programmer :angel: 

The first step, I am switching to Node 12 on my local machine.

kum-deepak:
Ready for review.

Issue #1747, opened 2020-08-20
Big Task: Separating data/filter concerns


kum-deepak:
After looking at the source code for a few years and some experiments https://github.com/dc-js/dc.js/pull/1665, I think I am ready to take a plunge. Observations so far:

- It appears that a few charts will fail to work with the non-default key accessor. In some cases, these use a d3 function to preprocess data, which generates a `.key` (like `d3.nest` used by DataGrid and DataTable).
- Cross filter requires the key as it generated for filtering to work correctly. This implies that in charts if a customed `keyAccessor` is defined, the filtering will be broken. Very few charts like `BubbleChart` uses `.key` directly to work correctly with custom `keyAccessor`.
- A simple summary - very few cases where a chart can actually work with a custom key accessor. This itself may not be a limitation all user-visible usage of the key, like labels, titles, etc., have accessors.
- `.value` is never used directly, the output of `.valueAccessor` is used instead. So, we can pre-process the data to calculate and store it.
- Some charts (like CoordinateGrid) call `.data` multiple times. The output of `.data` can be cached at the beginning of the render/redraw cycle.
- Currently, filters are stored within a chart. Storing these with the dimension will resolve quite a few issues. This will also ease the development of alternative data providers.
- Capping can be handled as pure data & filter concern (as shown in https://github.com/dc-js/dc.js/pull/1665/commits/1a63f53e5760d1b03bb3fbbcd29d8d9ce5c907dc). This will allow charts to be simpler. This also enables data providers to support capping differently. For example, a remote backend may cap and send the data.
- Filters need to be easily serializable (say to JSON) to be sent to a remote data provider.
- To show deselected items, few of the carts check if a data is filtered in or out. Currently, it is checked against current filters. Remote providers may find it easier to mark items filtered in/out.
- key, value and filtered status make sense to be pre-computed. We can make a decision for other accessors independently.
- In the case of Crossfilter, other accessors are likely to need computed key/value as well as original .key/.values from the call of group.all.

I will keep amending this for a while. Please free to raise questions/issues.

kum-deepak:
As part of this, I am working on the first PR to make `.data()` to use inheritence instead of being a callback.

gordonwoodhull:
Everything here is likely to cause backward-compatibility problems. I understand that you want to "clean things up" but people use dc.js in a lot of messy ways, because crossfilter doesn't provide all the functionality that is needed for all uses of the data. This is why "fake groups" are so prevalent.

[...]

PR #1765, opened 2020-09-10
Separate Data Concerns

kum-deepak:
Ref #1747. The following has been done:

- Filters related concerns have been moved to `CFFilterHandler`. The code has been extracted from `BaseMixin`.
- The data pipeline starts from `CFSimpleAdapter`, which also precomputes the output of `valueAccessor` and stores it as `d._value`.
- `CFDataCapHelper` implements capping, current code extracted from `CapMixin`.
- `CFMultiAdapter` implements the data pipeline for Stack charts.
- All data extracted from CrossFilter is two-level deep cloned before getting used.
- With all these changes `CapMixin` had become zero functionality, which is dropped.

Please check the structure of classes in `src/data`. Eventually, users will pass a `dataProvider` as per their needs. This helps to keep all data concerns neatly together and allows other types of data providers to be implemented.

For example, to create a `PieChart` with capping they would pass a `CFDataCapHelper`, otherwise `CFSimpleAdapter`.

I am not entirely happy with the current names of classes in `src/data` and ooking for suggestions.

As part of this PR, the StackMixin code will get minor cleanup. Subsequent steps will be carried out as independent PRs.


gordonwoodhull (approved):
I read over this and it looks like it preserves functionality while moving stuff around safely.

I think I am missing the big picture about what this all achieves, but overall it looks good.

gordonwoodhull:
To be clear, I think you have described what you are doing sufficiently. I just haven't assimilated it yet.

I probably need more time to digest what you are doing. Some examples and upgrade documentation would help. (All in due time, no rush.)

PR #1767, opened 2020-09-11
Gruntfile refactoring

kum-deepak:
Well, after procrastinating for a few years, finally took the plunge. :smile: 

The main purpose is to remove portions that are unused and to make it readable. Removed most of the string interpolations where the target string is well known by now (like `src`, `dest`, etc.).

I have also refactored grouped tasks to minimize duplicate/redundant subtasks.

The watch type of tasks needs further work. Over the years tooling has changed and so has dependencies. I will put more details in subsequent comments.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,7 +8,10 @@
     "ecmaVersion": 6,
     "sourceType": "module"
   },
-  "extends": "eslint:recommended",
+  "extends": [
+    "eslint:recommended",
+    "prettier"
+  ],
   "rules": {
     "no-useless-escape": "off",
     "no-console": "off",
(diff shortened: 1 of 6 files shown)


kum-deepak:
All the commits are small and should be clear on their intent. Please ask if any of those are unclear.

kum-deepak:
I think I have been able to regain my sanity.

I have created 3 server+wtach tasks:

- `test-n-serve` - this is suitable when doing active code changes, it starts server, watches, and keeps building, testing, and updates examples.
- `docs-n-serve` - this is suitable when doing doc updates. It starts a server, watches, and keeps docs up to date.
- `server` - equivalent to both of the above.

I have kept following across all of these tasks:

- These tasks should be faithful, i.e., during watch it should keep the output correct. In a few cases, it was cleaning old files (if a file needed to be removed). The removal is a bit aggressive now.
- All the tasks run as soon as it is started - this helps in keeping consistency across initial setup and rebuild. This also gives one additional benefit that even if the initial run fauns the watch process continues.
- In case of a change during a build, the task will interrupt and start afresh. This will avoid missing changes.

It is ready for merge. If there are issues we can do additional PRs.

gordonwoodhull:
Looks good!

It's going to take me a little time to get used to those trailing commas, but they help making commit diffs small and on point, so 👍 .

I agree that the config wasn't helping much. I guess it was supposed to make it easier to change directories and such, but we don't do that very often.

Task dependencies have always been a mess, glad you cleaned that up!

kum-deepak:
I am also getting used to the trailing commas - my head is not tuned to these. As a whole, I am finding that the more time I spend with prettier formatted code, I see the value. I think at least it is leading to a consistent code formatting.

PR #1773, opened 2020-10-25
Refactoring filters

kum-deepak:
Long gap before any progress. After a couple of unsuccessful attempts, finally a working version. Probably others were way bit too ambitious.

In the last commit (https://github.com/dc-js/dc.js/commit/4885b413f075f06c947d3e65699ab6cc48b8b14f), I have also updated an example that used filterHandler. I have made minimal changes for now. However, it might be possible to reimplement the example without overriding filterHandler.

I kept it lots of small commits as I was bitten badly in my previous attempts. I will roll these into fewer commits before merging.


kum-deepak:
This PR drops filterHandler related APIs, these however can be reimplemented using inheritance. I have updated one of the examples.

gordonwoodhull:
So, to summarize, it looks like this PR removes all the filter handlers in favor of overriding methods on the data adapter.

And the rationale is that using ES6 is cleaner than specifying handler functions.

This provides an example of overriding `applyFilters` instead of providing a `filterHandler`, and it looks like all the other methods of the data adapter are overridable as well,

OK, sounds good. Are there other benefits to the new approach, besides cleanliness?

kum-deepak:
You are totally correct. My current motivations were to simplify code structure for now.

kum-deepak:
After having reached here, I realized that there is only one method that uses dimenion and the crossfilter - which tempts me to break it into two classes. I will hold onto that thought for now.

I am merging this now.

At a later stage, I will start work on keeping the filter information at the dimension.

PR #1774, opened 2020-10-26
Make Filters as classes

kum-deepak:
Presently quite a filters were array instances blessed with additional methods. While this worked, the design was not desirable.

Currently, I have taken a minimal code change approach by inheriting from `Array` to acheive functionally similar code. In this step, I have also made all filter classes to follow a common interface.


kum-deepak:
I would have liked to not inherit from Array and keep the values as members. However, in a few places in the code, the filter is used as an Array. I tried keeping these as members, however, it breaks things, not worth the effort at this stage.

gordonwoodhull:
It's nice to have a correct way to extend Arrays! 👍

Issue #1772, opened 2020-10-16
Brush not working in LineChart after upgrade to D3 v6


ewaldman:
Hi.
I am using DC version 4.2.0. I upgraded my D3 to 6.2.0 and now the brush on my LineChart does not work.
When dragging the brush it does indeed filter out the other charts, however when releasing the brush it snaps back and nothing is filtered out. Also, after releasing the brush, the chart continuously redraws itself. When I moved back to D3 version 5.16.0 it works like a charm.

gordonwoodhull:
Hi @ewaldman, thanks for the report!

Brushing works okay in [the main page](http://dc-js.github.io/dc.js/) as well as [this example](http://dc-js.github.io/dc.js/examples/brush-ordinal.html).

Are you using any custom event handlers in your application? D3 changed the signature for event handlers so that the event is passed as the first argument instead of using a global, so pretty much all event handlers need to be changed.

ewaldman:
Thank you for responding. 
Once I added the 'event' parameter to events the brushes do indeed work.

However, I have another question.

The event I am using is renderlet.
The old code was:
this.timeLineChart.on('renderlet', ( chart) => {
          chart.select('svg').attr('width', null).attr('height', null).attr('viewBox',
            '0 0 ' + TIME_LINE_WIDTH + ' ' + 300).attr('preserveAspectRatio', 'xMinYMin'); 
        });

Based on your answer above I added event , and this is the new function:

this.timeLineChart.on('renderlet', (event, chart) => {
          chart.select('svg').attr('width', null).attr('height', null).attr('viewBox',
            '0 0 ' + TIME_LINE_WIDTH + ' ' + 300).attr('preserveAspectRatio', 'xMinYMin');
        });

However, google chrome is giving errors.  

Uncaught TypeError: Cannot read property 'select' of undefined
    at LineChart.<anonymous> (myService.js:1887)
    at Dispatch.call (dispatch.js:57)
    at SVGSVGElement.<anonymous> (dc.js:1979)
    at Dispatch.call (dispatch.js:57)
    at tick (schedule.js:141)
    at timerFlush (timer.js:61)
    at wake (timer.js:71)

Any tips ?

Thank you.

ewaldman:
On further investigation , it seems that my first way of the function was right.


this.timeLineChart.on('renderlet', ( chart) => {
chart.select('svg').attr('width', null).attr('height', null).attr('viewBox',
[...]

PR #1780, opened 2020-11-01
Fix the cumulative line chart.

kum-deepak:
It seems the cumulative line in the [Pareto chart example](http://dc-js.github.io/dc.js/examples/pareto-chart.html) is incorrect. Removing the sort in the fake group seems to fix it. Please review.

### DIFF
--- a/web-src/examples/pareto-chart.html
+++ b/web-src/examples/pareto-chart.html
@@ -86,6 +86,7 @@
         .yAxisLabel("The Y Axis")
         .legend(dc.legend().x(80).y(20).itemHeight(13).gap(5))
         .renderHorizontalGridLines(true)
+        .ordering(kv => -kv.value.value)
         .compose([
             dc.barChart(chart)
                 .dimension(dim_)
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
The bars should be sorted from greatest to least, but they aren't taking the ordering from the group.

It looks like the example worked with dc@2, but stopped working some time after that. The screenshot shows it worked [when I answered the question on SO](https://stackoverflow.com/questions/59224322/how-to-draw-a-pareto-chart-in-dc-js).


This fixes it:

```js
        .ordering(kv => -kv.value.value)
```

There are lots of confusing issues around ordinal charts, but I'm not going to get into it now.

kum-deepak:
Thanks, @gordonwoodhull! I have updated the code.

kum-deepak:
The scatter-series example was showing the issue reported in #1772. Switching off `mouseZoomable`.

gordonwoodhull:
Thanks @kum-deepak!

PR #1785, opened 2020-11-18
Share Filters

kum-deepak:
*This is still a preview.*

There is a structure to hold filter information by dimension. It is controlled by a flag `shareFilters`.

First round of work on Coordinate chart makes Range/Focus charts simpler. The filter information is already propagated, so, only the UI need to be updated. The current code in not final - I would do additional refactoring and cleanup.

Interesting observation - there is no special handling needed for Focus/Range. If we introduce a flag - say `autoFocus` - which causes the chart to focus when a filter is applied, we would not need to mark range/focus charts at all.

During this phase I realized that zoom acitivity on Focus chart need not cause a redraw on Range chart - just brush needs to be updated. This happens now and test cases have accordingly been updated.

I have checked with stocks example, which seem to be fine.

Tests in Composite charts are currently failing.

### DIFF
--- a/spec/composite-chart-spec.js
+++ b/spec/composite-chart-spec.js
@@ -714,6 +714,8 @@ describe('dc.compositeChart', () => {
             lineDimension = data.dimension(d => +d.value);
             lineGroup = lineDimension.group();
 
+            chart.dataProvider().configure({ shareFilters: false });
+
             chart
                 .dimension(scatterDimension)
                 .group(scatterGroup)
--- a/spec/coordinate-grid-chart-spec.js
+++ b/spec/coordinate-grid-chart-spec.js
@@ -839,7 +839,6 @@ describe('dc.coordinateGridChart', () => {
         });
 
         it('range filter should execute filtered listener and zoom focus chart', () => {
-            spyOn(chart, 'focus').and.callThrough();
             const expectedCallbackSignature = function (callbackChart, callbackFilter) {
                 expect(callbackChart).toBe(rangeChart);
                 expect(callbackFilter).toEqual(selectedRange);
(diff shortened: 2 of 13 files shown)


kum-deepak:
Composite charts work now. For hetereogenous filter spec - started using `shareFilters` as `false`.

kum-deepak:
Hierarchy of DataProvider classes:

- Filters:
  - `FilterHandler` - `add`, `remove`, `toggle`, and `filter`
  - `CFFilterHandler` extends `FilterHandler` - CrossFilter specific, apply filters on `dimension`
  - `FilterStorageHelper` extends `CFFilterHandler` - store filters with common storage,
    enables `shareFilters`. Actual storage via `ChartGroup`.
- Data generation:
  - `CFSimpleAdapter` extends `FilterStorageHelper` - data generation, value accessor
  - `CFMultiAdapter` extends `CFSimpleAdapter` - data generation for charts that need multiple layers,
    currently stack based charts, in future grouped row charts.
  - `CFDataCapHelper` extends `CFSimpleAdapter` - data capping.
    Moving it here allows it to be used with any chart.

kum-deepak:
The focus method has been refactored - a part that updates the filter followed by updaing the chart elements. This allows only UI to be updated when filters are updated by some other chart.

kum-deepak:
@gordonwoodhull it is ready for review and merge.

kum-deepak:
I have merged this. Feedbacks will be taken as new PRs as needed.

gordonwoodhull:
This is great. I love how it cleans up the data versus the UI layers.

To test this out, I added a [linked filter example](https://dc-js.github.io/dc.js/examples/linked-filter.html) to dc@4. This demo had been floating around as a fiddle for about a year.

Then I ported it to dc@5 by removing all the custom code. Works great!

I also ran through the examples and I'll file an issue for those which are currently failing.

kum-deepak:
This example is great, I was going to write a similar example. :smile: I should have known what is already there.

In the future, we can make one of the charts mouseZoomable and auto-focus - that would demonstrate that focus/range charts are no longer special.

PR #1792, opened 2020-11-29
Concept of `autoFocus`, range/focus moved to the compat layer - recreated

kum-deepak:
Recreated #1790

### DIFF
--- a/src/base/coordinate-grid-mixin.ts
+++ b/src/base/coordinate-grid-mixin.ts
@@ -70,8 +70,6 @@ export class CoordinateGridMixin extends ColorMixin(MarginMixin) {
     private _nullZoom: ZoomBehavior<Element, unknown>;
     private _hasBeenMouseZoomable: boolean;
     private _ignoreZoomEvents: boolean;
-    private _rangeChart: CoordinateGridMixin;
-    private _focusChart: CoordinateGridMixin;
     private _fOuterRangeBandPadding: number;
     private _fRangeBandPadding: number;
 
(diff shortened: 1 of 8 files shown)


src/base/coordinate-grid-mixin.ts
```diff
@@ -1220,11 +1191,6 @@ export class CoordinateGridMixin extends ColorMixin(MarginMixin) {
         if (this._conf.zoomOutRestrict) {
             // ensure range is within self._xOriginalDomain
             range = this._checkExtents(range, this._xOriginalDomain);
-
-            // If it has an associated range chart ensure range is within domain of that rangeChart
-            if (this._rangeChart) {
```
kum-deepak, line 1225: @gordonwoodhull this seems to be the only functional code that is getting dropped.

gordonwoodhull: Oh, I get it now, sorry I misunderstood.

I should not try to review PRs without trying them out!

Yes, I agree this is an obscure case. I misread your earlier comment "the earlier code was also checking the domain of the other chart and constraining based on that" and thought it meant that pan and zoom on the focus chart would not set the brush on the range chart, but of course it still does that.

Sorry again for the confusion.

kum-deepak: Thanks, @gordonwoodhull  for the review. :+1:

kum-deepak: English is my 3rd language and communication definitely not my strongest suit :smile:

kum-deepak:
I will go ahead and merge the PR.

I have opened an issue #1793 to explore if ElasicX can play well with zoom/focus.

src/base/coordinate-grid-mixin.ts
gordonwoodhull: dc.js is sufficiently complex that it’s not easy to write about clearly.

A lot of my own comments mystify me only months later. 🤣

Issue #1787, opened 2020-11-22
filter stacks example has invalid segment filtering


gordonwoodhull:
This example worked in dc@2 or dc@3, but currently clicking on a segment produces invalid selections with undefined in them:

<img width="933" alt="image" src="https://user-images.githubusercontent.com/1366709/99891914-11e85c80-2c3d-11eb-8c5c-6503b2631af0.png">

kum-deepak:
This example uses good knowledge of internal field names (not part of public API). Fixed to update the field names as per internal changes. Also needed was to use d3 v6 compact for the click event handler.

This example will undergo further changes to remove all deprecated calls.

gordonwoodhull:
I tried to cherry-pick b0be203 to develop in order to close this out, but for some reason, `adaptHandler` is not exported even though it clearly has `export` in its definition. Not sure how to troubleshoot this since rollup is a black box.

kum-deepak:
I will look into it. The entire commit is not needed here. Also, the `adaptHandler` may be exported from the file but may not be part of the library export. I will check and fix it.

kum-deepak:
Please see #1796.

PR #1796, opened 2020-12-02
Filter stacks dc v4

kum-deepak:
Ref https://github.com/dc-js/dc.js/issues/1787#issuecomment-736744459

I have made small commits, please feel free to squash. Even though the change is small, it will need a minor version upgrade and a release.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -10,6 +10,7 @@ export * from './core/logger';
 export * from './core/printers';
 export * from './core/units';
 export * from './core/utils';
+export * from './core/d3compat';
 
 export * from './base/base-mixin';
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -10,6 +10,7 @@ export * from './core/logger';
 export * from './core/printers';
 export * from './core/units';
 export * from './core/utils';
+export * from './core/d3compat';
```
gordonwoodhull, line 13: Aha, now I get why grepping for `adaptHandler` was not finding the export!

gordonwoodhull (approved):
Thanks! Will merge and release soon.

PR #1797, opened 2020-12-10
Compute innerRadius dynamically while pie-chart resizing

kum-deepak:
We were facing an issue with `innerRadius` in responsive Pie charts. I thought of adding the solution to the pie resizing example.


gordonwoodhull:
Looks nicer, too! Thanks @kum-deepak!

gordonwoodhull:
Released in 4.2.3

PR #1798, opened 2020-12-10
Current filters and save/restore widgets

kum-deepak:
After running `grunt server` go to http://localhost:8888/web/examples/current-filters.html. This example uses stock.js as is (this meant copying ndx.json).

This demonstrates two capabilities:

**Current filters widget**

- As filters are applied/removed it updates the widget. 
- Individual filters can be cleared by clicking the X icon.
- There is also a `clear all` which works similar to `chartGroup.filterall` however it does so by restoring an empty list as filters - effectively removing all filters.
- Currently, the filters are just converted to string - need to take the help of filter printers.

**Saved filter states**

- Click on `save current state` to save - it will get added to the saved state list.
- On mouse hover, it will show JSON representation of the state.
- Clicking on the state will restore it.
- Clicking on X will remove this.
- I am thinking of populating a couple of sample items from a JSON file to demonstrate that states can be loaded from external sources.

Currently experimental; will proceed based on feedback.


gordonwoodhull:
This is neat!

Do you think it would be worthwhile to make a reusable widget out of the Current Filters display? The Saved Filter States is probably just for the demo, but I imagine people will want the Current Filters widget, and that's a fair amount of code to copy/paste.

gordonwoodhull (approved):
Very cool. Great demo. Implementation looks clean.

kum-deepak:
I am not going to merge this as of now. Before the final release, the current filters would be created as a widget.

kum-deepak:
I decided to merge this. Keeping too many open PRs is not great - I realized the hard way :smile:

PR #1800, opened 2020-12-13
Fixed undefined this on SeriesChart redraw

blaze33:
I had the following bug when redrawing a SeriesChart:
```
dc.js:12456 Uncaught TypeError: Cannot read property '_charts' of undefined
    at _clearChart (dc.js:12456)
    at Array.map (<anonymous>)
    at SeriesChart._resetChildren (dc.js:12463)
    at SeriesChart.chart (dc.js:12482)
    at showUpdatedChart (panel.js:35)
```
The issue is that passing `this._clearChart` to `map` like that makes it unbound, thus the need to bind it again to `this` to clear the issue.
I spent a good time looking for this issue online and have found nothing, so if I'm misguided in any way, let me know, thanks!
Nonetheless, this fix solves correctly my use case.

### DIFF
--- a/src/charts/series-chart.js
+++ b/src/charts/series-chart.js
@@ -102,7 +102,7 @@ export class SeriesChart extends CompositeChart {
     }
 
     _resetChildren () {
-        Object.keys(this._charts).map(this._clearChart);
+        Object.keys(this._charts).map(this._clearChart.bind(this));
         this._charts = {};
     }
 


gordonwoodhull:
Thanks @blaze33!

Looks like we missed this in f172fb730 when we converted to ES2015/ES6.

blaze33:
Thanks for the quick answer! Glad to know I could help a bit :)

gordonwoodhull:
Released in 4.2.3

PR #1808, opened 2021-01-16
d3v6 compat, d3.mouse vs d3.pointer

### LINKED ISSUE #1807: Removed in d3v6, d3.mouse is used by `BubbleOverlay.debug`
This is not a commonly used option, so, it did not get detected earlier. It is possible to write a compact (new method `d3.pointer`). This needs to be applied to both dc v4 and v5.

kum-deepak:
Fixes #1807 for dcv4.

### DIFF
--- a/src/charts/bubble-overlay.js
+++ b/src/charts/bubble-overlay.js
@@ -1,11 +1,9 @@
-import {mouse} from 'd3-selection';
-
 import {BaseMixin} from '../base/base-mixin';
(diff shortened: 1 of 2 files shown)


kum-deepak:
Testest with d3@v5 and d3@v6 by turning on the `debug` flag in the `crimes` example.

gordonwoodhull:
Makes sense.

Tested with http://127.0.0.1:8888/web/crime/

Thanks @kum-deepak!

PR #1809, opened 2021-01-22
Alternate approach to Ordinal Brushing

kum-deepak:
This approach makes a key assumption that brushing makes logical sense only if the underlying data has a defined order.

In this approach:

- Approach to mapping string <-> int is similar (converted as a class for simpler syntax).
- Dimension key is an integer - derived based on the mapping.
- Where needed the integer key is mapped back to string - for example, the title, axis label, etc.
- There is no need for a custom filter handler.
- The previous implementation assumes that each row has unique keys. This may not be the case with most data sets. So, I have updated that part of the code to work with rows where keys may repeat.

The brushing and re-sorting, in my opinion, do not go well together. For example, if there is a selection and the data get re-sorted (for example because of activities on the other charts), the selection will now include different items, which may not be intended. In my opinion, unless there is a natural ordering in the data, range brushing does not make logical sense.

### DIFF
--- a/web-src/examples/brush-ordinal-dynamic.html
+++ b/web-src/examples/brush-ordinal-dynamic.html
@@ -1,0 +1,143 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>dc.js - Brushing on an Ordinal Bar Chart</title>
+    <meta charset="UTF-8">
+    <link rel="stylesheet" type="text/css" href="../css/bootstrap.min.css">
+    <link rel="stylesheet" type="text/css" href="../css/dc.css"/>
+</head>
+<body>
+
+<div class="container">
+<script type="text/javascript" src="header.js"></script>
+<p>This example demonstrates brushing on an ordinal bar chart, by mapping the values to
+integers and specifying a linear scale. The data is sorted by descending value in one
+fake group, and then another fake group provides the integer/ordinal mapping.</p>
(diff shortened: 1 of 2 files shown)


kum-deepak:
In dc-v5, this approach simplifies http://127.0.0.1:8888/web/examples/focus-ordinal-bar.html significantly with hardly any code that relies on the internal workings of dc. Please see commit https://github.com/dc-js/dc.js/pull/1810/commits/bf81f80211dfaa292fab47fbc2caf59907c8abc9

gordonwoodhull:
I agree this is a lot simpler and easier to understand, and also that it is a mess if there is an active selection when the domain gets rearranged. It's great not to have to use a `filterHandler`.

However, mapping the keys only at the start also precludes the use of [remove_empty_bins()](https://github.com/dc-js/dc.js/wiki/FAQ#remove-empty-bins) and streaming data.

Generally dc.js charts do try to adapt when the domain changes. That's why there was a fake group in the original example - it updated the mapping each time data was pulled. It is really messy to combine mapping and fake group, for sure, but that's the reason it was that way.

I think you've done three simplifications here:

1. using the integers instead of ordinals throughout
2. separating logic into a Mapper
3. only read the set of keys once when the chart is first drawn

I think 1 and 2 are real improvements, but 3 may not work in all situations. So I'm merging this but preserving the old example under the name brush-ordinal-dynamic.

Issue #1788, opened 2020-11-22
Examples currently failing on dc@5


gordonwoodhull:
The following examples are failing on `dc-v5` due to changes in the API.

No rush, we can deal with these in due time. But the examples are a manual test suite which needs to be fixed each major version, at least. They exercise weird features of dc.js!

http://127.0.0.1:8888/web/examples/text-filter-widget.html

```
Uncaught Error: Mandatory attribute chart.group is missing on chart[#search]
    at TextFilterWidget$1.checkForMandatoryAttributes (base-mixin.ts:448)
    at base-mixin.ts:466
    at Array.forEach (<anonymous>)
    at TextFilterWidget$1.render (base-mixin.ts:466)
    at ChartGroup.renderAll (chart-group.ts:39)
    at Object.renderAll (chart-registry.ts:164)
    at text-filter-widget.html:73
```

http://127.0.0.1:8888/web/examples/time-intervals.html

```
Uncaught (in promise) TypeError: monthChart.filterHandler is not a function
    at time-intervals.html:114
```

http://127.0.0.1:8888/web/examples/bar-single-select.html
```
Uncaught (in promise) TypeError: chart.width(...).height(...).x(...).xUnits(...).brushOn(...).yAxisLabel(...).dimension(...).group(...)
.controlsUseVisibility(...).addFilterHandler is not a function
    at bar-single-select.html:47
```

http://127.0.0.1:8888/web/examples/adjustable-threshold.html

is attempting to set

```js
        .colors(["orange"])
```
and this results in black bars.

http://127.0.0.1:8888/web/examples/brush-ordinal.html

when brushing
```
brush-ordinal.html:91 Assertion failed: console.assert
applyFilters @ brush-ordinal.html:91
filter @ filter-handler.ts:99
filter @ base-mixin.ts:660
replaceFilter @ base-mixin.ts:600
applyBrushSelection @ coordinate-grid-mixin.ts:918
(anonymous) @ coordinate-grid-mixin.ts:912
(anonymous) @ events.ts:42
setTimeout (async)
events.trigger @ events.ts:40
_brushing @ coordinate-grid-mixin.ts:911
(anonymous) @ coordinate-grid-mixin.ts:848
(anonymous) @ d3compat.ts:15
call @ d3.js:1224
emit @ d3.js:5162
start @ d3.js:5148
started @ d3.js:5254
(anonymous) @ d3.js:1985
[...]

PR #1810, opened 2021-01-23
Fixing examples for dc-v5

kum-deepak:
Ref #1788


kum-deepak:
I will go ahead and merge this. Any changes will be taken as new PRs.

PR #1816, opened 2021-02-25
feat(coordinateGridMixin): useTopXAxis

aberenyi:
@gordonwoodhull could you please cast your eye over this?

### DIFF
--- a/src/base/coordinate-grid-mixin.js
+++ b/src/base/coordinate-grid-mixin.js
@@ -323,6 +323,9 @@ export class CoordinateGridMixin extends ColorMixin(MarginMixin) {
      */
     xAxis (xAxis) {
         if (!arguments.length) {
+            if (!this._xAxis) {
+                this._xAxis = this._createXAxis();
+            }
             return this._xAxis;
         }
         this._xAxis = xAxis;
--- a/src/base/coordinate-grid-mixin.js
+++ b/src/base/coordinate-grid-mixin.js
@@ -546,8 +546,7 @@ export class CoordinateGridMixin extends ColorMixin(MarginMixin) {
         }
 
         let axisXLab = g.select(`text.${X_AXIS_LABEL_CLASS}`);
-        const axisXLabY = (this._useTopXAxis ? 0 : this.height()) +
-            this._xAxisLabelPadding * (this._useTopXAxis ? 1 : -1);
+        const axisXLabY = this._useTopXAxis ? this._xAxisLabelPadding : (this.height() - this._xAxisLabelPadding);
         if (axisXLab.empty() && this.xAxisLabel()) {
             axisXLab = g.append('text')
                 .attr('class', X_AXIS_LABEL_CLASS)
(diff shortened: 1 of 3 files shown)


src/base/coordinate-grid-mixin.js
```diff
@@ -429,6 +433,36 @@ export class CoordinateGridMixin extends ColorMixin(MarginMixin) {
...
+                'Make calls to useTopXAxis sooner in your chart creation process.');
+        }
+
+        this._useTopXAxis = useTopXAxis;
+
+        // adjust default top margin to make room for the labels
+        if (this._useTopXAxis && this.margins().top < 30)
+        {
+            this.margins().top = 30;
+        }
```
gordonwoodhull, line 461: We don't generally adjust the margins automatically, because it gets very complicated very quickly with all the interacting features.

I think it's better to change the example to set the margins. 

Right now the example has a big margin at the bottom which it is not using.

image

web-src/examples/scatter-top.html
```diff
@@ -0,0 +1,52 @@
...
+
+  var ndx                 = crossfilter(experiments),
+      runDimension        = ndx.dimension(function(d) {return [+d.Run, -d.Speed]; }),
+      speedSumGroup       = runDimension.group();
+
+  chart
+    .width(768)
+    .height(480)
+    .x(d3.scaleLinear().domain([6,20]))
+    .useTopXAxis(true)
```
gordonwoodhull, line 35: Instead of defaulting the margin, please change the margins in the example.

```diff
+    .margins({top: 30, right: 0, bottom: 10, left: 30})
```

src/base/coordinate-grid-mixin.js
```diff
@@ -503,11 +546,12 @@ export class CoordinateGridMixin extends ColorMixin(MarginMixin) {
         }
 
         let axisXLab = g.select(`text.${X_AXIS_LABEL_CLASS}`);
+        const axisXLabY = (this._useTopXAxis ? 0 : this.height()) +
+            this._xAxisLabelPadding * (this._useTopXAxis ? 1 : -1);
```
gordonwoodhull, line 550: I think it would be clearer to write this as

 ```suggestion
         const axisXLabY = this._useTopXAxis ? this._xAxisLabelPadding : (this.height() - this._xAxisLabelPadding);
```

File: web-src/examples/scatter-top.html [hunk 4]
```diff
@@ -0,0 +1,52 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>dc.js - Scatter Plot Example</title>
```
gordonwoodhull, line 4:
```suggestion
    <title>dc.js - Scatter Plot Top X Axis Example</title>
```

gordonwoodhull (changes requested):
Thanks @aberenyi! Overall the changes look great.

As a matter of policy, we usually don't try to adjust margins automatically. The interaction of features gets really complicated and we would have to devise a strategy (perhaps using flexbox) rather than writing special case logic everywhere.

So for now we leave this in control of the user. In this case it's better to change the example.

Also I found the double `?:` confusing so I made a suggestion there.

src/base/coordinate-grid-mixin.js
aberenyi: Makes perfect sense!

gordonwoodhull:
Thanks @aberenyi!

PR #1819, opened 2021-03-15
fix(coordinateGridMixin): renderVerticalGridLines should consider useTopXAxis 

### LINKED ISSUE #1818: Vertical grid lines aren't visible when useTopXAxis is true
As the title states, when using a top X axis, the vertical grid lines are rendered with 0 height.

aberenyi:
Fixes https://github.com/dc-js/dc.js/issues/1818

### DIFF
--- a/src/base/coordinate-grid-mixin.js
+++ b/src/base/coordinate-grid-mixin.js
@@ -579,7 +579,7 @@ export class CoordinateGridMixin extends ColorMixin(MarginMixin) {
                 .attr('x1', d => this._x(d))
                 .attr('y1', this._xAxisY() - this.margins().top)
                 .attr('x2', d => this._x(d))
-                .attr('y2', 0)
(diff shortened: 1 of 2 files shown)


aberenyi:
@gordonwoodhull would you mind to review this whenever you have a spare few mins?

gordonwoodhull:
Sorry for the slow response. This makes sense to me and I will merge it next release.

If you would be so kind as to add a test case or two, we can ensure that the new feature keeps working into the future. 

I think if you just add a test case that fails without this fix and succeeds with it, that will be sufficient to test `useTopXAxis`, although it wouldn't hurt to test something from the last PR, like `_xAxisY()` or the translate on the label position.

Thanks!

aberenyi:
@gordonwoodhull no probs at all.

Will try to write some tests in due course.

Cheers

PR #1825, opened 2021-04-05
Add d3@v5 tests at the Github CI

kum-deepak:
Test cases that checked for d3 formatted -ve numbers were failing. Made these generic.


gordonwoodhull (approved):
Nice! As I recall, it was uglier to test d3@4 and d3@5 simultaneously, so we decided not to test d3@5 when we moved to d3@6.

However, this is quite simple! (except for those silly minus signs)

kum-deepak:
:smile:

PR #1843, opened 2021-09-14
Fix broken links in latest API docs

peteretelej:
Due to change of domain from `dc-js.github.com` to `dc-js.github.io`, a lot of broken references to the old urls still exist.

This fixes broken links that exist in the latest API docs, as the docs are currently somewhat broken with them. Additional references exist either in older API docs or code comments but I didn't feel they need updating.


gordonwoodhull:
Thanks @peteretelej and sorry for the slow response. Indeed, GitHub Pages moved to github.io many years ago.

`api-latest.md` is an artifact generated from comments in the source code. I've fixed the sources in 05ed0c34bf9133 and updated the markdown artifact.

Issue #1846, opened 2021-10-27
remove d3v5 compatibility layer


gordonwoodhull:
I'm glad that we took a conservative approach for adding d3v6 compatibility, with the compatibility layer.

However, having this extra level makes the code hard to read - I was just trying to read how data table sections work, and it's confusing having also to read compat/d3v6.js in order to understand it.

I think we should remove d3v5 compatibility for dc-v5.

kum-deepak:
Thanks for raising this. I see one more reason: it will fundamentally fix https://github.com/dc-js/dc.js/issues/1822.

Modules support in the browsers is improving, and we should make dc to work as modules directly in the browser.

kum-deepak:
closed by #1847

PR #1847, opened 2021-10-29
Remove d3v5 compat

kum-deepak:
Ref #1846 

I am targeting d3@v7. I have checked the changelog between v6 and v7, it does not seem that dc is affected by any breaking changes. So, most likely dc will work with d3@v6 as well.

In the first commit, I have removed the nest helper and inlined relevant parts of the code.


gordonwoodhull (approved):
I haven't tested, but this looks great.

Interesting that we only use d3.mouse for a debug thing. Looks like d3.pointers is supposed to replace this. https://github.com/d3/d3-selection/pull/253

Only adaptHandler left! This one is a big task but it will make the code much more readable to use d3@6+ handler signature and no wrapper.

kum-deepak:
Next, I will work on d3.mouse. I realized it has been added to compact for dc@v4 (https://github.com/dc-js/dc.js/pull/1808 and https://github.com/dc-js/dc.js/issues/1807).

kum-deepak:
Completed for now. @gordonwoodhull please check when you have time.

gordonwoodhull (changes requested):
Wow, it's so much easier to read the event handlers with the idiomatic D3 handler signature and no `adaptHandler`.

One quibble: I'd prefer to inline `{}` rather than declaring `dummyEvent` every time we need a dummy event. It's not like we have to document what the parameter is for, since anyone working on dc.js should understand D3 idioms.

gordonwoodhull (approved):

gordonwoodhull:
Thanks @kum-deepak!

kum-deepak:
Next, I will sync examples from the current develop branch and upgrade those to the latest syntax. After that, I will complete the migration guide.

Issue #1848, opened 2021-11-02
all builds are failing


gordonwoodhull:
Presumably some of our tooling has gone stale, since all GitHub Actions are failing with the below error.

This started today; builds yesterday were ok.

(Same error on all platforms, even though the error references Safari.)

```
Run npm ci
npm ERR! Error while executing:
npm ERR! /usr/local/bin/git ls-remote -h -t git://github.com/RLovelett/karma-safari-launcher.git
npm ERR! 
npm ERR! fatal: remote error: 
npm ERR!   The unauthenticated git protocol on port 9418 is no longer supported.
npm ERR! Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
npm ERR! 
npm ERR! exited with error code: 128

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/runner/.npm/_logs/2021-11-02T13_31_53_947Z-debug.log
Error: Process completed with exit code 1.
```

kum-deepak:
I restarted the failed builds, it worked. In addition, I have updated the `git:` URL for one of the packages to `git+https:` which should resolve the underlying issue as well. Please check #1849

gordonwoodhull:
Fixed by #1849

gordonwoodhull:
They probably switched on more strict security rules temporarily in order to warn people to make this change. This seems to be standard GitHub practice.

PR #1849, opened 2021-11-03
Update package path to use https: instead of git: for github

kum-deepak:
Ref #1848

Updating path to use https:


gordonwoodhull:
Thanks @kum-deepak!

PR #1850, opened 2021-11-06
Updating examples for dc-v5

kum-deepak:
Working on updating the examples. I am keeping the following structure:

- Using ES6 syntax as far as possible.
- For charts, first data provider, followed by conf and then other methods.
- Prettier for formatting.
- eslint clean

Notes for myself:

- First, make the Google Analytics block to be ignored by eslint.
- Ensure it is clean as per eslint.
- Run prettier.
- Make changes.
- Run prettier.
- Check eslint


gordonwoodhull (approved):
Looks good!

Did you use any automation to do these updates? I am thinking that since user code will require extensive changes, it would be nice to provide scripts to do part of the work, if possible.

I think we have discussed [codemods](https://www.sitepoint.com/getting-started-with-codemods/) in the past. Have you ever tried such tools?

Disclosure: this kind of stuff is currently my day-job. I am using semantic patches (currently Semgrep's autofix) to remediate software vulnerabilities, but these tools are even better for API migration a.k.a. collateral evolution.

Edit: hmm, I see that JSCodeShift is not as friendly as other tools of this sort... you end up writing JavaScript to walk the AST and transform it, which is kind of confusing.

kum-deepak:
The codemods definitely make sense. Let me take a stab at that. I have not used any such tool so far. I understand the concept though. Considering JSCodeShift is not optimal, do you have any tool suggestions?

I see the following outline:

- identify a chart object - typically the creation and configuration are not likely to be together.
- on that chart object, typically, chained function calls will be there.
- out of the chained calls, some will move to `.configurae` and some to `dataProvider`.
- occasionally sematic of method change, like `.colors`, `.point` etc.
- the older function-based creation will need changing to constructor-based.
- I think whatever is covered in the current compat files can be automated.
- asking the user to make the code lint-clean before using the automated upgrade may be useful.

gordonwoodhull:
It looks like JSCodeShift is the standard here, and it should work.

I was hoping for something that had a domain specific language for doing the transformations, to make them easier to read, but I think we should use the standard.

As I understand it, these tools should easily cover the kinds of changes you describe.

gordonwoodhull:
Thanks @kum-deepak!

Have you checked in the JSCodeShift codemods? I am hoping that they will help explain the big changes in the API.

kum-deepak:
Well, the codemods are ready - https://github.com/dc-js/dc-codemods. The code is configuration-driven. Please have a look. Your feedback will be useful. I have also added a wrapper that allows it to work for HTML files (only the javascript script blocks).

Currently, it does quite a lot. It renames all the functions, converts calls to constructors, adds chartGroup concepts, creates configuration blocks, layers, points, and dataProvider.

Quite a few examples worked with just automated conversion - see commit https://github.com/dc-js/dc.js/pull/1850/commits/243a4813ab0dc5011184af3a6a2618ec219ded5e

In all other examples, the commit message lists key changes needed to make it work. These will help in completing the migration guide.

ScatterPlots are currently showing issues. I will investigate those. Interestingly these work with dc-compat.

The example switching-time-intervals.html uses many concepts apart from its core purpose. I will trim it to just its core purpose.

kum-deepak:
> Thanks @kum-deepak!
> 
> Have you checked in the JSCodeShift codemods? I am hoping that they will help explain the big changes in the API.

Wow, your comment crossed as I was posting mine. Please check my comment above.

gordonwoodhull:
Oh cool, I missed the new repo. Will check it out!

kum-deepak:
Trimmed switching-time-intervals to do exactly as the title says.

kum-deepak:
In ScatterPlot, we need an explicit default valueAccessor - d => d.key[1]. The compat code flow was attaching it. With this modification in codemod, examples needed very little change.

kum-deepak:
Other than transitions and resize folders, all other examples have been upgraded.

I am considering changing the default sizing mechanism, #1853. If that works it will obviate resize tests.

I am merging this PR now. Will work on the Upgrade guide.

Issue #1855, opened 2022-01-09
Stacked bar chart legend label alignment is off in Chrome


mgd722:
Opening the [stacked bar chart example](https://dc-js.github.io/dc.js/examples/stacked-bar.html) in Chrome 96.0.4664.110 (Official Build) (arm64) on an M1 MacBook Pro, the labels in the legend are offset:

![image](https://user-images.githubusercontent.com/12187845/148693751-835b68cf-79dc-44c1-8387-2a325f13e1e5.png)

Same page opened on the same computer, but using Safari 15.1 (17612.2.9.1.20):

![image](https://user-images.githubusercontent.com/12187845/148693809-eb28f8b0-6c7b-4334-9aa2-04b8ff968106.png)

This must have been caused by a recent Chrome update, because I would have noticed this before in my code.

mgd722:
Same with Chrome 97.0.4692.71. Not sure when exactly this broke but I am certain that this wasn't an issue before using this same code, so it must have been something that changed within Chrome.

kum-deepak:
I think it is the new Chrome. One of the automated tests started failing in Chrome recently (the automated tests pull the latest Chrome). See https://github.com/dc-js/dc.js/issues/1852. The failing test is related to legend label positioning.

congineer:
If someone will fix this please expose sub-label 'x', 'y', and 'LABEL_GAP' in dc.js API because hacking around is always bad for maintainability
![dc js-chrome-clientHeight-bug-fix-Screenshot_20220113_152726](https://user-images.githubusercontent.com/11654117/149349054-5bc96769-4286-4644-ab47-e2be7e1999a1.png)
.

Issue #1852, opened 2021-12-20
A test case failing on latest Chrome - Chrome Headless 96.0.4664.110


kum-deepak:
A single test case started failing in Chrome Headless 96.0.4664.110. First detected by failed CI runs in the `dc-v5` branch. On manual testing I found that the case fails even in the `develop` branch.

I am excluding the test case in the `dc-v5` branch. Not touching the `develop` branch as of now.

PR #1856, opened 2022-01-10
update legend text y calculation #1855 & #1852

Spacarar:
After reading about these two issues, I was investigating the cause of this and found that on chrome the `clientHeight` was returning a much smaller value than `getBoundingClientRect().height` which was returning more consistent results across environments for me.

Please let me know if there are any issues with the code here so I can adjust accordingly.

### DIFF
--- a/src/charts/legend.js
+++ b/src/charts/legend.js
@@ -322,7 +322,8 @@ export class Legend {
                 .classed('dc-tabbable', this._keyboardAccessible)
                 .attr('x', self._itemHeight + LABEL_GAP)
                 .attr('y', function () {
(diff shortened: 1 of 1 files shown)


kum-deepak:
Many thanks! I have checked the code and it seems all fine. The CI testes Edge, Chrome, Firefox, and Safari, so, I guess we are good. This change request will get merged and released as part of the next release.

ufundo:
Any chance of this fix being released for existing stable version?

powelleric:
I would also love to have this fix.  Thanks a bunch!

Issue #1853, opened 2021-12-30
Approach to resize


kum-deepak:
While checking the resize test folder, I realize that we can simplify the code using ResizeObserver. This considers padding, borders, etc.

```javascript

          const handleResize = (width, height) => {
            chart.width(width);
            chart.height(width * 0.6);
            chart.configure({ innerRadius: width * 0.14 });
          };

          const withoutTransitions = callback => {
            const oldVal = chart.conf().transitionDuration;
            chart.configure({ transitionDuration: 0 });
            callback();
            chart.configure({ transitionDuration: oldVal });
          };

          const resizeObserver = new ResizeObserver(entries => {
            const entry = entries[0];
            const width = entry.contentRect.width;
            const height = entry.contentRect.height;
            withoutTransitions(() => {
              handleResize(width, height);
              chart.redraw();
            });
          });

          resizeObserver.observe(document.querySelector('#test'));
```

I will investigate later if this can be used as default in case `height` and `width` are not explicitly set.

gordonwoodhull:
Nods, Safari [finally got ResizeObserver last year](https://caniuse.com/resizeobserver), so we can get rid of these workarounds.

kum-deepak:
Thanks, @gordonwoodhull for the quick feedback. I am also thinking of a single callback handling both width and height. This will make sense in most of the cases. On purpose, I picked an example where `innerRadius` also needs to be adjusted.

I will come back to this after working on the Upgrade guide.

kum-deepak:
Closed in favor of #1858

PR #1858, opened 2022-02-14
Auto resizing using ResizeObserver

kum-deepak:
This implementation picks from #1853. The key points:

- In the auto-mode, the chart will observe the CSS box size of the anchor element, excluding padding and border, and keep resizing.
- `width` and `height` are no longer modified by the charts, so, these have now been moved to `configuration`.
- If `width` (or `height`) is provided that will be used, if not the corresponding dimension will be picked as per the CSS box, subject to `minWidth`.
- As part of `configuration` there is a callback `beforeResize`, which can be used to adjust other dimension-related parameters. For example, the `resizing-pie` uses it to adjust the inner radius and the `resizing-series` uses it to keep the legends right aligned.
- The resizing examples use margin, border, and padding to check that dc honors those.

Future considerations:

- Make this the preferred mode, update examples accordingly.
- Make default values of `minWidhth` and `minHeight` smaller, say 50. The minimum should be controlled by CSS if someone needs it. The library default should only be to protect developers from setting the height to `0` without realizing it. If we keep it high, users may get frustrated that their responsive charts are not working.


kum-deepak:
@gordonwoodhull, I am merging this as I need it for one of the subsequent changes. Please review, I will create a new PR with your feedback.

gordonwoodhull:
Nice changes, really cleans things up to work with width and height this way.

I'm sure you're on it, but please continue to update the codemodes as the API changes.

kum-deepak:
Thanks for your review! Codemods is configuration-driven, adding `width` and `height` to the configuration was all that it needed (https://github.com/dc-js/dc-codemods/commit/d5eb6cac4e594ae033719e13414344d4f75468bc). 

I am considering removing explicit width and height from examples. Still thinking about that.

gordonwoodhull:
Do you mean setting the width and height on the elements using CSS instead? Seems like it still needs to be set somewhere.

kum-deepak:
Yes, that way the examples will be responsive.

I realize that people pick code from examples (like I did when I started using dc), so, if the examples are as per what we would recommend, it will help users.

gordonwoodhull:
Oh, you mean making all of them responsive instead of having separate `resizing/` examples? That sounds good. I guess we should keep one or two non-responsive for ppl who still want to do that, but I agree responsive is probably the more popular & impressive case.

PR #1859, opened 2022-02-14
Update legend text y calculation

kum-deepak:
Many thanks @Spacarar! Please see #1855, #1852 & #1856.


kum-deepak:
This fixes the issue in the `dc-v5` branch.

PR #1870, opened 2022-09-06
Added new Bootstrap version to three examples

kleczekr:
Added new Bootstrap version to three examples (for now from a CDN) and fixed formatting issues in one of the examples caused by old classes.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "liveServer.settings.port": 5501
+}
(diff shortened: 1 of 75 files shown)


kum-deepak:
A few of the test cases were failing, I have updated the CSS class name in the test cases, and now all test cases pass. Other than that I have reordered and squashed commits into fewer commits.

kum-deepak:
Please check the last few commits in dc-v5 branch - https://github.com/dc-js/dc.js/commits/dc-v5

